PR #169, opened 2015-01-07
168 Fixes test config for local (and Travis, I assume)

otupman:
Essentially adds a test.js file which would not have been committed because everything other than `_template.js` was git-ignored in the secret directory. 

Also had to change the (only) test as the test usernames were invalid as per the username validation regex! 

:ocean:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ tmp
 # Configs
 config/secret/*
 !config/secret/_template.js
+!config/secret/test.js
 
 # Deploy & maintenance scripts
 scripts/*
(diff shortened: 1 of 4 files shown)


otupman:
Hmm, hold on with any merge - Karma doesn't work, fixing that atm.

otupman:
Okkk, so in theory this now results in Karma not failing _but_ there are failing tests because the tests are not up-to-date with the code base.

Example: [messages.client.controller.test.js](https://github.com/Trustroots/trustroots/blob/master/public/modules/messages/tests/messages.client.controller.test.js) - it references the 'MessagesController', however this no longer appears to exist (been replaced by the inbox.client & inbox.thread controllers).

I'd argue that the best way of proceeding is commenting out (_shudder_) the out-of-date tests and then add a new ticket to bring the broken tests up to date (or remove them entirely). Reasoning: the actual tests running (and failing) is better than the test suite simply not working.

If agreeable, then regard this PR as ready for merging :sunglasses:

guaka:
Thanks, @otupman!

PR #219, opened 2015-04-05
Block empty profiles

chmac:
Unless a user has a description > 140 characters (including HTML tags), they are not able to send messages to other users.

@simison Not sure if this is implemented in the best way, but it seems to work in my (very crude) testing.


simison:
Finally done with https://github.com/Trustroots/trustroots/commit/897fe355e4ceb75c3e54389590f9b94ff6f5146b :-)

PR #220, opened 2015-04-05
Merge configs

chmac:
This change recursively overwrites configs. However, this may have unintended consequences if the current code relies on the fact that subsequent configs overwrite the entire objects of previous ones. For example, with the `mailer`. If the `all.js` config has `mailer: { service: 'gmail', auth: {user: 'user', pass: 'pass'}}` and then a later config has `mailer: { host: 'localhost' }` then the resulting config will now be `mailer: { service: 'gmail', auth: {user: 'user', pass: 'pass'}, host: 'localhost'}` while previously the whole `mailer` object would have been overwritten.

In the long term I think this change makes a lot of sense, and is much more intuitive for new developers. However, it does add some complication.

The pull request also includes the `block-empty-profiles` branch (which should be merged first), or this branch will need to be rebased (is that the right term?) from master.

### DIFF
--- a/app/controllers/messages.server.controller.js
+++ b/app/controllers/messages.server.controller.js
@@ -6,6 +6,7 @@
 var mongoose = require('mongoose'),
     async = require('async'),
     _ = require('lodash'),
+    config = require('../../config/config'),
     errorHandler = require('./errors'),
     sanitizeHtml = require('sanitize-html'),
     userHandler = require('./users'),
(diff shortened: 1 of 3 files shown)


simison:
Ended up doing this a bit differently: https://github.com/Trustroots/trustroots/commit/8bdc4f49b5848c4de56415372a2b4351b2604043

It's a combo of both _.extend() and _.merge().

Issue #214, opened 2015-04-05
Cancel should clear unsaved profile changes


chmac:
If you edit your profile and click "Cancel", then click "Edit Profile" again, the changes are still there. I'm guessing it's because angular doesn't destroy and recreate the form. The effect is that I need to reload in order to be able to edit my profile as it is currently stored.

The "Cancel" link / button should clear any changes I've made.

simison:
Good catch.

PR #230, opened 2015-04-10
Fixed bug #214  canceling edit profile

unsanded:
I guess having a copy of the user profile for editing makes sense, because it is not "real" until it is committed.
Just let me know if i'm being naive or stupid here ;)


simison:
Thanks! I tested it, works perfect.

Issue #240, opened 2015-04-13
Look into username lowercase/CamelCase details


simison:
- [x] If user signs up with CamelCase username, show it at the profile like that (now enforced lowercase)
- [x] Logins with lowercase will work when real username is CamelCase
- [x] Logins with CamelCase will work when real username is lowercase
- [x] Logins with CamelCase and lowercase emails work 
- [x] [Password recovery](https://www.trustroots.org/#!/password/forgot) works same as above
- [x] Looking for profiles with [lowercase](https://www.trustroots.org/#!/profile/dummy) and [CamelCase](https://www.trustroots.org/#!/profile/Dummy) should be no difference
- [x] Make sure you can't sign up with two usernames; "camelcase" and "CamelCase" (keep lowercase db field unique?)

Probably makes sense to keep saving lowercase version of the username to the db, to make things quicker. We'd just need separate "display" version of the username. Then all the DB queries can be against lowercase (like they're now) and when sending stuff to frontend, switch it to the display one. [See this StackOverflow question](http://stackoverflow.com/questions/1863399/mongodb-is-it-possible-to-make-a-case-insensitive-query).

Username validator/unique/model: https://github.com/Trustroots/trustroots/blob/master/app/models/user.server.model.js

Frontend bit at auth controller: https://github.com/Trustroots/trustroots/blob/master/public/modules/users/controllers/authentication.client.controller.js#L32-L37

simison:
Made a few small additions and tested everything. Big thanks!

PR #252, opened 2015-04-16
#240 Desensitized username casing + fixed dummy datafill script

katsuroo:
issue #240
- Desensitized casing in profile lookups , login
- Profiles reflect original entered username casing
- migration script added
- fixed minor error with filltestdata.js


katsuroo:
Made the recommended changes. Thanks for catching that :)

PR #256, opened 2015-04-23
Stricter username validator

katsuroo:
-Removed username upper limit
-Username must contain at least one alphanumeric character
-Username cannot begin or end with .
-Added tests for username validation

*note
Have to look into test env. When running test with grunt test the first test(regardless of what it is) returns an error:

     Uncaught TypeError: Cannot set property 'tagline' of undefined
      at module.exports (config/express.js:46:34)
      at server.js:32:40


simison:
Obs, I forgot to merge this earlier after tested it. Sry!

PR #361, opened 2016-06-12
Improving fuzzyLocation(). It was not very fuzzy before.

mrkvon:
The `fuzzyLocation` function in previous version was quite predictable. By moving from the fuzzy location 354m (sqrt(2) * 250) south-west we can get less than 71m (sqrt(2) *50) from the original location on a single line.

The new function `fuzzyOffset(minimum,maximum)` chooses random distance within a specified range `[minimum, maximum)` and random direction (radians) and returns offset in `[x, y]` format. It is random in 2 dimensions, not in 1. The `fuzzyOffset(100, 200)` is called from `fuzzyLocation` function.

This is my first pull request on GitHub, so I appologize if any mistakes were made. It is learning in process.

### DIFF
--- a/modules/offers/server/controllers/offers.server.controller.js
+++ b/modules/offers/server/controllers/offers.server.controller.js
@@ -13,6 +13,26 @@ var path = require('path'),
     User = mongoose.model('User');
 
 /**
+ * Create a fuzzy offset between specified distances
+ * @param {Number} minimum - Minimal distance
+ * @param {Number} maximum - Maximal distance
+ * @returns {Array<Number>} - array of length 2 (horizontal and vertical offset)
+ */
+function fuzzyOffset(minimum, maximum) {
+  //please note that Math.random() is not cryptographically secure.
+  //for this purpose it's probably ok, but can be improved i.e. with node crypto module.
+  if(maximum < minimum) throw new Error('maximum must be greater than minimum');
+  var difference = maximum - minimum;
+  var randomDistance = Math.floor(difference*Math.random()+minimum); //Distance will be from interval [minimum, maximum)
+  var randomDirection = 2*Math.PI*Math.random(); //random direction is from interval [0, 2*PI) radians
+
+  var horizontal = randomDistance * Math.cos(randomDirection);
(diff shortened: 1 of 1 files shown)


simison:
Hi, thanks!

Though as far as I could remember, the old one was two dimensional as well. It sorta picked a location from a donut:

image

;-)

First random direction from the point and then distance between 100 and 200 meters from the point.

But any improvement or improved readability is welcome, I'll check review this a bit later.

mrkvon:
We might be talking about different version of the `fuzzyLocation` function. Maybe a different one is present somewhere else and used?

The function i talk about _had_ the problem described. One hint may be that it called Math.random() only once (for a single (diagonal) dimension).

fuzzy

I attach a simple comparing test for location `[0,0]`. The old function has only positive output in north-east direction. The new one outputs in all directions (donut).

If one sets her position precisely on trustroots.org, one can check on the map whether it is approximately 350 metres north-east from the original place. In my case it is like that. I suspect that all the fuzzy locations are biased in this way.

[fuzzy-location.txt (should be renamed to *.js which GitHub doesn't allow)](https://github.com/Trustroots/trustroots/files/310687/fuzzy-location.txt)
`node fuzzy-location.(txt|js)`
[output.txt](https://github.com/Trustroots/trustroots/files/310685/output.txt)

simison:
Oh, then this is excellent. Good finding!

Could you remove lines 38 and 43-46. After that squash and I'll merge.

If you have any ideas how to write a CI test for this, go for it, but it's not necessary as it might be quite hard...

mrkvon:
@simison i did as you suggested, but left line 38. Imho the link is still a valid resource (just explaining how to convert offset in meters to latitude/longitude).

I don't know what are or how to write CI tests (yet).

Now the donut is set by 100-200 m distance from the original location. Maybe different values are wanted.

example of 1000 points generated by old (violet) and new (green) `fuzzyOffset` function (scripts & gnuplot commands on request).

fuzzy

simison:
Super. Excellent to have those generate graphs to illustrate this. Thanks a bunch!

PR #364, opened 2016-07-09
Fix checking redirect messages in tests for >=4

nicksellen:
Previously it just checked for v4 or v5.
Now uses semver to properly check it is >=4.0.0.


simison:
Great!

PR #371, opened 2016-07-18
Fix missing returns in callback error cases

### LINKED ISSUE #370: Forgetting to return in error case
There are a number of places where the error handling calls `done()` with an error, but does not return, this would lead to it being called twice.

e.g.:

```
if (!friend) done(new Error('Failed to load user ' + req.body.friendUserId));
done(err, contact, messageHTML, messagePlain, friend);
```

nicksellen:
Fixes #370


simison:
:thumbsup:

PR #372, opened 2016-07-18
Prevent duplicate contacts

nicksellen:


nicksellen:
It's supposed to be broken :) I just added failing tests...

modules/contacts/server/controllers/contacts.server.controller.js
```diff
@@ -45,6 +48,29 @@ exports.add = function(req, res) {
...
+          { users: [ req.body.friendUserId, req.user._id ] },
+          { users: [ req.user._id, req.body.friendUserId ] }
+        ]
+      }).exec(function(err, existingContact) {
+        if (err) return done(err);
+
+        if (existingContact) {
+          // Contact already exists!
+          return res.status(400).json({
+            message: errorHandler.getErrorMessageByKey('invalid-id')
```
nicksellen: This is not a very satisfactory error message

simison: You're right.

I'm not huge fan of passing error messages directly to front anyway though, because it easily ends up user seeing too technical messages. So I'd be inclined to see if the error can be reflected with correct 4xx status code. WDYT?

simison: Actually, if users are already connected, it shouldn't yeld scary error at the frontend (error at the API is good), but rather just semi success informing something like "Hey great, you guys are already connected!"

nicksellen: [409 Conflict](https://httpstatuses.com/409) seems a good choice:
http://stackoverflow.com/a/3826024/2922612
http://stackoverflow.com/a/25541368/2922612

And I agree about the message, I'll fix something up.

modules/contacts/server/controllers/contacts.server.controller.js
```diff
@@ -45,6 +48,30 @@ exports.add = function(req, res) {
...
+          { users: [ req.user._id, req.body.friendUserId ] }
+        ]
+      }).exec(function(err, existingContact) {
+        if (err) return done(err);
+
+        if (existingContact) {
+          // Contact already exists!
+          return res.status(409).json({
+            message: errorHandler.getErrorMessageByKey('conflict'),
+            confirmed: existingContact.confirmed
```
nicksellen: > The response body SHOULD include enough information for the user to recognize the source of the conflict.
Recommended at http://www.restpatterns.org/HTTP_Status_Codes/409_-_Conflict

Hence, passing _confirmed_ status back in the body too. Could pass entire contact back, but this seems enough.

modules/contacts/client/controllers/add-contact.client.controller.js
```diff
@@ -68,7 +68,12 @@
         vm.success = 'Done! We sent an email to your contact and he/she still needs to confirm it.';
       }, function(error) {
         vm.isLoading = false;
-        vm.error = error.message || 'Something went wrong. Try again.';
+        if (error.status === 409) {
+          // 409 means contact already existed
+          vm.success = (error.confirmed) ? 'You two are already connected. Great!' : 'Connection already initiated; now it has to be confirmed.';
```
nicksellen: I thought showing it as a success is appropriate. It is then in the same state as if they had just visited that page and the contact already existed. There is nothing the users needs to do too, so an error seems harsh.

nicksellen:
I might have a pop at improving the frontend tests now I've modified that too, they don't test very much in this area right now.

nicksellen:
Ok, this is ready for review.

Summary of changes:

* backend
  * checks if contact exists first, if it does, *409 Conflict* error is returned
  * if contact is created, _then_ error occurs before end of request, the contact is cleaned up (i.e. 
  * tests added for normal cases and these new ones
* frontend
  * if contact is added but already exists, it shows a success message
  * tests added for ContactAddController for most cases

simison:
Tested this out by letting email sending fail. :thumbsup:

PR #373, opened 2016-07-18
Add gulp client watch

nicksellen:
Added client test watching to match the server test watching.


nicksellen:
Oops I made this on top of the wrong branch.

PR #374, opened 2016-07-18
Add watch mode to client tests

nicksellen:
Add client test watching to match server test watching.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "fontello": "gulp fontello",
     "test": "gulp test",
     "test:client": "gulp test:client",
+    "test:client:watch": "gulp test:client:watch",
     "test:server": "gulp test:server",
     "test:server:watch": "gulp test:server:watch",
     "migrate": "migrate",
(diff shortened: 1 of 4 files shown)


simison:
Good one!

Could you add this to [package.json](https://github.com/Trustroots/trustroots/blob/master/package.json#L25-L27) as well?

These, too:
- https://github.com/Trustroots/trustroots/blob/master/INSTALL.md#running-tests
- https://github.com/Trustroots/trustroots/blob/master/INSTALL-DOCKER.md#running-tests

package.json
```diff
@@ -23,6 +23,7 @@
     "fontello": "gulp fontello",
     "test": "gulp test",
     "test:client": "gulp test:client",
+    "test:client:watch": "gulp test:client:watch",
```
nicksellen: I did already :)

simison: Oh, clearly bed time for me. :-)

nicksellen:
Just added it to the install docs too now

simison:
Tested, works great.

PR #376, opened 2016-07-25
Replace JShint with ESLint

simison:
Slightly outdated version of `eslint` + `eslint-config-airbnb`, but I just took these from [Mean.js boilerplate](https://github.com/meanjs/mean).

Works for now, will update later or wait for mean.js folks to update theirs.


simison:
@nicksellen LGTY? I rebased your latest PR into this as well.

https://github.com/Trustroots/trustroots/blob/467f57ac9c5b007ab47cc387fe6051caa7de55f4/.eslintrc.js is the only file that really matters here.

nicksellen:
Yup looks good :)

Maybe I have more thoughts when I play with it, but it's great to have it in master as soon as possible.

simison:
Done!

PR #380, opened 2016-07-30
Add POST /api/auth/resend-confirmation endpoint

nicksellen:
Ok, should all work now.

* added under `/api/auth` not `/api/users` as I think it relates more to auth concerns
* it sends the `signup` email template but with the _campaign_ field marked as `resend-confirmation`. I don't know if you need to configure those somewhere

There is quite a lot of non-DRY code in this area, but I kept it simple as this is another concern.

### DIFF
--- a/modules/users/client/controllers/profile-edit-account.client.controller.js
+++ b/modules/users/client/controllers/profile-edit-account.client.controller.js
@@ -40,7 +40,7 @@
                           'If you don\'t see this email in your inbox within 15 minutes, look for it in your junk mail folder. If you find it there, please mark it as "Not Junk".';
         vm.user = Authentication.user = response;
       }, function(response) {
-        vm.emailError = response.data.message || 'Something went wrong.';
+        vm.emailError = (response.data && response.data.message) || 'Something went wrong.';
       });
     }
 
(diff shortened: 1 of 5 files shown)


nicksellen:
Rebased to master

simison:
>Rebased to master

Oh, thanks!

Sorry this has been hanging out here for a while — I've had pretty hectic times. Today finally a proper full office day, gonna look into this now.

simison:
Actually, there are two distinct features in play here:
1. Signup ➡️  re-send the signup email confirmation
2. Change email to new one ➡️  re-send email change confirmation

Signup re-sending works great now.

Now if I try feature 2:
1. User is `public:true`,
2. Initiate email change from page `/profile/edit/account`
3. You can see "Confirm email change" email dropping in
4. Come back to that account page and click "resend" button ➡️  gets error `{message: "Already confirmed."}` and at console log also `TypeError: Cannot read property 'message' of undefined`.

Should we use user-controller's `update()` for the feature 2 and leave `/api/auth/resend-confirmation` for signup confirmations only?

nicksellen:
I would just add it into the `/api/auth/resend-confirmation` endpoint, changing email is an auth concern (as it is the method that users are authenticated with), and it's kind of similar to doing the original confirmation.

Maybe different email templates would be useful though (this will be simpler once the email service is ready though).

simison:
There are separate templates for signup confirmation and email change confirmation:

- `modules/core/server/views/email-templates/signup.server.view.html`
- `modules/core/server/views/email-templates/email-confirmation.server.view.html`

nicksellen:
Quite right :)

Ok, works in manual and automated testing now. Can't do automated email testing yet, as that work is in another branch.

simison:
Tested and works great. Thanks a bunch!

Issue #379, opened 2016-07-25
Email sending tests


nicksellen:
Originally mentioned in #377, but it's a separate issue really. Right now there is no testing for whether the correct emails get sent. This is not for testing whether the network sending works, but whether the app logic is deciding to send the correct email.

I looked at a mock transport lib [1] but it doesn't handle the case where the `to` is an object [2], which is how it is used in trustroots.

But peeking at how nodemailer does it's testing [3], I just wrote a simple function to collect sent emails into an array, you can see it in action in this branch [4] - I just dumped `testutils.js` into the root as I couldn't see any logical place to put it, maybe there is a better place.

Whaddya think?

If you like the approach I would add tests for all email sending, and remove the stub transport as it would no longer be needed (except in one place where I used it to trigger an error on send).

[1] https://github.com/bcoe/nodemailer-mock-transport
[2] https://github.com/bcoe/nodemailer-mock-transport/blob/master/index.js#L37
[3] https://github.com/nodemailer/nodemailer/blob/master/test/nodemailer-test.js#L27
[4] https://github.com/Trustroots/trustroots/compare/email-testing

chreekat:
Great idea, imho

simison:
Great, a very elegant approach.

I've been putting `testutils.js` kind of stuff to either under `modules/core/server/controllers` or `config/lib` depending if it's used to config/init something or used regularly from other controllers. 

Or maybe this stuff should be either at the root of `modules/core/tests` or under some folder there.

BTW, as a side note, [redirectMessage()](https://github.com/Trustroots/trustroots/blob/master/modules/users/tests/server/user.server.routes.tests.js#L1447-L1458) function from user tests could be moved to `testutils.js` as well.

nicksellen:
This is paused whilst email queue work in #381 is in progress

simison:
@nicksellen This can be closed, right?

nicksellen:
Indeed

Issue #363, opened 2016-06-29
notification emails not arriving


platschi:
Received a hosting request 5 days ago by Trustroots. Usually don't check the website, but rely on those messages arriving into my email inbox. That email never arrived. Unfortunately, saw the request only today on the website itself. Might there be problems with email delivery? And no, no Trustroots emails have been caught by Spam/Junk/Antispam/Virus whatever filters. 15 days ago, I received a request, back then all was fine with the notification email.

simison:
The notification setup got two big overhauls in attempt to get them more error resistant — should work much better now.

PR #381, opened 2016-07-31
[WIP] Add job queue for sending emails

simison:
There was a need to make email delivery more robust, failure safe and easier to test.

**Minus:**
- Adding Queue server introduces yet one more obstacle on getting Trustroots running, but on the other hand installing RabbitMQ seems veeery straightforward (on OSX just `brew install rabbitmq`). Nothing more complicated with Docker compared to current setup. It could be possible to build our own queue script which simply puts jobs to MongoDB, but the complication cost vs time spent building something that's more integrated 

- Worker that picks emails from the queue is a separate instance, but `npm start` now runs both using [concurrently](https://www.npmjs.com/package/concurrently). It might be unclear from which instance stdout/stderr logs are coming from. I clarified it a bit by adding `[worker]`/`[server]` prefixes here and there. Worker could also be just [Agenda](https://github.com/rschmukler/agenda) cron job running at the main instance, but the main instance is being forked to serve more people, so it's not the best place to run these sort of things.

**Plus:**
- Waaaay more robust setup. Easier to just set email aside for sending it later and let application process move forward. In case sending fails, we can try again n times.

- Email handling should anyway be moved to separate services from controllers.

- It will be easier mark bouncing email addresses and thus improve our deliverability.

**Todo:**
- [ ] Handle failures better at signup
- [ ] Remove failing emails from the queue (now it just gets stuck to infinite loop with them)

--- 
Related to #363, #379


nicksellen:
I'm not sure why Agenda isn't usable for this?

It has a [now](https://github.com/rschmukler/agenda#nowname-data-cb) scheduling time which would save it to db, and run it in a worker right away (the sending email use case). It supports multiple workers connected to the same db (worker locks each job), and has a [reasonable looking ui](https://github.com/joeframbach/agendash).

Getting job management right (retries, errors, visibility, management, etc) correct is fiddly so I would advocate to outsource this to a lib. See all the stuff [kue](https://github.com/Automattic/kue) does beyond just the message queue aspect (kue being another redis-backed option of course, which has a [reasonable looking ui](https://cloud.githubusercontent.com/assets/1458008/5229932/76dd0e70-7716-11e4-9551-e87ce799d8dc.png) too)

I think rabbitmq is ultimately a better solution than polling a mongodb table as Agenda does, but given Agenda is already in place...

At least from the controller perspective it can be implementation agnostic anyway, just an email service method that will send it at some point.

simison:
Hm, could be, it would be simpler setup then. I was initially worried that running such jobs at the main process would be blocking the app, but reading https://github.com/rschmukler/agenda/issues/8 it seems like Agenda handles it pretty well.

nicksellen:
Yes, you can run all the jobs in a totally separate process https://github.com/rschmukler/agenda#multiple-job-processors. It would seem nice to have `server.js` and `worker.js` still though, and they can use node cluster if they want multiple processes.

The node cluster setups on that link [1] are missing _intentional exit_ handling though [2].

[1] https://github.com/rschmukler/agenda/issues/8
[2] https://nodejs.org/api/cluster.html#cluster_worker_exitedafterdisconnect (called _suicide_ in older node versions)

nicksellen:
https://github.com/Trustroots/trustroots/pull/383 is my agenda version

Issue #338, opened 2016-03-03
Update to MongoDB 3.2


simison:
Currently running 2.x

https://www.mongodb.com/mongodb-3.2

Might require changes to tests/code at least due changes in unique field errors.

simison:
Also make sure to lock down the version here: https://github.com/Trustroots/trustroots/blob/master/docker-compose.yml#L46

simison:
[Agenda](https://github.com/rschmukler/agenda) might not support 3.2 yet.

simison:
This needs fixing:

```
trustroots_1 | /trustroots/node_modules/mongoose/node_modules/mongodb/lib/utils.js:98
trustroots_1 |     process.nextTick(function() { throw err; });
trustroots_1 |                                   ^
trustroots_1 | 
trustroots_1 | TypeError: Cannot read property '1' of null
trustroots_1 |     at beautify (/trustroots/node_modules/mongoose-beautiful-unique-validation/index.js:20:28)
trustroots_1 |     at module.exports.schema.methods.trySave.proxyFn (/trustroots/node_modules/mongoose-beautiful-unique-validation/index.js:110:21)
trustroots_1 |     at /trustroots/node_modules/mongoose/lib/document.js:1813:19
trustroots_1 |     at proto.(anonymous function).handleError (/trustroots/node_modules/mongoose/node_modules/hooks-fixed/hooks.js:40:22)
trustroots_1 |     at proto.(anonymous function).next_ (/trustroots/node_modules/mongoose/node_modules/hooks-fixed/hooks.js:75:26)
trustroots_1 |     at fnWrapper (/trustroots/node_modules/mongoose/node_modules/hooks-fixed/hooks.js:186:18)
trustroots_1 |     at /trustroots/node_modules/mongoose/lib/model.js:287:15
trustroots_1 |     at /trustroots/node_modules/mongoose/lib/model.js:188:14
trustroots_1 |     at /trustroots/node_modules/mongoose/lib/model.js:132:9
trustroots_1 |     at /trustroots/node_modules/mongoose/node_modules/mongodb/lib/collection.js:491:20
trustroots_1 |     at /trustroots/node_modules/mongoose/node_modules/mongodb/lib/collection.js:604:14
trustroots_1 |     at handleCallback (/trustroots/node_modules/mongoose/node_modules/mongodb/lib/utils.js:96:12)
[...]

PR #383, opened 2016-08-10
[WIP] Agenda job queue

nicksellen:
Here's a bunch of stuff to send emails in agenda jobs, and using a service layer too. Most of the work is not actually the agenda bit, but swapping out the current email sending stuff into the service layer. Agenda is not deeply involved. I added various tests to check the content of emails now.

It's not finished yet though. Want to add some more tests. And do some manual testing. And a bit more careful checking. Toooo sleepy now.

I thought it might be a benefit for you to glance over the general approach.


nicksellen:
Hmm, I think travis-ci has old mongo, and agenda uses some new mongo stuff. They use travis themselves and manually install a newer version [1].

[1] https://github.com/rschmukler/agenda/blob/master/.travis.yml#L15.

nicksellen:
http://www.theodo.fr/blog/2014/12/choose-a-specific-mongodb-version-in-travis/ has a nice guide for using a newer mongo with travis. will try it out tomorrow prob.

simison:
Interesting. It's also Agenda that's stopping me upgrading to Mongo v3 #338

config/lib/render.js
```diff
@@ -0,0 +1,4 @@
+/**
+  Template rendering function
+*/
+module.exports = require('consolidate').swig;
```
simison: Nice move!

Actually Swig is quite a bit outdated and I'd like to replace it with Nunjucks or similar, but I'll wait for this before doing anything about it: https://github.com/meanjs/mean/issues/1286

Just sayin, in the scope of this PR Swig is still okay.

modules/core/server/services/email.server.service.js
```diff
@@ -0,0 +1,219 @@
...
+  return params;
+};
+
+exports.renderEmail = function(templateName, params, callback) {
+
+  var htmlTemplatePath = path.join(htmlTemplateDir, templateName + '.server.view.html');
+  var textTemplatePath = path.join(textTemplateDir, templateName + '.server.view.html');
+
+  // Rendering in parallel leads to an error. maybe because
+  // swig is unmaintained now https://github.com/paularmstrong/swig)
```
nicksellen: Yup, swig is definitely outdated/unmaintained, using parallel async fails here as state seems to get mingled between between renders, which is deeply worrying! I didn't dig into it any more yet.

simison: Hmm, would it make sense if I spend a couple hours getting rid of Swig? I had an old branch where I tested it (very briefly) but since it wasn't very high on priorities, I dropped it. Just published it here:

https://github.com/Trustroots/trustroots/tree/replace-swig-with-nunjucks

nicksellen:
Agenda seems work fine for me in development (using mongo v3.2.8).

Also, it's possible to pass a mongodb client into the agenda options, so you can use whichever mongodb drive you want: https://github.com/rschmukler/agenda/blob/master/lib/agenda.js#L46

nicksellen:
> Disallowing sources: mongodb-2.6-precise
> If you require these sources, please review the source approval process at: https://github.com/travis-ci/apt-source-whitelist#source-approval-process

Waaa, travis maintain an _approved sources_ list [1], and mongo 2.6 is not in it (3.x is though). Will upgrade it to 3.

[1] https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json#L203

modules/core/server/services/email.server.service.js
nicksellen: I don't think it is a big issue to keep swig here for now. But also wouldn't be a problem to rebase either branch on top of any changes. I'd keep them in seperate branches/PR's though.

nicksellen:
Job failure retries working now. I'd like to add some tests for this though.

I added agendash (and maildev) to `npm start` to make it simpler to run up. The agendash is a bit buggy (requeue and delete commands are broken), but basically does the job. If we stick with agenda, could spruce it up a bit...

nicksellen:
Rebased onto master and fixed up signup/confirmation stuff. Also I was sending the email change confirmations to the old email address before, whoops. Covered by the tests now though.

Maybe check the utm campaigns are set as intended in [1]? Not sure where they go afterwards...

nicksellen:
Two more things I am intending for this:

1. get the job retry logic under test
1. manually testing the unread message emails

Then I think that is it from my perspective.

simison:
>Maybe check the utm campaigns are set as intended in [1]? Not sure where they go afterwards...

UTM campaign codes are just for analytics, there isn't any prefixed list so they can really be anything that just describes where the user landed to the page from. I've been meaning to write a quick short guide to them but seems like just looking how they've done so far works. ;-)

nicksellen:
So maybe it's worth distinguishing between change email confirmation and signup email confirmation?

simison:
>So maybe it's worth distinguishing between change email confirmation and signup email confirmation?

Yeah, would be. (I didn't yet go trough them from your code.)

nicksellen:
Closing as this has the wrong branch target :( Will reopen a new one shortly...

simison:
Follow ups #386

PR #385, opened 2016-08-11
Refactoring server application service to enable modularity

simison:
- Use callbacks
- Provide app, db, config variables outside of server.js
- Pass Mongoose NodeJS ES6 promises instead of internal, deprecated
mpromise, fixes this warning:
```
Mongoose: mpromise (mongoose's default promise library) is deprecated, plug in your own promise library instead: http://mongoosejs.com/docs/promises.html
```

From most parts work from
https://github.com/meanjs/mean/commit/edb62344bc4af3cb28c3469dc7463c13a2
fc76a1 — it'll be easier to keep up with their work when our codebase is more similarly structured.

### DIFF
--- a/config/lib/app.js
+++ b/config/lib/app.js
@@ -1,0 +1,66 @@
+'use strict';
+
+/**
+ * Module dependencies.
+ */
+var config = require('../config'),
+    mongoose = require('./mongoose'),
+    express = require('./express'),
+    chalk = require('chalk');
+
+// Initialize Models
+mongoose.loadModels();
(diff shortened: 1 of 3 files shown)


simison:
Oh yeah and fixes that
```
Mongoose: mpromise (mongoose's default promise library) is deprecated, plug in your own promise library instead: http://mongoosejs.com/docs/promises.html
```
—warning.

simison:
Nick says "looks good". :-)

PR #386, opened 2016-08-11
Agenda job queue

nicksellen:
Continuation of https://github.com/Trustroots/trustroots/pull/383 (which had wrong target branch).


nicksellen:
Rebased to master, and added agenda/worker tests :)

nicksellen:
... and manual tests of unread message notifications passed

modules/core/tests/server/worker.tests.js
```diff
@@ -0,0 +1,152 @@
+var _ = require('lodash'),
```
simison: Could you add `'use strict';` here?

config/lib/worker.js
```diff
@@ -0,0 +1,73 @@
+var path = require('path'),
```
simison: Could you add 'use strict'; here?

package.json
```diff
@@ -28,7 +31,8 @@
     "test:server:watch": "gulp test:server:watch",
     "migrate": "migrate",
     "update": "npm install --quiet && npm prune && migrate",
-    "postinstall": "bower install --allow-root && bower prune --allow-root"
+    "postinstall": "bower install --allow-root && bower prune --allow-root",
+    "agendash": "agendash --db=mongodb://localhost/trustroots-dev --collection=agendaJobs --port=3001"
```
simison: Agendash could be mentioned at [Install instructions](https://github.com/Trustroots/trustroots/blob/master/INSTALL.md#running--development)?

simison:
Been testing this locally, looks great and works perfect!

Left a couple inline comments. (I can also update install instructions later.)

nicksellen:
I added maildev and agendash to the start script, so if you like that approach, the logical thing to do would be remove the seperate install steps for them, add the maildev smtp config in the example config file, and then just mention would urls are available after running start.

nicksellen:
I updated install docs as if you like including them in `npm start` :) - didn't look at the docker setup though...

config/lib/agenda.js
```diff
@@ -0,0 +1,9 @@
+var Agenda = require('agenda'),
```
nicksellen: oh, missed out the `'use strict';` here too

simison: I looked into switching [eslint strict](http://eslint.org/docs/rules/strict) rule on because of this, but got too many warnings and backed up. ;-) Seems like we're using it in quite mixed fashion, which complicates linting.

simison:
Yep, install.md + npm looks good for me now.

I think this is good for merging now.

Issue #284, opened 2015-07-29
AngularJS best practises


simison:
Basically following up with https://github.com/johnpapa/angular-styleguide

Mostly, but not limited to these:
- [JavaScript Closures](https://github.com/johnpapa/angular-styleguide#javascript-closures)
- [Route Resolve Promises](https://github.com/johnpapa/angular-styleguide#route-resolve-promises)
- [Move non-view logic out of controllers](https://github.com/johnpapa/angular-styleguide#defer-controller-logic-to-services)
- [Named functions](https://github.com/johnpapa/angular-styleguide#named-vs-anonymous-functions) for controllers/factories/etc
- [controllerAs Syntax](https://github.com/johnpapa/angular-styleguide#controlleras-with-vm)

This will make everything more clear and jumping into frontend development much easier for new devs.

Already started with this, todolist:

**Core**
- [x] Configs
- [x] Controllers
- [x] Directives
- [x] Filters
- [x] Services

**Contacts**
- [x] Configs
- [x] Controllers
- [x] Services

**Messages**
- [x] Configs
- [x] Controllers
- [x] Directives
- [x] Services

**Offers**
- [x] Configs
- [x] Controllers
- [x] Services

**Search**
- [x] Configs
- [x] Controllers

**Pages**
- [x] Configs
- [x] Controllers

**Statistics**
- [x] Config
- [x] Controllers
- [x] Services

**Users**
- [x] Configs
- Controllers
  - [x] Authentication (tests, too!)
  - [x] Avatar-editor
  - [x] Confirm email
  - [x] Password (tests, too!)
  - [x] Profile
  - [x] Profile-edit
  - [x] Settings
  - [x] Signup
- Directives
  - [x] Monkeybox
  - [ ] Avatar
- [x] Services

**Users/Tests**
- [x] authentication.client.controller.tests.js
- [x] password.client.controller.tests.js

simison:
Only this needs sorting out: https://github.com/Trustroots/trustroots/blob/master/modules/users/client/directives/tr-avatar.client.directive.js

PR #387, opened 2016-08-12
Add eslint-angular rules

simison:
Lints client side javascript files with [Angular rules][1] using a
separate gulp task, because eslint [doesn’t support glob based
configurations][2] just yet.

[1]: https://github.com/Gillespie59/eslint-plugin-angular#rules
[2]: https://github.com/eslint/eslint/issues/3611

See Angular best practises #284

### DIFF
--- a/.eslintrc-angular.js
+++ b/.eslintrc-angular.js
@@ -1,0 +1,45 @@
+module.exports = {
+  plugins: [
+    'angular'
+  ],
+  rules: {
+    'angular/component-limit': 0,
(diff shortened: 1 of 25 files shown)


simison:
@nicksellen LGTY?

nicksellen:
Yup, the other approach would be putting eslint config files in the client directories, as it will use the local rules, but then that would mean lots of duplicated files. This approach with gulp looks nice :) (as gulp chooses which files to pass through).

PR #390, opened 2016-08-15
Send recurring reminders to complete signup

simison:
Sends 3 reminder emails. First one already after 4 hours of signup if user didn't confirm their email. Next two in 2 day intervals.

The email template/text is the same for all 3 emails.

### DIFF
--- a/modules/users/server/controllers/users/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users/users.profile.server.controller.js
@@ -359,6 +359,8 @@ exports.update = function(req, res) {
       delete req.body.resetPasswordToken;
       delete req.body.resetPasswordExpires;
       delete req.body.additionalProvidersData;
+      delete req.body.publicReminderCount;
+      delete req.body.publicReminderSent;
 
       // Merge existing user
       var user = req.user;
@@ -632,6 +634,10 @@ exports.sanitizeProfile = function(profile, authenticatedUser) {
   delete profile.emailToken;
   delete profile.password;
   delete profile.salt;
+
+  // This information is not sensitive, but isn't needed at frontend
+  delete profile.publicReminderCount;
+  delete profile.publicReminderSent;
 
   return profile;
 };
--- a/modules/users/server/jobs/user-finish-signup.server.job.js
+++ b/modules/users/server/jobs/user-finish-signup.server.job.js
@@ -86,7 +86,7 @@ module.exports = function(job, agendaDone) {
                   publicReminderCount: 1
                 }
               },
-              function(err, user) {
+              function(err) {
                 if (err) {
                   console.error('Failed to mark user\'s reminder sent.');
                 }
--- a/modules/users/server/jobs/user-finish-signup.server.job.js
+++ b/modules/users/server/jobs/user-finish-signup.server.job.js
@@ -2,8 +2,8 @@
 
 /**
  * Task that checks for users who didn't finish their signup process and thus
- * have `public:false` in their profile. The script sends 3 reminder emails
- * to these users in 3 day intervals, starting 24h after signup.
+ * have `public:false` in their profile. The script sends 3 (configurable)
+ * reminder emails to these users in 2 day intervals, starting 4h after signup.
  *
  * Keeps count of reminder emails at user's model.
  */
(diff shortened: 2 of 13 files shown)


modules/users/server/jobs/user-finish-signup.server.job.js
```diff
@@ -0,0 +1,129 @@
...
+
+    // Send emails
+    function(users, done) {
+
+      if (users.length > 0) {
+
+        // Create a queue worker to send reminders in parallel
+        // Process at most 3 reminders at the same time
+        // @link https://github.com/caolan/async#queueworker-concurrency
+        var remindersQueue = async.queue(function (user, reminderCallback) {
```
nicksellen: I would do away with the queue here (now it doesn't actually do the smtp stuff directly), agenda handles queueing, it just does 1 write to db

nicksellen:
Could add a test for the job too, see https://github.com/Trustroots/trustroots/blob/master/modules/core/tests/server/jobs/send-email.server.job.tests.js. Esp as this job is doing some more complex stuff.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -642,6 +644,10 @@ exports.sanitizeProfile = function(profile, authenticatedUser) {
   delete profile.password;
   delete profile.salt;
 
+  // This information is not sensitive, but isn't needed at frontend
+  delete profile.publicReminderCount;
```
nicksellen: Should probably change all these deletes into a whitelist instead at some point, so new params aren't automatically sent to frontend...

modules/users/server/jobs/user-finish-signup.server.job.js
```diff
@@ -0,0 +1,129 @@
...
+        }, 3); // How many reminders to process simultaneously?
+
+        // Start processing reminders for all the found users
+        remindersQueue.push(users);
+
+        // Assign a final callback to work queue
+        // All jobs done, continue
+        remindersQueue.drain = function(err) {
+          if (err) {
+            console.error('Sending signup reminder mails caused an error:');
```
nicksellen: Is it intended that an error here will not cause the job to fail?

nicksellen:
The worker retry logic will currently retry any type of job if it has a network error [1]. I imagine that logic still works fine for the 3 jobs in use now. Not actually sure how it interplays with scheduled repeated jobs though...

[1] https://github.com/Trustroots/trustroots/blob/master/config/lib/worker.js#L47

modules/users/server/jobs/user-finish-signup.server.job.js
```diff
@@ -0,0 +1,129 @@
...
+
+module.exports = function(job, agendaDone) {
+  async.waterfall([
+
+    // Find un-confirmed users
+    function(done) {
+
+      // Ignore very recent messages and look for only older than x seconds
+      var createdTimeAgo = new Date();
+      createdTimeAgo.setSeconds(createdTimeAgo.getSeconds() - moment.duration(firstReminderAfter).asSeconds());
```
nicksellen: seeing as `moment` is available, could be something like `moment().subtract(firstReminderAfter).toDate()` (didn't run it)

modules/users/server/jobs/user-finish-signup.server.job.js
```diff
@@ -0,0 +1,129 @@
...
+        };
+      } else {
+        // No users to send emails to
+        done(null);
+      }
+
+    }
+
+  ], function(err) {
+    if (err) {
```
nicksellen: can just pass `err` into `agendaDone` according to the docs [1], e.g.:

```js
agenda.define('registration email', function(job, done) {
    User.get(job.attrs.data.userId, function(err, user) {
       if(err) return done(err);
       email(user.email(), 'Thanks for registering', 'Thanks for registering ' + user.name(), done);
     });
  });
```

[1] https://github.com/rschmukler/agenda#example-project-structure

modules/users/server/controllers/users/users.profile.server.controller.js
simison: True, these are getting hard to manage.

Added issue; https://github.com/Trustroots/trustroots/issues/402

modules/users/server/jobs/user-finish-signup.server.job.js
simison: It might be just one user's reminder that fails, it should still loop trough the rest of the reminders and try sending them, no?

simison:
Updated tests.

PR #391, opened 2016-08-15
Search members from tribes

simison:
#400 

Work in progress


simison:
Open questions:
- should "Show only members from my tribes" toggle be handled at the frontend (as it's now at PR) or at better the backend. Might be easier for other clients.
- How to handle "Allow only people from my tribes find me" #397

simison:
Been photoshopping today a bit:

image

PR #403, opened 2016-08-29
Update engines

simison:
- MongoDB 3.2
- NodeJS 6v, drops 5 from tests


simison:
Aight, passes CI tests nicely.

Issue #407, opened 2016-09-05
Fix $http error/success is deprecated warnings


simison:
Running `gulp eslint` now warns:
```
modules/users/client/controllers/authentication.client.controller.js
  28:7  warning  $http error is deprecated. Use then or catch instead  angular/no-http-callback
  28:7  warning  $http success is deprecated. Use then instead         angular/no-http-callback

modules/users/client/controllers/confirm-email.client.controller.js
  28:7  warning  $http error is deprecated. Use then or catch instead  angular/no-http-callback
  28:7  warning  $http success is deprecated. Use then instead         angular/no-http-callback

modules/users/client/controllers/password-forgot.client.controller.js
  25:7  warning  $http error is deprecated. Use then or catch instead  angular/no-http-callback
  25:7  warning  $http success is deprecated. Use then instead         angular/no-http-callback

modules/users/client/controllers/password-reset.client.controller.js
  25:7  warning  $http error is deprecated. Use then or catch instead  angular/no-http-callback
  25:7  warning  $http success is deprecated. Use then instead         angular/no-http-callback

modules/users/client/controllers/profile-edit-account.client.controller.js
  92:7  warning  $http error is deprecated. Use then or catch instead  angular/no-http-callback
  92:7  warning  $http success is deprecated. Use then instead         angular/no-http-callback

modules/users/client/controllers/profile-edit-networks.client.controller.js
  53:7  warning  $http error is deprecated. Use then or catch instead  angular/no-http-callback
  53:7  warning  $http success is deprecated. Use then instead         angular/no-http-callback

modules/users/client/controllers/signup.client.controller.js
  94:7  warning  $http error is deprecated. Use then or catch instead  angular/no-http-callback
  94:7  warning  $http success is deprecated. Use then instead         angular/no-http-callback
```

From https://docs.angularjs.org/api/ng/service/$http:

>The `$http` legacy promise methods `success` and `error` have been deprecated. Use the standard `then` method instead. If [`$httpProvider.useLegacyPromiseExtensions`](https://docs.angularjs.org/api/ng/provider/$httpProvider#useLegacyPromiseExtensions) is set to false then these methods will throw [`$http/legacy`](https://docs.angularjs.org/error/$http/legacy) error.

See also https://github.com/meanjs/mean/issues/1479

rpaggi:
Hello, i know how fix this, i'll act in this issue. When i finish i send my PR.

simison:
Excellent, thanks!

[...]

PR #409, opened 2016-09-13
Changes to attend the demand of issue #407

rpaggi:
Hello,
I change the code to fix the warnings messages related on issue #407


simison:
I gave this a spin and it works + looks good. Thanks!

PR #423, opened 2016-10-09
added emailTemporary to fix (sometimes) failing test

mrkvon:
Otherwise the `_user2.emailTemporary` equals `_user1.emailTemporary`.

Then in [`users.update`, the user2 matches and can be returned](https://github.com/Trustroots/trustroots/blob/master/modules/users/server/controllers/users/users.profile.server.controller.js#L294-L299), the [check on line 303](https://github.com/Trustroots/trustroots/blob/master/modules/users/server/controllers/users/users.profile.server.controller.js#L303) passes and the test fails with [`Error: expected 403 "Forbidden", got 200 "OK"`](https://travis-ci.org/mrkvon/trustroots/builds/166139128).


simison:
:thumbsup:

Issue #421, opened 2016-10-06
Show reply rate in profile


simison:
"Replied to % of messages."

Same rules apply as to _"Show average time to reply in profile"_ #420

---

It could look like this:

```html
<div class="profile-sidebar-section text-muted">
  Reply rate 90%.<br>
  Replies within a day.
</div>
```

![image](https://cloud.githubusercontent.com/assets/87168/20387253/e2ca2b5c-acc7-11e6-875c-a086e0d56b6d.png)

mrkvon:
implemented with #436 and #448

mrkvon:
implemented with #436 and #448

Issue #420, opened 2016-10-06
Show average time to reply in profile


simison:
Include only threads:
- from the past 6 months (moving window)
- where user didn't initiate the message thread

Show some sort of fuzzy indication "replies within a day" etc

Value certainly requires some sort of caching.

The reasoning for this features is that 
1. people would tend to write to users who have better reply time
2. people would want to keep their reply time short publicly
3. highlights unresponsive users

Related to _show reply rate in profile_ (#421)

---

It could look like this:

```html
<div class="profile-sidebar-section text-muted">
  Reply rate 90%.<br>
  Replies within a day.
</div>
```

![image](https://cloud.githubusercontent.com/assets/87168/20387253/e2ca2b5c-acc7-11e6-875c-a086e0d56b6d.png)

mrkvon:
Here is an attempt to mongoDB query, which returns amount of non-replied threads by user, amount of replied threads by user and average reply time. It works, the questions are: Does it work _correctly_? Can it be _simplified_? It fetches all the messages of a given user and then transforms them.

Detail:
- match all messages in which user is involved
- sort by `created`
- group by the thread and note who is the sender (`fromMe` boolean)
- match only threads started less than 180 days ago and started by the other user
- unwind the messages and then group the sent vs. received messages separately
- from each thread keep only the oldest message received and sent
- put the thread back together and count replyTime when available
- group, count amount of unreplied & replied threads and average replyTime

```
db.messages.aggregate([
{
  $match: {
    $or: [
      {userFrom: ObjectId("57f6ffc99838fb3ecdc94047")},
      {userTo: ObjectId("57f6ffc99838fb3ecdc94047")}
    ]
  }
},
{
  $sort: {created: 1}
},
{
  $group: {
    _id: {$cond: [
      {$eq: ["$userFrom", ObjectId("57f6ffc99838fb3ecdc94047")]},
      "$userTo",
      "$userFrom"
    ]},
    messages: {
      $push: {
        fromMe: {$cond: [
          {$eq: ["$userFrom", ObjectId("57f6ffc99838fb3ecdc94047")]},
          true,
          false
        ]},
        created: "$created"
      }
    }
  }
},
{
  $match: {
    "messages.0.fromMe": false,
    "messages.0.created": {$gte: new Date(Date.now()-180*24*3600*1000)}
  }
},
{
  $project: {
    "messages": 1,
    "id": "$_id",
    "_id": 0
  }
},
{
  $unwind: "$messages"
},
[...]

PR #424, opened 2016-10-09
user reply rate [finished, abandoned]

mrkvon:
see #420 and #421 
- The `replyRate` and `replyTime` are currently served as a part of `GET /api/users/:username` JSON response
- The information is currently displayed in user profile in a form `Replied to {number}% in {time} on average`
  - `{time}` has values `less than a day`, `1 day`, `{2-7} days`, `1 week`, `{2-5} weeks`, `more than a month`
- The above is a first try and is open to suggestions for improvement
### TODO:
- [x] test + implement reading user's reply rate and time from database
- [x] wired into profile controller
- [x] basic view in client
- [x] rebase when #423 is merged to fix failing CI test
- [x] functional tests (testing server routes)
- [x] decide and implement a proper place for results within api
- [x] decide/implement caching the results somewhere
- [ ] review, fix, improve if applicable
- [ ] mean or median?
### maybe TODO:
- show absolute numbers ?
- choose a proper place for the `user-reply-rate.server.controller` and its tests
  - is the place of the new files ok?
- wire the user-reply-rate to the routes and api properly
  - is the agenda calling in messages.server.controller ok?
  - is the api output acceptable? (currently the /api/users/username has parameters i.e. `replyRate: '23%'` and `replyTime: '3 weeks'`); it can be changed to numbers if preferred.
- client tests (maybe unnecessary?)
- decide and implement a better UI design (should the data presentation in client be different?)

### DIFF
--- a/modules/messages/server/controllers/messages.server.controller.js
+++ b/modules/messages/server/controllers/messages.server.controller.js
@@ -17,7 +17,8 @@ var _ = require('lodash'),
     agenda = require(path.resolve('./config/lib/agenda')),
     Message = mongoose.model('Message'),
     Thread = mongoose.model('Thread'),
-    User = mongoose.model('User');
+    User = mongoose.model('User'),
+    ObjectId = mongoose.Types.ObjectId;
 
 
 /**
@@ -308,17 +309,26 @@ exports.send = function(req, res) {
           res.json(message);
 
           // ...and run agenda job to update sender and receiver reply rates
-          // naive, no check whether it is necessary
-          // TODO maybe check if the message is first or first reply in thread
-          // if first, update only receiver
-          // if first reply, update only sender
-          agenda.now('update reply rate',
-            { userId: String(message.userFrom._id) });
-          // we schedule updating reply rate of receiver later
-          // give the receiver opportunity to reply fast
-          // without her reply rate going down
-          agenda.schedule('in 6 hours', 'update reply rate',
-            { userId: String(message.userTo._id) });
+          //
+          // find out which is the position of the message
+          exports.positionInThread(message, function (err, position) {
+            switch (position) {
+              case 'first':
+                // if first, update only receiver
+                // we schedule updating reply rate of receiver later
+                // give the receiver opportunity to reply fast
+                // without her reply rate going down
+                agenda.schedule('in 6 hours', 'update reply rate',
+                  { userId: String(message.userTo._id) });
+                break;
+              case 'firstReply':
+                // if first reply, update only sender
+                agenda.now('update reply rate',
+                  { userId: String(message.userFrom._id) });
+                break;
+              default:
+            }
+          });
 
           return;
         });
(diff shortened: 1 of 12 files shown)


modules/users/client/views/profile/profile-view-basics.client.view.html
```diff
@@ -15,6 +15,23 @@ <h4 class="profile-username">@{{ profileCtrl.profile.displayUsername || profileC
...
+        <span ng-bind="profileCtrl.profile.replyRate"></span>
+      </span>
+      <span ng-if="profileCtrl.profile.replyTime">
+        in
+        <span ng-bind="profileCtrl.profile.replyTime"></span>
+        on average
+      </span>
+      <span ng-if="!profileCtrl.profile.replyRate && !profileCtrl.profile.replyTime">
+        Reply rate not available
+      </span>
```
simison: "Reply rate not available" can be removed, just show nothing in this case.

modules/users/client/views/profile/profile-view-basics.client.view.html
```diff
@@ -15,6 +15,23 @@ <h4 class="profile-username">@{{ profileCtrl.profile.displayUsername || profileC
       <span ng-if="profileCtrl.profile.gender" ng-class="{ 'text-capitalize': !profileCtrl.profile.birthdate }">{{ profileCtrl.profile.gender }}.</span>
     </div>
 
+    <!-- Reply Rate -->
+    <div class="profile-sidebar-section">
+      <i class="icon-fw icon-reply text-muted"></i>
+      <span ng-if="profileCtrl.profile.replyRate">
+        Replied to
+        <span ng-bind="profileCtrl.profile.replyRate"></span>
+      </span>
```
simison: Use [one-time binding](https://docs.angularjs.org/guide/expression#one-time-binding) here so we don't get extra watchers for something that's not gonna change.

simison:
Regarding caching we could merge this without caching feature and build it once we would use it at search feature for sorting results, or if loading profiles turns out to be damn slow because of this.

modules/users/client/views/profile/profile-view-basics.client.view.html
mrkvon: :heavy_check_mark:

modules/users/client/views/profile/profile-view-basics.client.view.html
```diff
@@ -15,6 +15,17 @@ <h4 class="profile-username">@{{ profileCtrl.profile.displayUsername || profileC
       <span ng-if="profileCtrl.profile.gender" ng-class="{ 'text-capitalize': !profileCtrl.profile.birthdate }">{{ profileCtrl.profile.gender }}.</span>
     </div>
 
+    <!-- Reply Rate -->
+    <div class="profile-sidebar-section">
+      <i class="icon-fw icon-reply text-muted"></i>
```
simison: Icon can be removed here.

mrkvon:
Unfortunately it seems that the counting reply rate slows the response down quite a lot. I haven't done any benchmark, but just from the observation `GET /api/users/:username`from localhost on my (rather slow) computer:

- 50-150 ms without replyRate+Time

get-without-reply

- 500+ ms with replyRate+Time


get-with-reply

How does it compare? Is it acceptable? I doubt that.
But maybe the production server is fast enough so the _counting time << network time_?

Thoughts?

i might write a simple benchmark

modules/users/client/views/profile/profile-view-basics.client.view.html
mrkvon: done.

(Seems that `::(profileCtrl.profile.replyRate || profileCtrl.profile.replyTime)` is necessary.)

(btw fyi my experience in angular is low)

simison: Yup, so `::` checks for the variable only once. Without it, Angular will check set a watcher to look up on every digest if those variables changed. Makes the app faster if we know we're not gonna change that variable in our view.

simison:
Yeah, those 5 sec loading times are unacceptable, definitely needs caching. How, that's a good question.

Added +-1 about search.

I'd approach this by storing `replyRate` and `replyTime` at the profile, but also storing pre calculated `expires` date. —> Agenda job goes and re-calculates expired values at the background.

Otherwise, when message gets sent —> check if it was first reply to the thread —> use above Agenda job again to calculate fresh values.

mrkvon:
@simison 

i'll try to incrementally implement the storing in the profile and then we'll see. : )

modules/messages/server/controllers/messages.server.controller.js
```diff
@@ -303,8 +304,23 @@ exports.send = function(req, res) {
...
           }
 
-          // Finally return saved message
-          return res.json(message);
+          // Finally return saved message...
+          res.json(message);
+
+          // ...and run agenda job to update sender and receiver reply rates
+          // naive, no check whether it is necessary
+          // TODO maybe check if the message is first or first reply in thread
```
simison: >"TODO maybe check if the message is first or first reply in thread"

Yep!

modules/messages/server/controllers/messages.server.controller.js
```diff
@@ -303,8 +304,23 @@ exports.send = function(req, res) {
             return done(err);
           }
 
-          // Finally return saved message
-          return res.json(message);
+          // Finally return saved message...
+          res.json(message);
+
+          // ...and run agenda job to update sender and receiver reply rates
```
mrkvon: When the sender's message is the _first reply_, her statistics need to change. (Her reply rate should go up.) Receiver's statistics don't need to change in this case.

mrkvon:
@simison 
The main work is done; tests are written for functions and API.
Before finishing i'd appreciate feedback on the things written in **maybe TODO** on top of the discussion.
Otherwise it can be considered finished. I'll just probably reread the code to improve aesthetic details.
What do you think?

mrkvon:
Wondering about psychology of reply rates:

How much does the reply time shown in profile support higher and faster replying and how much does it become a self-fulfilling prophecy for the future?

_Average reply rate_ can very easily go bad by replying very late to one message, while normally replying very quickly. _Median reply rate_ is much more optimistic in this case (and in fact it says more about the usual behaviour, ignoring the late exceptions). It is more expensive to count, but better for positive reinforcement. (Median is a value in the middle.)

_More expensive_ means:
1. mongodb `$group` can output average value of a field. Mean is not supported.
2. in `$group` all the `replyTime` values of user will have to be collected with `$push`
3. median has to be counted by sorting the array of `replyTime` values and picking 1 or average of 2 values in the middle.

Is there a better way?

chmac:
I would suggest we think about performance optimisation later and go with what makes most sense right now from a data point of view only. The point about median vs mean is a very good one and definitely warrants further thought. I'd suggest we research what other people are doing and see if there's any good advice online.

mrkvon:
Now the `medianReplyRate` is also implemented (but the average `replyRate` is displayed in the client until this is decided).

Should we also show the absolute numbers like warmshowers do? [(#420)](/Trustroots/trustroots/issues/420#issuecomment-252477108) It would be easy to implement.
It can be an interesting information (to see who is overwhelmed by requests and who is not). Or maybe it is too much and too complex. :tropical_fish:

chmac:
@mrkvon I'm going to be hanging out with @simison this week (Thursday to Saturday). Maybe we can schedule a call to discuss, maybe loop in @Ghana also. Could plan / agree a time on slack...

chmac:
I just did some fast research, read [this](https://www.airbnb.com/help/article/430/how-are-my-response-rate-and-response-time-calculated), [this](https://support.rover.com/hc/en-us/articles/203062560-What-do-the-statistics-on-my-profile-page-mean-) and [this](https://www.facebook.com/help/1625715647696553?helpref=faq_content).

Looks like there's a few trends:

* Only consider the last 10 messages
* OR the last 30 days, whichever is the larger (so 10 messages or 30 days if >10 messages in the last 30 days)
* Track response rate (percentage) alongside average reply time
  - Calculate both from the same set of messages
* Call these Response Rate and Response Time

mrkvon:
This PR is probably not going to be used in production. Improvement upon this is #436 and (other PR to be created).

Issue #434, opened 2016-10-30
Refactor for arbitrary stats backends


chmac:
Before we tackle #432 I think it makes sense to refactor the `influx.writePoint()` API into something generic which supports arbitrary backends. @simison and I put some thought into it, and I wrote up an implementation suggestion.

NOTE: The [pseudo-code](https://github.com/Trustroots/trustroots/compare/stats-api-brainstorming) I've written is 100% untested, there are almost certainly syntax errors, logic errors, and all kinds of other horrendous problems in it. It's purely for illustration purposes, although the general pattern might make sense. Some of the comments and variable names may also be useful during the actual refactor.

PR #436, opened 2016-11-01
Collect message/reply statistics

mrkvon:
see #420, #421, #424

As of now, the `updateMessageStat(message)` of `message-stat.server.service` needs to be called after the new `Message` is saved to MongoDB. _Why?_ It doesn't fetch the information from the new `Message`, but from the database. It uses the `Message` object provided just to fetch `userFrom` and `userTo` ids. _Well, why again?_ This way we can ensure idempodence and we update the stats only after the message is saved.

Goal:

- [x] a new collection in database `MessageStats`; purpose: count user reply statistics more easily (store statistical info about first message, first reply, time it took from first message to first reply)
- [x] when a message is sent, create/update `MessageStat` document
  - [x] service done
- [x] write a script to fill historical `MessageStats` (only last 3 months)

### The script
The script is named [create-missing-message-stats.js](/mrkvon/trustroots/blob/aa323f9acc1f2b4413d5fbd18e26c9289f1b61ab/scripts/db-maintenance/create-missing-message-stats.js). It can be run i.e. inside trustroots main folder with a command
```bash
NODE_ENV=development node scripts/db-maintenance/create-missing-message-stats.js
```
Replace `development` with an environment of your choice (test, development, production).
MongoDB must be up.
The script will iterate only through the last 90 days of messages. If you want to iterate through different or all messages, change the script as it is indicated there.


modules/messages/server/models/thread-stat.server.model.js
```diff
@@ -0,0 +1,47 @@
...
+  firstReplyCreated: {
+    type: Date,
+    default: null
+  },
+  firstReplyLength: {
+    type: Number,
+    default: null
+  },
+  // how long did it take the receiver of the first message to reply to it
+  firstReplyTime: {
```
chmac: I think `firstReplyTime` is a little misleading, it sounds like it's the time of the first reply, maybe `timeToFirstReply` would be clearer?

chmac:
@mrkvon Great start, thanks for digging into this.

I have difficulty following the logic flow of the code. Specifically, I'm unclear on the idea of treating existing threads differently compared to new threads. Can you give a high level overview of the process by which you fill the `ThreadStat` collection?

@simison will be able to comment better on whether this logic belongs in a controller. It doesn't seem like a controller to me, but he has a better overview of how the trustroots code is structured.

Also, if you check out my [implementation suggestion](https://github.com/Trustroots/trustroots/compare/stats-api-brainstorming) for #434, I'd love to see that level of commenting in trustroots. I'm a huge believer in keeping the cognitive load as low as possible for the reader of code. There are probably more lines of comments than code in my branch, and I think that's a pretty good balance.

simison:
>@simison will be able to comment better on whether this logic belongs in a controller. It doesn't seem like a controller to me, but he has a better overview of how the trustroots code is structured.

There's a bunch of code at files called `*.controller` around the codebase, but indeed, trying to move this sorta stuff towards "services".

mrkvon:
> Can you give a high level overview of the process by which you fill the `ThreadStat` collection?

1. Currently, we create a new `ThreadStat` document when a new `Thread` is started. We fill the `firstMessage(Created|Length)` properties of `ThreadStat` during the creation. The `firstReply(Created|Length)` and `timeToFirstReply` remain `null`.

2. When a new `Message`'s `Thread` already exists && it is directed from the receiver to initiator of the first message && the `ThreadStat` document's `firstReply...` properties are `null`, the `Message` should be the first reply, so we update `firstReply...` and `timeToFirstReply`properties.

3. In all the other cases, we do nothing with `ThreadStat`.

We treat `ThreadStat` document based on its previous state (i.e. we use it to find out whether a message is the first reply). The related `Thread` document doesn't have enough information for this (the `userFrom` and `userTo` fields change all the time and there is not enough information about the position of a message). In order to keep it working, we need to preserve the flow of _creating when first message_ > _updating when first reply_ > _ignoring the rest_. If we try to apply the flow to the `Threads` older than this feature, we'll end up in mess or too much complexity. Simply ignoring the older `Threads` and filling their stats by a script is the simplest solution i found.

Does it make sense? Should i explain something more?
There are certainly different ways to do this. My view is necessarily limited/flawed. Can you see a better way?
I'll try to make it clearer in the code.

> @chmac It doesn't seem like a controller to me

In fact i put the code to controller just because it's not model nor router. I'm not even sure what exactly is the domain of controllers. Thanks for pointing it out. If the _place for general control flow functions_ is a service, i'll gladly move it there.

> I'm a huge believer in keeping the cognitive load as low as possible for the reader of code. There are probably more lines of comments than code in my branch, and I think that's a pretty good balance.

Thank you, @chmac for the feedback. I'm in a process of trying to learn to comment more/better and structure the code better. Any suggestions for improving (anything) are much appreciated.

simison:
@nicksellen was suggesting to use services for this sorta stuff; e.g. I called [text-processor](https://github.com/Trustroots/trustroots/blob/master/modules/core/server/controllers/text-processor.server.controller.js) to a controller, whereas it should really be called a service.

I'll update https://github.com/Trustroots/trustroots/wiki/Development regarding this.

Not a biggie, never be afraid to refactor anyway. :-)

chmac:
@mrkvon Thanks for the breakdown, super helpful to understand your approach.

I'm looking at line 54 in `modules/messages/server/controllers/thread-stat.server.controller.js`:

        Message.find({
          $or: [
            { userFrom: message.userFrom, userTo: message.userTo },
            { userFrom: message.userTo, userTo: message.userFrom }
          ]
        }

As I understand it, this retrieves every message in history between 2 users. From that list of messages, we could compute everything we need to know about ThreadStat. Of course, if it's already been computed, it would make sense to skip that process. As I read the current code, this query will be run on every single message send event.

Maybe we could:

* Get the ThreadStat
* If the ThreadStat **does** already exist:
  - Does it have a `timeToFirstReply`?
  - Yes:
    - Nothing to do, move on
  - No:
    - Get all the messages between those two users
    - Scan the list of messages to see if we find a `firstReply`
    - If we do:
      - Update the ThreadStat with the `timeToFirstReply` etc
* If it does **not** exist:
  - Get all the messages between those two users
  - Find the first message
  - Create the ThreadStat
  - Then do the same the search for `firstReply` from above

As I understand it, this approach would be idempotent. We could run the operation on every message send, and if the ThreadStat exists and has all the data, nothing happens. But if any data is missing, we simply get it and fill in the missing data if we can. Then the data should fill in over time, and if we wanted to, we could run a process to run through every thread and run this process to fill in any missing data.

chmac:
Thinking about it further, if we use the approach I outlined above, we should really call this collection `MessageStats` rather than `ThreadStats` and we can probably ignore the `Threads` collection altogether when computing these values. I guess this simplifies the whole thing, because `Threads` are actually quite complex.

simison:
Cool, lots of new stuff here. :thumbsup: Wanna have a look @chmac ?

chmac:
@mrkvon The latest updates look great. I've added some extra comments to help explain what's happening. Would be great if you can check them to be sure that they correctly describe what's happening.

In general, I think the code is much clearer, much easier to follow, and it looks to be idempotent, which is great.

mrkvon:
The goals of this pull request are reached.
The purpose was to create the `MessageStat` model and wire it into the application.
The further goals of #420, #421 will be worked on in a future PR.

simison:
:thumbsup: :thumbsup:

PR #437, opened 2016-11-01
Upgrade node-influx to v5

simison:
See breaking changes:
https://github.com/node-influx/node-influx/blob/master/CHANGELOG.md#2016-10-24-500-alpha4

Most notably rename `writePoint` to `writeMeasurement` to keep it clear
it uses `.writeMeasurement` from node-influx under the hood:
https://node-influx.github.io/class/src/index.js~InfluxDB.html#instance-method-writeMeasurement

### DIFF
--- a/modules/core/server/services/influx.server.service.js
+++ b/modules/core/server/services/influx.server.service.js
@@ -4,7 +4,7 @@
  * Module dependencies.
  */
 var path = require('path'),
-    influx = require('influx'),
+    Influx = require('influx'),
     _ = require('lodash'),
     config = require(path.resolve('./config/config'));
 
@@ -23,12 +23,14 @@ exports.getClient = function(callback) {
     return callback(new Error('No InfluxDB configured.'));
   }
 
-  // check passed so we send configuration to influx()
-  callback(null, influx(config.influxdb.options));
+  // Init Influx client with configuration
+  var client = new Influx.InfluxDB(config.influxdb.options);
+
+  callback(null, client);
 };
 
 /**
- * Write point to InfluxDB
+ * Write measurement to InfluxDB
  *
  * fields - object of field key: value pairs. To save to influxdb.
  *   - key in camelCase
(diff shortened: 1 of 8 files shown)


simison:
@mrkvon wanna have a look + test?

Note that node-influx's `writeMeasurement` expects an array of points, e.g.:

```
influx.writeMeasurement('perf', [
  {
    fields: { host: 'box1.example.com' },
    tags: { cpu: getCpuUsage(), mem: getMemUsage() },
  }
])
```

modules/core/server/services/influx.server.service.js
```diff
@@ -44,12 +46,12 @@ exports.getClient = function(callback) {
  * influxdb (camelCase)
  * @param {Object} fields - key: value pairs will be saved in influxdb as field
  * key: field value
- * @param {number|Date} [fields.time=new Date()] - time of measurement with precision ms
+ * @param {number|Date} [fields.time=new Date()] - time of measurement with precision nanosecond
```
simison: I updated this based on this comment from changelog: "_write methods now write in nanosecond precision by default_". Any idea if I should change any code regarding this?

mrkvon: Based on the description of [writePoints](https://node-influx.github.io/class/src/index.js~InfluxDB.html#instance-method-writePoints) method, passing javascript `Date` is fine. However, the messageToInflux service is indeed using timestamp (number) with millisecond precision (see the [related test](/Trustroots/trustroots/blob/upgrade-node-influx-v5/modules/messages/tests/server/message-to-influx.server.service.tests.js#L240)). It's not clear to me how the defaults of node-influx work now.

Note1: `influx.writeMeasurement` uses `writePoints` method internally.
Note2: there is a lot of space for improving that linked test, i can see.
Note3: To remove confusion, we could require `Date` type for time or [explicitly provide the precision](/Trustroots/trustroots/blob/upgrade-node-influx-v5/modules/core/server/services/influx.server.service.js#L79) as the 3rd parameter of [`writeMeasurement`](https://node-influx.github.io/class/src/index.js~InfluxDB.html#instance-method-writeMeasurement). Node-influx docs [suggest the precision to be as rough as possible](https://node-influx.github.io/typedef/index.html#static-typedef-IWriteOptions) for efficiency. I doubt we need nanosecond precision.

mrkvon: @simison do you prefer me committing the changes directly, or leaving it for you to decide and execute?

This is a pretty general question, really. The first option opens a space for both (a mess and misunderstanding) and (a higher and exciting level of collaboration).

simison:
Thanks for asking; feel totally free to push changes here — I have lots on
my plate already so any help is great.

modules/core/server/services/influx.server.service.js
mrkvon: We could consider limiting the `fields.time` to `Date`, which would remove the confusion (node-influx takes care of that), but force the precision to be only _ms_ or lower. It is possible to [get nanosecond precision in node via `process.hrtime()`](http://stackoverflow.com/a/18197438) and if needed, it would have to be passed as a timestamp. The question is, would it ever be needed?

simison: I think we can force `Date` and let Influx take care of it. If we ever need `ns`, it's gonna be a place for refactoring, which is fine.

PR #448, opened 2016-11-13
Count user's reply statistics

mrkvon:
Counting the reply rate and reply time statistics of a user based on `MessageStat` documents of #436.

This work is done on top of #436. That should be merged first.

### Overview:

For brevity we define age of `MessageStat` as a value of its `firstMessageCreated` property.

1. Provided `userId` and arbitrary `timeNow` (timestamp) we search database for `messageStats` which are between `timeNow` and `timeNow - 90 days` old (`messageStats` of the user within last 3 months)
2. We choose subset of the `messageStats` found:
  - if we found less than 10 `messageStats`, choose all of them
  - else if some of the youngest 10 `messageStats` are older than 30 days, choose the 10 youngest `messageStats`
  - else choose `messageStats` which are younger than 30 days (counted from `timeNow`)
3. Based on the chosen subset above, we count the `replyRate` and `replyTime` of the user
  - if the subset is empty, `replyRate` and `replyTime` should be `null`
  - if there are no replied `messageStats`, `replyRate` should be `0` and `replyTime` `null`
  - if there are some replied `messageStats`, `replyRate` should be `0 - 1` (replied / all messages), `replyTime` should be average (mean) of `timeToFirstReply` of the replied `messageStats` (milliseconds)

### TODO

- [x] write a service which converts replyTime to text (i.e. 23 hours => "within 1 day")
- [x] integrate the whole thing to the user profile routes
- [x] display the statistics in client in user's profile
- [x] comment, review and improve the code
- [x] make sure it doesn't significantly slow down the displaying of profiles
- [x] think about indexes, how smart introducing of indexes could improve performance


simison:
>write a service which converts replyTime to text (i.e. 23 hours => "within 1 day")
 
[Moment.js](http://momentjs.com/docs/#/displaying/from/) could come handy for this.

>display the statistics in client in user's profile

I updated here how this could look like at the sidebar: 
https://github.com/Trustroots/trustroots/issues/421 (feel free to modify)

mrkvon:
## Simple benchmarks :weight_lifting_woman: 

The reply statistics now made it all the way to the profile. (The way still wants some attention. (= comment and review the code))

Here's a simple comparison of querying users' profile data without and with fetching the statistics. The tests were performed on a bit slow netbook.

How to run a benchmark with `curl` is described in #424 discussion.

```
curl http://localhost:3000/api/users/[username]
```

1. 300 users, approx. 50000 messages within last 3 months, approx. 20000 threads

  - __no stats:__ around 16 (11) s / 300 users (average 53 (37) ms), stable responding times
  - __stats:__ around 30 (25) s / 300 users (average 100 (83) ms), responding times 50 - 150 ms depending on number of threads the user received

2. 3000 users, approx. 50000 messages within 2 years, approx. 20000 threads
  - __no stats:__ 2m 37s / 3000 users
  - __stats:__ 2m 55s / 3000 users

3. to be continued...

### Summary

The degree of slowing the request down depends on amount of the threads user received in the period of last 1-3 months. It probably doesn't depend much on total amount of `MessageStat` documents in database (probably thanks to some indexes).

The raising of the response time from 30-40 to 50-150 ms (the latter happens with testing user who received around 130 threads within last 3 months) might be acceptable.

@simison what do you think? Is it too slow or acceptable?

_Note: These tests are approximate. More precision and rigor may be needed._

### Improve?

  - better design of database indexes (to achieve O(1) query times)
  - pre-counting the statistics data (we already tried that, it brings other issues, compare #424 )
  - something else?

mrkvon:
This is still related to #436. There is some very low probability that the `MessageStat` document will be duplicate (having 2 `MessageStat` documents with the same `firstMessageUserFrom` and `firstMessageUserTo` fields)

We perform the check whether the `MessageStat` already exists. If not, we create the `MessageStat`. Problem is that this happens asynchronously and nothing prevents another `MessageStat` from being saved to database between these two events. Putting unique compound index on these two fields is not enough, because the two fields can switch and we still need uniqueness.

### Remedy

- ignore this altogether.

- ignore this partially. There is a very low probability that users will send a first message to each other at the same time. Just put unique index on the two fields in case the same sender sends 2 messages to the same receiver at the same moment.

- introduce a new field to model: a string with ordered `firstMessageUserFrom`+`firstMessageUserTo`. Unique index on this field should prevent duplicates.

- ... ?

mrkvon:
If the #436 was already merged to production, i suggest deleting all the `MessageStat` documents and re-running the script 'scripts/db-maintenance/create-missing-message-stats.js' when this will be finished.
 
i.e. now i fixed some bugs there (used `limit()` instead of `sort()` when sorting `Messages`).

simison:
>If the #436 was already merged to production

It's just on master and not merged to production yet. I was planning to do it today but it can wait a couple days for this to be merged.

mrkvon:
@simison 
i reviewed the code now. I think to get any further i would need new points of view. Generally it should be ok and working.

There still is the very low probability of creating duplicate `MessageStats` (per `Thread`) as described above, but that may not be a big issue. The possible effects of that would not do any harm probably.

simison:
Let's go ahead with risk of ending up with double entries and observe it a bit after a week/two. If it happens too much, we'll work the problem then.

Issue #455, opened 2016-11-20
Mongo indexes on Messages


mrkvon:
Probably related to #59 

Are there any custom [indexes](https://docs.mongodb.com/manual/indexes/) on `Message` and `Thread` mongoose models? I can't see any, but it is possible that they're defined elsewhere (?).

Indexing `userFrom` and `userTo` fields may improve performance (a lot) and also keep a stable query time as the number of messages will grow. Without index, the database has to scan all the documents of that type when queried on `userFrom` and `userTo` fields.

The indexes come with a cost: need for further storage space and computation power when creating and updating the index. Benefit should outweigh the cost though. (I don't know details.)

I would suggest separate indexes on `userFrom` and `userTo`. There could be another compound index on both of these (useful when searching a specific thread).

simison:
Chatted about this => merged the PR.

PR #456, opened 2016-11-20
added indexes to Message and Thread models

mrkvon:
#455 

__Edit:__

Now, I'm not sure if this is not too rushed. There are many [things](https://docs.mongodb.com/manual/core/data-model-operations/#data-model-indexes) - [to](https://docs.mongodb.com/manual/applications/indexes/) - [consider](https://docs.mongodb.com/manual/indexes/#additional-considerations) when designing indexes.

Were there any attempts to achieve this in the past? Is there any MongoDB indexing pro?

I.e. now i commented out the compound indexes on `userFrom` and `userTo` based on [this article](https://docs.mongodb.com/manual/core/index-intersection/#index-intersection-and-compound-indexes).

### DIFF
--- a/modules/messages/server/models/message.server.model.js
+++ b/modules/messages/server/models/message.server.model.js
@@ -41,12 +41,6 @@ var MessageSchema = new Schema({
   }
 });
 
-// This index is useful when searching Messages by both userFrom and userTo
-// Is probably not necessary thanks to Index Intersection
-// https://docs.mongodb.com/manual/core/index-intersection/#index-intersection-and-compound-indexes
-//
-// MessageSchema.index({ userFrom: 1, userTo: -1 });
-
 MessageSchema.plugin(mongoosePaginate);
 
 mongoose.model('Message', MessageSchema);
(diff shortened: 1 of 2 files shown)


modules/messages/server/models/message.server.model.js
```diff
@@ -39,6 +41,12 @@ var MessageSchema = new Schema({
   }
 });
 
+// This index is useful when searching Messages by both userFrom and userTo
+// Is probably not necessary thanks to Index Intersection
+// https://docs.mongodb.com/manual/core/index-intersection/#index-intersection-and-compound-indexes
+//
+// MessageSchema.index({ userFrom: 1, userTo: -1 });
+
```
simison: I think this would be useful for `$and` queries but since we have only `$or` queries everywhere, let's drop these lines out. After that I'll merge.

mrkvon:
@simison 

the commented compound indexes are now removed completely

simison:
:thumbsup:

PR #463, opened 2016-11-28
Simplify Influxdb service API and fix errors around tests

simison:
Always expect`fields.time` to be a `Date` and turn it to integer (ms
precision)

### DIFF
--- a/modules/core/server/services/influx.server.service.js
+++ b/modules/core/server/services/influx.server.service.js
@@ -6,7 +6,8 @@
 var path = require('path'),
     Influx = require('influx'),
     _ = require('lodash'),
-    config = require(path.resolve('./config/config'));
+    config = require(path.resolve('./config/config')),
+    log = require(path.resolve('./config/lib/logger'));
 
 /**
  * Get InfluxDB Client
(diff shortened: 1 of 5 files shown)


modules/core/server/services/influx.server.service.js
```diff
@@ -88,5 +111,7 @@ exports.writeMeasurement = function(measurementName, fields, tags, callback) {
       }
     ]);
 
+    callback();
```
simison: Is this causing tests to hang up? https://travis-ci.org/Trustroots/trustroots/builds/179439231

modules/messages/tests/server/message-to-influx-functional.server.service.tests.js
```diff
@@ -20,8 +20,9 @@ var reachEventEmitter = new EventEmitter();
 
 // mocking the influxdb.writeMeasurement()
 var influxdb = {
-  writeMeasurement: function (measurement, fields, tags) {
+  writeMeasurement: function (measurement, fields, tags, callback) {
     reachEventEmitter.emit('reachedInfluxdb', measurement, fields, tags);
+    if (callback) callback();
```
simison: Is this causing tests to hang up? https://travis-ci.org/Trustroots/trustroots/builds/179439231

Removing it yelds some errors, tho.

simison:
Thanks @mrkvon for helping out with this!

Issue #393, opened 2016-08-16
Send another notification about unread messages


simison:
Currently only one notification message is sent out, there should be at least two more between next 24 hours, if user doesn't read them. These timings should be easily adjustable.

In future there could be other types of notifications:
- SMS #159 
- Desktop #235 
- ~~Facebook #369 (requires #250 first)~~ (done)
- Mobile push notifications

All these should probably be piped trough same service. I don't think it's necessary to make a huge rewrite of the bit that sends these notifications, but it's good to keep all this in mind.

nicksellen:
I'm not sure this is needed. I'm not aware of any other sites that sends multiple notifications for one message.

Imagine I have had a conversation with someone, then close the browser and go out. Then they send a short message, like _ok, great thanks, see you then_. I read it on my phone via the email and it doesn't need a reply. But it will then keep sending me the message until I go to the web interface? Seems a bit much.

simison:
Hmm, a valid point. Let's leave this on for thinking a bit longer.

simison:
Actually, this would be needed if we'd have hosting requests instead of just messages.

nicksellen:
Yeah, some things need reply. In letsmake a booking in _open_ state would represent something that could do with repeated reminders.

simison:
@nicksellen actually, perhaps one notification could be sent about messages which are first in the thread? There's only small change that there's no need to react to those messages and we'd annoy smaller portion of people and make bigger amount of people happy because their first-contact messages won't go unotified.

Perhaps there could be "don't notify me about new messages in this thread" button anyway, though.

mrkvon:
@simison you collected your previous effort to the branch `remind-about-unread-messages`, is that correct? I'm slowly but certainly starting on it now.

mrkvon:
The first attempt to basic implementation is in #470 

It is a bit naive, i.e. not caring whether the message is first in a thread or further. It will send the reminder after 10 minutes passed and after 24 hours passed. The text can be different for the first vs further reminders (not implemented yet).

Can we specify in more detail how this feature should behave?

@nicksellen , @simison What do you think? How do you want this to work?

simison:
@mrkvon great, looks like a good start!

So yeah, further notifications only on threads which don't have any replies for reasons mentioned in discussion above. 1st notification to all the unread messages in the thread.

We'll probably need some limit so that when this will go to production, we don't send 2nd notification about unread messages sent year ago — wouldn't make much sense.

[...]

PR #470, opened 2017-03-05
Further reminders about unread messages

mrkvon:
Implements #393 

Send multiple notifications about unread messages.

We specify reminders in config:
```js
[{ minutes: 10 }, { hours: 24 }, ...]
```
and we send the reminders when the specified time since the last unread message has passed.

We send the further reminders for the messages in unreplied threads only. (Decide: Send for the first message in thread only?)

- [x] send those further notifications for the unreplied threads only
- [x] send different messages when the reminder is first vs further.
- [x] remove the time of last notification from Message schema and from migration (if we keep this approach, it's not needed)
- [x] set notificationCount to config notifications length where previously was notified: true
- [x] do the migration of messages by streaming; there are many of them.
- [x] write and implement pending tests
- [x] comment and refactor where necessary
- [ ] edit to conform with results of further discussion:
    - [ ] group messages together or send just the first one in a single 2nd notification
    - [ ] change the subject of the second notification (when decided)
- [x] rebase against master and resolve conflicts


mrkvon:
The subject of the second and further email notification now is

```
You still have unread messages from <Display Name> on Trustroots
```
Compare to the first notification:
```
<Display Name> wrote you from Trustroots
```

I'll change it on request. I thought about i.e. saying how many unread messages, change singular/plural etc. Or complete rewording?

simison:
I feel something more alerting as 2nd notification subject would get better open rates, like:

> _Don't leave <Name> hanging, reply now!_

@guaka @chmac @nick thoughts?

config/lib/worker.js
```diff
@@ -42,7 +42,7 @@ exports.start = function(options, callback) {
 
     // Schedule job(s)
 
-    agenda.every('5 minutes', 'check unread messages');
+    agenda.every('30 seconds', 'check unread messages');
```
simison: Oh yea, I think I had this change somewhere as well just for development.

You recon it would be good to set it this low anyway?

chmac:
I'd guess that something like "chmac is still waiting for a reply on Trustroots" would get better open rates, but @simison I'd suggest we (meaning you!) ask Simona and follow her advice... :-)

config/lib/worker.js
mrkvon: @simison It is actually just your testing code thoughtlessly merged.

modules/messages/server/jobs/message-unread.server.job.js
```diff
@@ -21,27 +21,81 @@ var _ = require('lodash'),
...
+   */
+
+  var remappedConfig = _.reverse(_.map(sortedConfig, function (value, index) {
+    return {
+      order: index, // nth notification
+      delay: value // when to send the notification
+    };
+  }));
+
+  // TODO check that the config has a proper format
```
mrkvon: Thanks. Yes, here i was hoping momentjs would have the `isValid` function.

config/lib/worker.js
mrkvon: reverted

modules/messages/server/jobs/message-unread.server.job.js
```diff
@@ -21,27 +21,81 @@ var _ = require('lodash'),
...
+
+  // sendUnreadMessageReminders(remindersConfig[0], 0, agendaDone);
+  async.eachSeries(remappedConfig, sendUnreadMessageReminders, agendaDone);
+};
+
+/**
+ * // send the reminders about unread messages to various services
+ * @param {Object} reminder - info about the reminder
+ * @param {Object} reminder.delay - a time object as momentjs would expect it, i.e. { 'minutes': 10 }
```
mrkvon: explain

modules/messages/server/jobs/message-unread.server.job.js
```diff
@@ -21,27 +21,81 @@ var _ = require('lodash'),
...
+   * that way we avoid duplicates when more reminders are pending
+   * i.e. when notificationCount is 0 and 24 hours already passed (maybe due to some agenda outage)
+   * we run the 2nd reminder first, the notificationCount is set to 2
+   * and the 1st reminder is not sent at all
+   *
+   * otherwise both reminders would be sent at the same time o_O
+   *
+   * hopefully there is not much harm in sending just one reminder in total
+   * this case shouldn't happen too often
+   */
```
simison: 👍  excellent

modules/messages/server/jobs/message-unread.server.job.js
```diff
@@ -21,27 +21,81 @@ var _ = require('lodash'),
...
 
-      // Ignore very recent messages and look for only older than 10 minutes
+      // We want to remind user about messages, which remain unread for more than `timePassed`
+      // But only once for each `timePassed`
       // Has to be a JS Date object, not a Moment object
-      var createdTimeAgo = moment().subtract(moment.duration({ 'minutes': 10 })).toDate();
+      var createdTimeAgo = moment().subtract(moment.duration(timePassed)).toDate();
 
+      //
       Message.aggregate([
```
mrkvon: This code finds all the unread, unnotified messages older than a specified period of time, groups them together by sender+receiver and sends notifications.

If i send 4 messages in 5-minute intervals, it seems to me that 4 separate notifications will be sent in 5-minute intervals starting 10 - 15 minutes later. (Only less than 5 minute messages will be sometimes grouped, because agenda runs every 5 minutes). That seems to be the original behaviour. Is it intended?

If that's true the same would happen 24 hours later with 2nd notifications. In that case i suggest to wait until 24 hours pass since the very last unread not-2nd-time-notified message, and then to send one reminder with all the unnotified unread messages included. Does this make sense?

@simison @chmac @nicksellen how do you imagine the desired behavior of notifications, and 2nd notifications?

chmac: @mrkvon I was thinking about this the other day, and I'd assume we only want to send a second notification for first messages, but not for other messages. For example, if I sent you a messages saying "Sure, see you there", then I can read it in my email, so I don't need to login / view it online, and I definitely don't need a second reminder. But in almost all cases, I guess I need to reply to first messages, so a reminder would be great.

mrkvon: @chmac

> we only want to send a second notification for first messages

currently the 2nd+ reminders are sent for the messages in unreplied threads only. (I.e. the "see you" 2nd notification wouldn't be sent, because that's already part of a bi-directional conversation). That can be changed to the very first message only, if that's preferred.

I thought the messages sent within 10 minutes should be grouped into a single 1st notification. Now i noticed that's not the case (only the messages which are sent between 2 agenda jobs (5 minutes or less) are grouped).
So maybe the same grouping, just for 24-hour intervals would be nice for 2nd notifications. So when the request is scattered among multiple messages, all will be included in one 2nd reminder.

For sure we can as well 2nd-notify the very first message only. The rest of the request messages are already somewhere in the receiver's inbox anyways.

@chmac, @simison Any final decision/specification on this?

chmac: Here's my assumptions.

* A hosting request will almost always be a "first message"
* We really want people to reply to hosting requests
* If you send me a message, and I reply, we've no way to know whether you need to reply or not, maybe I replied with just the word "no" in which case you probably don't need to do anything else!

Very open to flaws in my logic. Maybe there's stuff I'm missing.

In terms of decisions. If you want to move forward immediately, let's send a follow up notification after 20 hours for first messages without a reply only.

I have the idea that we want to start tracking replies within 24 hours and so the 20 hour thing would mean we can say something like "just a few hours to keep your reply rate up" or something like that in the future... :-)

simison: @chmac Yeah, that's what @mrkvon is saying and that was initial idea of the issue. ;-)

What @mrkvon is pondering if in this situation:

- 1st (me)
- 2nd (me, e.g. a week later)
- 3rd (you)

...we should send 2nd notification about that 2nd message. For 3rd message we'll never send 2nd notification anyway.

simison: @mrkvon @chmac So we don't send even 1st notification for read messages in the thread.

Rather than sending "unread notification", should we send "unreplied notification" even when those threads are read? Suppose I check out the message but forget to reply.

chmac: I guess we're going to track "time to reply" from the very first message until the very first reply. I'd suggest that we focus our reply notifications on that. I can imagine that this is also simpler in the code, and less likely to cause issues.

mrkvon: This unresolved discussion disappeared because i edited a comment.

mrkvon:
[The unresolved discussion](https://github.com/Trustroots/trustroots/pull/470#pullrequestreview-26189851) was hidden due to some minor changes in code.

I suggest moving the discussion to an issue and i'll remove the __[WIP]__ from the PR. Perhaps it could still benefit from someone else's review.

_In what circumstances should we send the 1st and 2nd notifications? Unreplied? Unread? First in thread? All in unreplied threads?_

My original issue was slightly different, and now it grew in complexity.

***

Also rewording of the 2nd notifications' email Subject needs to be decided:
- > You still have unread messages from <Display Name> on Trustroots (current, sub-optimal)
- > Don't leave {name} hanging, reply now!
- > {Name} is still waiting for a reply on Trustroots
- > ?

mrkvon:
_(continuing)__

My original issue was:

Let's assume the following events happen (`(number)` shows the timing in minutes):
```msg0(0)---job(1)---msg1(3)---msg2(4)---job(6)---msg3(7)---msg4(8)---job(11)---job(16)---job(21)---...```

Notifications will be sent for `msg0` at `job(11)`, for `msg1&2` at `job(16)`, for `msg3&4` at `job(21)`.

That's 3 notifications for 5 messages, which were sent within 8 minutes.

Wasn't the intended behaviour to send all of them in one notification? What else is the rationale of waiting for 10 minutes before sending the notification?

Now, the same will happen for the 2nd notifications. Just 24 hours later.

I guess sending slightly more notifications will cause not much harm. Resolving this could be a part of another PR, when the specifications of intended behaviour are clearer.

mrkvon:
One last note: This PR assumes that first and further notifications should be treated in a similar way. It uses the same generalized function in both cases.

If you decide that first and further notifications should be treated in a significantly different way, each might need to be implemented separately. In that case it might be better to ignore this PR. I won't mind.

chmac:
@mrkvon Thanks for the update.

I'm starting to lose track of exactly how this works. Bottom line, I think more notifications is probably a good thing. Let's deploy it as it is. Forwards progress beats more detailed discussion. :-)

I think as a general principle, I'd probably remove the 10 minute delay logic as it sounds complicated. If I get 3 emails in 3 minutes for 3 trustroots messages, that's probably fine. I can read them all in my email client. So in general, my suggestion would be favour simplicity over sophistication. The simpler option is probably the better option.

But that's longer term philosophy stuff. For right now, if the PR is ready to deploy, let's deploy. Then we can observe, see what happens, and take any action that might be required after that. :-)

simison:
I'm also a bit lost at the moment — @mrkvon wanna have a quick call this week, gimme a brief how it works and we can then probably merge and publish? :-)

simison:
I checked this trough and it looks good to me. I just changed 2nd email subject to `{Name} is still waiting for a reply on Trustroots`

Let's deploy!

simison:
(Looks like I broke on test with that subject change, my bad, fixing)

Issue #416, opened 2016-10-02
Data questions & KPI


mrkvon:
- Questions we want to answer by looking into data
- Data/events we need to collect to answer those questions
- Top priority: get the data to answer KPI

---
#### Users
- [ ] % of users that have logged in the past week, month, year

#### Messages
- [x] n of messages sent (basically for monitoring if something's broken)
- [ ] % of new threads with response within a day/2 days/etc... based on data within last n weeks
- [ ] % of new threads started in past two weeks without reply
- [ ] % of unread threads within last n weeks without reply
- [x] how many new message threads are answered?
- [x] what is the average response time to the new thread (often a request)?
- [x] are the messages long or short?
- [x] how long are they?
- [ ] a graph to see if someone _ever_ received or sent a reply (possibly also over the past 3 months). This would show user involvement. It would also be nice to limit it to a certain period, e.g. last n months. (by @guaka )

#### References

~
#### Offers
- [ ] % of members have hosting offer
#### Tribes

~
#### Search

~
#### Contacts

~

#### More
Some [great ideas from the early days of CouchSurfing](/jxe/jxe.github.com/blob/master/csmetrics/macro.md) pointed out by @guaka 

---
## Daily stats to collect
- [x] Users count
- [ ] Contacts count
- [ ] Pending contact requests count
- [ ] KPI "Useful message threads" (see below)

---
## Events to collect

_List under event possible meta data to collect._
#### Users
- [x] Signup success
- [x] Signup failed
- [x] Login success
- [x] Login failed
#### Messages
- [x] Message sent
  - First message / first reply / other
  - Time between messages
- [ ] Thread opened
#### References
- [x] Vote message thread up/down
#### Offers
- [ ] Saving hosting offer
  - New / modifying old
  - Previous status if modifying
  - New status
- [ ] Change of hosting states: yes/maybe => no
- [ ] Change of hosting states: no => yes/maybe
#### Tribes
- [x] Join tribe
- [x] Leave tribe
#### Search
- [ ] Search event
  - How many tribe filters active?
  - Continent?
  - n of results
#### Contacts
- [ ] Contact request sent
- [ ] Contact request accepted
  - Time to answer

---
## KPI "Useful threads"

Calculated on weekly basis.
- Response within a day is great, 1 point. no response is 0.
- In between, response after a week is probably half a point
- After 2 weeks it's pointless

Also think about:
- discard messages that are rated negatively
- give x% bonus for threads that are rated positively

[...]

PR #478, opened 2017-03-19
#416 Add statistics for joining/leaving a tribe/tag

simison:


modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -793,10 +794,28 @@ exports.modifyUserTag = function(req, res) {
...
       message += ' ' + ((tag && tag.tribe) ? 'tribe' : 'tag') + '.';
 
-      return res.send({
-        message: message,
-        tag: pickedTag,
-        user: user
+      statService.stat({
+        namespace: 'tag',
+        counts: {
+          count: joining ? 1 : -1
```
simison: @mrkvon can I use this `counts` like this?

(feel free to suggest any better format)

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -793,10 +794,28 @@ exports.modifyUserTag = function(req, res) {
...
+        },
+        tags: {
+          type: tag.tribe ? 'tribe' : 'tag'
+        },
+        meta: {
+          slug: tag.slug
```
simison: @mrkvon These meta values aren't ending up to Stathat, correct?

Issue #394, opened 2016-08-17
Push notifications


simison:
- [x] **1. Add gateway** (Done #491)

  - [FireBase/FCM](https://firebase.google.com/docs/notifications/) — free
  - ~~Pushwoosh.com — [No API for free](https://github.com/Trustroots/trustroots/issues/394#issuecomment-289838189)~~
  - ~~Onesignal.com — not a great privacy policy #485~~
  - ~~[Boxcar.io](https://boxcar.io/) — only 7e/mo~~
- [x] **2. Add desktop notifications** (Done #491, to be done: #498, #497)
- [x] **3. Add mobile push notifications**: https://github.com/Trustroots/trustroots-expo-mobile

simison:
@nicksellen [commented](https://github.com/Trustroots/trustroots/issues/485#issuecomment-289833641):
>Pushwoosh does not support **Remote API** [1] in the free tier [2]. 
>
> [1] http://docs.pushwoosh.com/docs/createmessage
> [2] https://www.pushwoosh.com/pricing/

simison:
Asked Boxcar.io about non-profit/opensource pricing.

nicksellen:
[batch](https://batch.com/pricing) supports unlimited push notifications using the [transactional API](https://batch.com/doc/api/transactional.html).

... but I spent a bit of time reading into push notifications as I did not know much about them really (there is the [Web Push API](https://developer.mozilla.org/en/docs/Web/API/Push_API) now too). I then read the slack chat from before - my past self re-explained it to my current self that had forgotten it all :)

Basically, why not just use [FCM](https://firebase.google.com/docs/cloud-messaging/)? The requirement here is to do push notifications, which FCM is for, works across iOS/Android, _and_ the new Web Push API, and is free with no limits.

All the other fancy startup services are there because they promise to offer something over and above it - perhaps easier APIs, more platforms (kindle fire anyone?), better error handling ... but they are just [using](https://documentation.onesignal.com/v3.0/docs/generate-a-google-server-api-key) [FCM](http://developer.boxcar.io/sdk/android/4.0/#for-android-devices) under the hood anyway, and don't seem dramatically easier to implement.

simison:
@nicksellen nice, thanks for studying this! I was looking into Boxcar.io/Onesignal.com docs and it indeed doesn't seem so simple as they advertise.

I think I'll try to set up a demo with FCM. Any help building this feature would be greatly appreciated! ;-)

[...]

PR #491, opened 2017-04-10
Add firebase cloud messaging and browser push

nicksellen:
Initial work only.

Currently only pushes messages when you register for push.

Need to have `firebase.json` in project root with service account credentials from FCM. Will move to config somewhere in future.

No tests yet...


nicksellen:
I changed the plan a bit from #394 -

There are two concepts:
1. does the user want push notifications enabled (intention)
2. are push notifications enabled (reality)

I realised just storing the tokens in the database does not reflect intention properly. To support removing all registrations in one go it would need to keep the registration but mark it for removal, so when that browser connects again it can uninstall itself. This seems complicated and not especially useful.

Part of the complexity is that you cannot assume the registration tokens will stay the same forever, and the browser can always get one. If intention were stored in the database the browser would need an identifier so it can know which registration is intended for it (or maybe add something to the session). But again, getting complicated and little benefit right now.

I changed it so you so that now you and enable/disable push for the _current browser_ you are using, and nothing else. The _intention_ is stored in the browser (in a cookie), and the client app does it's best to make it so. I think this fits well to how mobile works too.

So UX is nice and simple and looks like:

pushconfig

Thoughts?

And some tasks from the issue:

- [ ] get feedback on UX (not quite convinced yet, maybe a remove from this browser too?)
- [ ] I would like to only load the firebase js libs if they are needed (only ~11kb though... in dev even)
- [x] I would like to fully remove the service worker when you turn off notifications, fcm doesn't seem designed to make this easy though, seems possible though, but maybe a bit more manual work
- [x] some code tidy/rejig
- [ ] notification actions when you click the message
- [ ] plug it into the job thing to send notifications when needed
- [x] check browser supports enough things first
- [ ] ... tests

simison:
> plug it into the job thing to send notifications when needed

This would be at [`message-unread.server.job.js`](https://github.com/Trustroots/trustroots/blob/master/modules/messages/server/jobs/message-unread.server.job.js#L144-L153) for unread notifications. I don't think we want to complicate this with any other kind of notifications just yet.


>So UX is nice and simple and looks like
>...
>get feedback on UX

I like it! We'll probably want to nag users who don't have it checked by asking them if they want to enable it, store that we've nagged them to the browser session and ask again only when they login with new browser. But that can be the next step once this PR is done.


> load the fcm server config from somewhere better

You can just add an example config block to [`config/env/default.js`](https://github.com/Trustroots/trustroots/blob/master/config/env/default.js) and the real one to your `config/env/local.js` (and perhaps to [`config/env/local.sample.js`](https://github.com/Trustroots/trustroots/blob/master/config/env/local.sample.js))


> I would like to only load the firebase js libs if they are needed (only ~11kb though... in dev even)

Another feature for another PR. :-)

> notification actions when you click the message

Link to the message thread. Just add analytics tags to the URL, [see an example](https://github.com/Trustroots/trustroots/blob/master/modules/core/server/services/email.server.service.js#L39-L44).

nicksellen:
> We will probably want to nag users

Yup, getting it working and live first seems more useful, make sure it actually works... I think it's pretty easy to get too naggy and annoying with push.

I suspect it will be better to notify on every new message too. I believe you can tag the messages too such that newer ones replace older ones so they don't stack up.

nicksellen:
Aha, my fight with travis was victorious! 🥇

simison:
>I suspect it will be better to notify on every new message too. I believe you can tag the messages too such that newer ones replace older ones so they don't stack up.

Yep.

What's up with the change to Firefox?

nicksellen:
PhantomJS doesn't have much of the newer browser stuff supported and I'd like to add tests for the browser notifications.

Although using PhantomJS locally (same version too, 2.1.1 iirc) doesn't fail in the tests (isSupported gets set to true) but in travis it failed inside the firebase js lib because `Promise` doesn't exist.

Would be good to check it more properly with an unsupported browser to make sure it works fine.

simison:
>Would be good to check it more properly with an unsupported browser to make sure it works fine.

We have a BrowserStack account you could use to test old browsers. @chmac could share it with you via LastPass. (@chmac could you also re-share it with me and tick "can modify the password"?)

chmac:
@simison Don't you have your own account? Anyway, I've just shared the login with you.

@nicksellen If you send me your lastpass email I can share it with you.

simison:
> Don't you have your own account? Anyway, I've just shared the login with you.

Sure, but for some reason I couldn’t see the password nor share that one.

Thanks!

nicksellen:
@chmac lastpass@nicksellen.co.uk

nicksellen:
There are nice tests now :) It was quite a battle to get karma/etc to be happy, so easy to lose track of a promise somewhere...

I split the client service into firebaseMessaging and push, so I can test the push service without loading any firebase libs. I'm quite pleased with the simplicity of the promise-based stuff [1].

[1] https://github.com/Trustroots/trustroots/blob/a5800a8d5c7b634018d5f2bd99058491d4d15c44/modules/core/client/services/push.client.service.js#L117

nicksellen:
F*** you Travis!

nicksellen:
Ready for review!

Just a few things to look at https://github.com/Trustroots/trustroots/pull/491#issuecomment-293013012

modules/core/tests/client/push.client.service.tests.js
```diff
@@ -0,0 +1,262 @@
...
+          }
+        });
+
+      // if we turn it on...
+      $cookies.put('tr.push', 'on');
+      firebase.permissionGranted = true;
+      firebase.token = token;
+
+      // .. and enable it to be initialized
+      firebaseMessaging.shouldInitialize = true;
```
nicksellen: set this to false in a `beforeEach` hook

modules/core/client/config/core.client.run.js
```diff
@@ -6,12 +6,14 @@
...
 
   /* @ngInject */
-  function coreRun(Facebook) {
+  function coreRun(Facebook, push) {
 
     // Attempt to initialize Facebook SDK on first page load
     // If this fails, we'll try this again on successfull login
     Facebook.init();
 
+    push.init();
```
simison: A short comment over this could do. :-)

modules/core/server/routes/core.server.routes.js
```diff
@@ -22,6 +22,9 @@ module.exports = function(app) {
   // Return a 404 for all undefined api, module or lib routes
   app.route('/:url(api|modules|lib|developers)/*').get(core.renderNotFound);
 
+  // Gives the service worker access to any config it needs
+  app.route('/config/sw.js').get(core.renderServiceWorkerConfig);
```
simison: Normally these been exposed to index html template from [config/lib/express.js](https://github.com/Trustroots/trustroots/blob/master/config/lib/express.js#L31).

modules/core/server/controllers/core.server.controller.js
```diff
@@ -57,3 +58,11 @@ exports.receiveCSPViolationReport = function(req, res) {
...
   res.status(204).json();
 };
+
+/**
+* Render javascript content containing service worker config.
+*/
+exports.renderServiceWorkerConfig = function(req, res) {
+  res.set('Content-Type', 'text/javascript')
+     .send('var FCM_SENDER_ID = ' + JSON.stringify(config.fcm.senderId) + ';\n');
+};
```
simison: Express has support for [jsonp](https://expressjs.com/en/api.html#res.jsonp), but normally configs have been exposed to index html template from [config/lib/express.js](https://github.com/Trustroots/trustroots/blob/master/config/lib/express.js#L31).

public/push-messaging-sw.js
```diff
@@ -0,0 +1,30 @@
...
+
+firebase.initializeApp({
+  'messagingSenderId': FCM_SENDER_ID
+});
+
+var messaging = firebase.messaging();
+
+messaging.setBackgroundMessageHandler(function(payload) {
+  // not actually used, but without it here firefox does not receive messages...
+  console.log('received payload', payload);
```
simison: This function _never_ gets fired but it has to be there? `messaging.setBackgroundMessageHandler(Function.prototype)` might be handy.

Also if Angular is already in `window` at this point, [angular.noop](https://docs.angularjs.org/api/ng/function/angular.noop) is available.

public/push-messaging-sw.js
```diff
@@ -0,0 +1,30 @@
+/* global importScripts firebase FCM_SENDER_ID */
+// Give the service worker access to Firebase Messaging.
+// Note that you can only use Firebase Messaging here, other Firebase libraries
+// are not available in the service worker.
+
+importScripts('/lib/firebase/firebase-app.js');
+importScripts('/lib/firebase/firebase-messaging.js');
+importScripts('/config/sw.js');
```
simison: Okay, now I understood why the config was loaded like it was earlier! Ignore my earlier comments about this.

We should probably move all the configs to load like this. :-)

package.json
```diff
@@ -140,6 +141,7 @@
     "nodemailer-stub-transport": "~1.1.0",
     "phantomjs-prebuilt": ">=2.1",
     "promise": "~7.1.1",
+    "proxyquire": "^1.7.11",
```
simison: `~` just for consistency.

package.json
```diff
@@ -61,6 +61,7 @@
     "express-paginate": "~0.2.1",
     "express-session": "~1.14.0",
     "fbgraph": "~1.3.0",
+    "firebase-admin": "^4.2.0",
```
simison: `~` just for consistency.

public/push-messaging-sw.js
nicksellen: No angular here, the service worker is a standalone script and they run with a limited environment.

https://github.com/firebase/quickstart-js/issues/71 has a discussion about it - the gist is that by sending a notification payload (as opposed to data) you are specifying it should be a regular notification, and not be overridable.

my suspicion about why it didn't work without that in firefox is that without defining _some_ kind of listener it ends. it would be useful to verify it properly though...

modules/core/server/routes/core.server.routes.js
nicksellen: The service worker does not have access to the HTML/DOM. It's very restricted.

Perhaps a better option is sending a message to it once it's started up [1]. I'm not very happy with the current approach.

[1] http://craig-russell.co.uk/2016/01/29/service-worker-messaging.html

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -875,3 +876,130 @@ exports.getUserMemberships = function(req, res) {
...
+
+    // Notify the user we just added
+
+    function(done) {
+      if (process.env.NODE_ENV === 'production') return done(); // only in dev...
+
+      pushService.sendUserNotification(user, {
+        title: 'Trustroots',
+        body: 'A ' + platform + ' push device was added!',
+        click_action: 'http://localhost:3000/profile/edit/account'
```
simison: This might work here better for flexibility with different setups (Docker IP, https):

```js
var protocol = (config.https === true || req.protocol === 'https') ? 'https' : 'http';
var host = protocol + '://' + req.get('host'); // http://localhost:3000
```

...or actually this might be already available as `res.locals.hostPort`

public/push-messaging-sw.js
nicksellen: Ah, I liked them in `index.html`. This way would be 1 extra http request too, but could load the config into a standard `TrustrootsConfig` object or something. I'm easy.

modules/users/server/controllers/users/users.profile.server.controller.js
nicksellen: Oh yeah, I forgot to fix that bit.

I also wonder whether it would be a good idea to send a notification at that point in production too, but only to the device you just added?

public/push-messaging-sw.js
simison: They're handy in `index.html` indeed but stops making the app work self hosted and offline (e.g. on mobile). It's another discussion and another PR anyway.

modules/users/server/controllers/users/users.profile.server.controller.js
nicksellen: Also, maybe in development some other way to send notifications could be handy for manual testing/playing purposes...

public/push-messaging-sw.js
simison: Okidoki, thanks for clarifying!

modules/users/server/controllers/users/users.profile.server.controller.js
simison: >I also wonder whether it would be a good idea to send a notification at that point in production too, but only to the device you just added?

Yeah I think so, it kinda confirms to the user that it worked.

So to send push to all the devices, just remove that [`if`](https://github.com/Trustroots/trustroots/pull/491/files#diff-028054ee3eec2278e53157ccf4668fe4R973)? I think it's okay to send it to all the devices even (how many there can be), just to avoid extra work. Feels kind of a security feature even in that way anyway. :-)

simison: Copy could be something like _"You just enabled Trustroots `platform` push notifications. Yay!"_

modules/users/client/views/profile/profile-edit-account.client.view.html
```diff
@@ -137,6 +137,44 @@
...
+                       name="browser"
+                       ng-model="profileEditAccount.pushEnabled"
+                       ng-disabled="profileEditAccount.pushIsDisabled()"
+                       ng-change="profileEditAccount.pushUpdate()">
+                Current web browser
+              </label>
+            </div>
+
+            <p class="help-block" ng-if="profileEditAccount.push.isSupported"><small>This will send notifications even when the website is closed.</small></p>
+            <p class="help-block" ng-if="!profileEditAccount.push.isSupported"><small>Your browser does not support push notifications.</small></p>
```
simison: When unsupported, the whole block could just be hidden?

nicksellen: I did it that way so it communicates: _if you use another browser then you can get push notifications_

It also helps to keep a consistent interface, so users don't get confused wondering where the option went when they had it available on another browser.

I don't feel _that_ strongly about it though. Happy to hide it all if unsupported if you prefer.

simison: Good thinking :thumbsup:, let's leave it visible.

simison:
>use browserstack... automated or manual?

There's an old script for automated testing: https://github.com/Trustroots/trustroots/tree/master/scripts/selenium

...but frankly I'm not too keen to keep that up to date. I think manual testing and those few unit tests are great, we should notice pretty quick if it fails, esp. if we push some events to Influx stats.

bower.json
```diff
@@ -4,6 +4,7 @@
   "dependencies": {
     "angular": "1.5.8",
     "angular-animate": "1.5.8",
+    "angular-cookies": "1.5.8",
```
simison: BTW there's already https://github.com/tymondesigns/angular-locker which might save you some kb's since no need to add this lib, but I'm fine either way.

modules/users/client/views/profile/profile-edit-account.client.view.html
simison: The section could go above password tho, so it would be more noticeable.

image

bower.json
nicksellen: Seems a good idea

modules/core/server/services/push.server.service.js
```diff
@@ -7,6 +7,26 @@ var _ = require('lodash'),
...
+  var editAccountUrl = url + '/profile/edit/account';
+
+  var notification = {
+    title: 'Trustroots',
+    body: 'You just enabled Trustroots ' + platform + ' push notifications. Yay!',
+    click_action: analyticsHandler.appendUTMParams(editAccountUrl, {
+      source: 'push-notification',
+      medium: 'fcm',
+      campaign: 'device-added',
+      content: 'reply-to'
```
nicksellen: Not sure what this should be?

nicksellen:
Hmm a weird error from a simple change:

```
dst is undefined in public/lib/angular/angular.js
```

Investigating...

nicksellen:
Aha, `SettingsService` is better than `SettingsFactory`...

nicksellen:
@simison not sure why the _pr_ build fails so frequently when the _push_ one seems fine? I don't think there is an actual error.

One thing I haven't done is tried it in other browsers, but this is not totally simple because the security stuff requires https when not using localhost, so I can't access it from my phone via my machines ip address.

One solution would be to setup https locally, but also need to change a few settings to use `DOMAIN` env value instead of localhost (well, maybe just the csp dev config). I don't have time for this right now, and would suggest just trying it out when properly deployed.

nicksellen:
Perhaps someone at least has a windows machine or mac to try it on though?

simison:
I'll build this to dev machine, just a mo.

modules/core/server/services/push.server.service.js
simison: `medium` could probably be just `push-notification`, I'm not sure if we gonna differentiate different platforms in the future but doesn't matter for now.

Since there is just one place (notification click) pointing to the same url, `content` can probably be left empty. It's used to make difference e.g. between different parts of email linking to the same URL (e.g. "footer" and "header"). ...I think.

They're explained in [this doc](https://support.google.com/analytics/answer/1033863#parameters) and at [query builder](https://ga-dev-tools.appspot.com/campaign-url-builder/)

simison:
>not sure why the pr build fails so frequently when the push one seems fine? I don't think there is an actual error.

Switching to Firefox is probably "causing" it to fail now, but I'm not sure why exactly. Might be just some `done()`:s missing since it's just timeouting.

nicksellen:
Tried it on dev.trustroots.org on desktop and mobile, it all works!

## Chrome

  

Chrome did not show the notification when the browser tab was the active one on the screen, but did when the tab was closed, and when the browser was not in the foreground. Not actually sure about if the browser is completely closed (I guess not).

When loading the site again after enable notifications it  asked me if I wanted to add it to the home screen (so I suspect it only gets used once a web worker has been installed).

The notification icon is the chrome icon.

## Firefox

 

Firefox always showed the notification (well, didn't try with browser completely closed, I guess not).

The notification icon is the firefox icon.

## Android browser (not supported)

simison:
Great!

So my todo seems to be:
- [ ] get FCM account and config for Trustroots
- [ ] look into those failing Karma test if I could help it out somehow
- [ ] merge + deploy live

nicksellen:
> Switching to Firefox is probably "causing" frontend Karma tests to fail now, but I'm not sure why exactly. Might be just some done():s missing since it's just timeouting.

I do not feel motivated to try and fix it for this issue. I think it could take some time and is not really related to the push features. Actually it might be possible to switch to back to PhantomJS as I have abstracted the firebase stuff further away now - just pushed a commit to revert it back (although I think it will have to change again eventually now the maintained stepped down).

simison:
>>Switching to Firefox is probably "causing" frontend Karma tests to fail now, but I'm not sure why exactly. Might be just some done():s missing since it's just timeouting.

>I do not feel motivated to try and fix it for this issue. I think it could take some time and is not really related to the push features. Actually it might be possible to switch to back to PhantomJS as I have abstracted the firebase stuff further away now - just pushed a commit to revert it back (although I think it will have to change again eventually now the maintained stepped down).

I just checked, our Karma is 1.3.0 and newest is 1.6.0 so that might help as well. Updating it + switching to Firefox seems like a separate, not so urgent PR.

PR #493, opened 2017-04-13
Add manifest.json and base service worker

simison:
Adds [W3C manifest ](https://developers.google.com/web/updates/2014/11/Support-for-installable-web-apps-with-webapp-manifest-in-chrome-38-for-Android) so that:
- Chrome will [suggest to add](https://developers.google.com/web/fundamentals/engage-and-retain/app-install-banners/) the site to home screen
- We can define [service workers](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer) for push notifications (issue #394, PR #491) by [adding a key to this manifest file](https://firebase.google.com/docs/cloud-messaging/js/client#configure_the_browser_to_receive_messages).

Read more:
https://developer.mozilla.org/en-US/docs/Web/Manifest


simison:
@nicksellen so this PR should go in before #491 — the you can then rebase and add `"gcm_sender_id": "<Sender ID Here>"` to `manifest.json` as mentioned at [Firebase docs](https://firebase.google.com/docs/cloud-messaging/js/client#configure_the_browser_to_receive_messages).

PR #495, opened 2017-04-16
Fixes #348 Typing reply message scrolls down the existing messages

### LINKED ISSUE #348: Typing reply message scrolls down the existing messages
I want to reply to a message thread which is long enough that it needs to scroll. I want to see a specific part while writing, so I scrolled up a bit. However when I type, after some seconds the message view is scrolled down. This is quite annoying as you can imagine. :)

vssdeo:
- newly added event threadMessageSend is broadcasted to handle scroll
- scroll enable / disable functions added
- behaviour is: if user seeing scrolled history, scroll is locked
- if user sends the message, scroll is unlocked and scroll happens
- used mouseover events
- tested on fresh install

### DIFF
--- a/modules/messages/client/controllers/thread.client.controller.js
+++ b/modules/messages/client/controllers/thread.client.controller.js
@@ -291,7 +291,12 @@
 
         // $timeout ensures scroll happens only after DOM has finished rendering
         $timeout(function() {
-          $scope.$broadcast('threadScrollToBottom');
+          // changing event as scroll should happen not at
+          // every char type as person would be looking at
+          // scroll history, but once he sends the message
+
+          // $scope.$broadcast('threadScrollToBottom');
+          $scope.$broadcast('threadMessageSend');
         });
 
       }, function(errorResponse) {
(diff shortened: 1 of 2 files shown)


simison:
Thanks! Did you test this on mobile as well?

vssdeo:
You were right, I didn't think of touchscreen events, which was locking up the view in one use case, though rest of functionality was ok, but this could only have been caught on physical device. I have added the event touchstart handling and tested on ios iphone 7 device, unfortunately I don't have more (android devices) to cross check with.

simison:
Tested this out.

Otherwise nice but there's a lot of visual jumping happening now because the scroll bar is disappearing depending on if I'm hovering the messages list or text input:

image

Additionally I can't type anymore if I'm not keeping the mouse hovered over the textarea, which is even a bigger bummer.

I think it's less annoying to have automatic scroll happening instead of those two. ;-)

simison:
I'll close this as stale; it's a nice attempt but has some bugs so it could not be merged at the time of implementing.

PR #499, opened 2017-04-22
Tweaking the manifesto

chmac:
Small change requested by @Tartesos.

@simison Untested as I don't have the code running...


simison:
Since someone asked; our willingness to help each other shouldn’t still be monetised of course. ;-) 

...but like this the text is a bit less direct attack to CS. We’re not “a CS clone” and we are not based on being just “anti CS”. Something like that shouldn’t be the only reason to join TR.

We stand on our own and do our own thing. :sparkles: :beers:

PR #500, opened 2017-04-22
Fixes #422 Quickreplies / hosting requests

### LINKED ISSUE #422: Quickreplies
- Allow just pressing "accept" or "decline" as a reply
- Show these buttons directly at the email (important)
- Show these buttons on top of (or inside?) the message field
- Show the answer right away to the sender (basically just send predefined text as a normal message)
- ~~Add a simple/rough checkbox when sending a message to new thread "[x] this is a hosting request"~~ — this is not be required and should be its own issue anyway. We can just have these buttons always there and assume first message in the thread is a hosting request _most of the time_.

Related in future: _Attached metadata for messages_ #233. At first "quick reply" can be just pre-formatted text though.

vssdeo:
- Adds very basic functionality
- No state of request stored
- Just enables only quick response in a one-liner
- More to be added, after discussion

### DIFF
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -49,13 +49,6 @@ module.exports.initLocalVariables = function (app) {
   app.locals.appSettings.fcmSenderId = config.fcm.senderId;
   app.locals.siteAnnouncement = config.siteAnnouncement || { enabled: false };
 
-  // Predefined to be localized
-  app.locals.appSettings.messages = {
-    msgCanHost: 'Yes, I can host!',
-    msgCannotHost: 'Sorry, I can\'t host',
-    msgWriteBack: 'Write back'
-  };
-
   if (process.env.NODE_ENV !== 'production') {
     app.locals.jsFiles = _.concat(config.files.client.js, 'dist/uib-templates.js');
     app.locals.cssFiles = _.map(config.files.client.css, function(file) { return file.replace('/client', ''); });
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -49,6 +49,13 @@ module.exports.initLocalVariables = function (app) {
   app.locals.appSettings.fcmSenderId = config.fcm.senderId;
   app.locals.siteAnnouncement = config.siteAnnouncement || { enabled: false };
 
+  // Predefined to be localized
+  app.locals.appSettings.messages = {
+    msgCanHost: 'Yes, I can host!',
+    msgCannotHost: 'Sorry, I can\'t host',
+    msgWriteBack: 'Write back'
+  };
+
   if (process.env.NODE_ENV !== 'production') {
     app.locals.jsFiles = _.concat(config.files.client.js, 'dist/uib-templates.js');
     app.locals.cssFiles = _.map(config.files.client.css, function(file) { return file.replace('/client', ''); });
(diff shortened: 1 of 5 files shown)


modules/messages/client/views/thread.client.view.html
```diff
@@ -105,7 +105,22 @@
...
+      <!-- quickreply -->
+      <div class="btn-toolbar" id="message-quick-reply" ng-hide="thread.hideQuickReplyPanel">
+          <button class="btn btn-success icon-ok" ng-click="thread.sendMessage('<b><i>'+app.appSettings.messages.msgCanHost+'</i></b>');">
+            {{app.appSettings.messages.msgCanHost}}
+          </button>
+          <button class="btn btn-danger icon-minus-squared-alt" ng-click="thread.sendMessage('<b><i>'+app.appSettings.messages.msgCannotHost+'</i></b>');">
+            {{app.appSettings.messages.msgCannotHost}}
+          </button>
+          <button class="btn btn-info icon-reply" ng-click="thread.hideQuickReplyPanel=true;">
+            {{app.appSettings.messages.msgWriteBack}}
```
simison: Use angular [one time binding](https://docs.angularjs.org/guide/expression#one-time-binding) (`::`) here:
```js
{{::app.appSettings.messages.msgWriteBack}}
```
...since those labels don't change and no need to watch for changes.

modules/messages/client/views/thread.client.view.html
```diff
@@ -105,7 +105,22 @@
       </threads>
       <!-- /Thread -->
 
+      <!-- quickreply -->
+      <div class="btn-toolbar" id="message-quick-reply" ng-hide="thread.hideQuickReplyPanel">
+          <button class="btn btn-success icon-ok" ng-click="thread.sendMessage('<b><i>'+app.appSettings.messages.msgCanHost+'</i></b>');">
+            {{app.appSettings.messages.msgCanHost}}
+          </button>
+          <button class="btn btn-danger icon-minus-squared-alt" ng-click="thread.sendMessage('<b><i>'+app.appSettings.messages.msgCannotHost+'</i></b>');">
```
simison: Add `<p></p>` around the html here, so:

`'<p><b><i>' + app ... </i></b></p>`

modules/messages/client/views/thread.client.view.html
```diff
@@ -105,7 +105,22 @@
       </threads>
       <!-- /Thread -->
 
+      <!-- quickreply -->
+      <div class="btn-toolbar" id="message-quick-reply" ng-hide="thread.hideQuickReplyPanel">
+          <button class="btn btn-success icon-ok" ng-click="thread.sendMessage('<b><i>'+app.appSettings.messages.msgCanHost+'</i></b>');">
```
simison: Let's remove icons for now to make sure it fits better with mobile.

config/lib/express.js
```diff
@@ -49,6 +49,13 @@ module.exports.initLocalVariables = function (app) {
   app.locals.appSettings.fcmSenderId = config.fcm.senderId;
   app.locals.siteAnnouncement = config.siteAnnouncement || { enabled: false };
 
+  // Predefined to be localized
+  app.locals.appSettings.messages = {
+    msgCanHost: 'Yes, I can host!',
+    msgCannotHost: 'Sorry, I can\'t host',
+    msgWriteBack: 'Write back'
+  };
```
simison: Let's move these to Angular controller instead. We'll have a proper i18n later on, no need to prepare for it just yet because it might look very different.

Or actually just having them written static would be okay at this point to avoid complicating it.

vssdeo:
all changes done except for "To be removed before merge."

PR #507, opened 2017-04-29
506 request hosting info panel

vssdeo:
To complement #422 Quickreplies / hosting requests, when the first person is sending the first message, he can:

*while requesting to be hosted, click a checkbox & give details of
- number of people
- date from
- date to


modules/messages/client/controllers/thread.client.controller.js
```diff
@@ -32,6 +32,12 @@
         messageIdsInView = [],
         editorContentChangedTimeout;
 
+    // Vars for Host Module
+    var fromDateDisp,
```
mrkvon: It would be nice to have short comments about purpose of the variables here.

vssdeo: Ah ! Sure I was putting comments for functions need to put more here, I think I will put more in detail in thread dimension directive & other little places too !

vssdeo:
@mrkvon added the comments explaining variables

simison:
We might get back to implementing this at some point but for now, it's not on our agenda.

PR #508, opened 2017-05-03
Fixes #496 copy & paste in text area of send messages removes formatt…

### LINKED ISSUE #496: copy & paste in text area of send messages removes formatting especially newlines
Sending messages on trustroots I am copying and pasting the travel template of my plans, and it loses it formatting especially the newlines.

Reproduce.

1. Copy a block of text with formatting or just with new lines
2. Go to messages & paste while sending a message
3. As soon as you paste all formatting gone

vssdeo:
…ing especially newlines

- Upgraded since I was working on the code base: medium-editor ~5.23.0
~ created a set of editor options: tr-editor-options="::app.editorOptionsThreadReply, but only for thread part so can be given options independent of other text input parts

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -26,7 +26,7 @@
     "ng-file-upload": "~12.0.4",
     "PruneCluster": "2.0.0-beta.2",
     "angular-locker": "~2.0.4",
-    "medium-editor": "~5.22.1",
+    "medium-editor": "~5.23.0",
     "angular-confirm-modal": "~1.2.5",
     "angulartics": "~1.4.0",
     "angulartics-google-analytics": "~0.4.0",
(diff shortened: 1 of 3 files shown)


simison:
Thanks. I prefer to have limitations on what kind of html people end up copy-pasting. Most of the time it's something that messes up with the layout and that message thread CSS/html is already complicated enough, adding any extra html there might just risk breaking it.

So [this](https://github.com/Trustroots/trustroots/commit/51d6c67250f5a6baa08016104b785ae3fc7e6f2f) cleans up pretty much everything else but allowed tags, so you can now copypaste "message templates" into messages. :-)

PR #510, opened 2017-05-06
490 precomposed template contact info location

### LINKED ISSUE #490: Precomposed template contact info/location message to share with guests
Idea! a private "welcome to my house" section, maybe just a plaintext textfield, that users can use to compose a message with their contact info, address, house rules, etc. Once they decide to host someone, they can use a "send welcome message" button within message threads to give guests their information.

And maybe a similar "traveler contact info" section, a similar plaintext field where travelers can compose a message with their contact info to share back to hosts.

--

Background information: 

I notice a pattern when I'm communicating on hospitality sites: after a host decides to welcome a traveler into their home, they share some contact information, generally along the lines of:

- synchronous contact info (whatsapp, messenger, sms)
- address (plaintext, or google maps link)
- additional instructions ("go through the side gate", "I get off work at 6pm, so please arrive after that")
- maybe public transit info
- other house stuff, like wifi passwords, please take anything you'd like from the fridge, take off shoes before entry, book exchange shelf in the livingroom, no meat zone, etc

I end up re-assembling a message like this from scratch each time, or copy-pasting from a past message. Some of the hosts I've had have something clever ready, such as a "welcome guide" page on their personal website, so they only need to send a link that has all the information. My idea here is to improve ease of communication by simplifying the action of rewriting (with the possibility of leaving important stuff out), copy/pasting this message each time, or creating a separate external webpage (I like doing that, but not everyone does :P )

Just some ideas! I'm kind of excited about this so I may give it a shot after I'm back from the mountains/highlands in May (: what do yall think?

vssdeo:
Fixes #490 Precomposed template contact info/location message to share with guests
    -Adds a little check box under Yes,I can host, button in quick-reply panel
    -User template is stored in offers, under Hosting Status
    -Adds a function to send template
    -Requires #422 (committed as part of this too)
    -The quick reply panel does not self hide, may be the behaviour can be discussed and finalised

### DIFF
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -49,6 +49,13 @@ module.exports.initLocalVariables = function (app) {
   app.locals.appSettings.fcmSenderId = config.fcm.senderId;
   app.locals.siteAnnouncement = config.siteAnnouncement || { enabled: false };
 
+  // Predefined to be localized
+  app.locals.appSettings.messages = {
+    msgCanHost: 'Yes, I can host!',
+    msgCannotHost: 'Sorry, I can\'t host',
+    msgWriteBack: 'Write back'
+  };
+
   if (process.env.NODE_ENV !== 'production') {
     app.locals.jsFiles = _.concat(config.files.client.js, 'dist/uib-templates.js');
     app.locals.cssFiles = _.map(config.files.client.css, function(file) { return file.replace('/client', ''); });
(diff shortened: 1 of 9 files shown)


simison:
Closing as stale; templates is a "nice to have" feature so might be we'll get to this again one day!

PR #516, opened 2017-05-20
Fixes #513 better-sharing-invite-code

### LINKED ISSUE #513: Better sharing of invite code
It would be awesome to enhance the invitation page to make it easier to share the codes. A couple of things would make this easier.

* Wrap the code in a sentence so that folks can easily share the code plus the fact it expires, a link to trustroots.org, and so on
  - First draft of that sentence, feel free to enhance / improve / etc, @simison is the expert on this
  - Would you like to join me on Trustroots? It's an invitation only travellers community at www.trustroots.org. Your invite code is XXXXX which expires in 24 hours or less.
  - For email, the subject could be "Want to me on trustroots.org?"
* Add share buttons to the page so that this sentence can be easily shared to these services (in approximate order of priority)
  - Facebook Messenger (ideally messenger only instead of Facebook generally, but if that's not possible, Facebook generally is good)
  - WhatsApp (there are whatsapp:// URLs on mobile / desktop)
  - Gmail (it's possible to link directly to a prefilled gmail compose window
  - Mailto (can prefill subject and body)
  - Whatever other social networks the kids are using these days
  - Twitter (possible to share to a direct message on twitter?)

## Implementation

It's probably easiest to implement this via addthis.com which has a free plan, and will handle most of these services for us. See @simison or @chmac for access to our account.

vssdeo:
- uses addthis.com
- email templates are set on addthis.com
- twitter, title, description is set in js block configuration
- for now js is embedded in html but can be moved once finalized
- testing for facebook share and facebook-messenger needs to be checked as in former case code is added as part of share but for messenger, I am unable to find an option


vssdeo:
Edit the templates for language or something https://www.addthis.com/dashboard#profile-options/ra-591c2040c6459c4a/email-templates-settings

simison:
@vssdeo can you rebase with latest master? You can add this repo as a new remote (`git remote add someremotename https://github.com/Trustroots/trustroots.git`) and fetch it, then `git rebase -i someremotename/master`

someremotename  could be e.g. "upstream".

If you're not familiar with rebasing, I always recommend to take backup of your folder first. ;-)

modules/users/client/controllers/signup.client.controller.js
```diff
@@ -21,7 +21,7 @@
     vm.suggestionsLimit = 3; // How many tribes suggested (including possible referred tribe)
 
     // Variables for invitation feature
-    vm.invitationCode = '';
+    vm.invitationCode = $location.search().code;
```
simison: You can use `$stateParams.code` here, and should probably be `$stateParams.code || ''` just to keep it as a string.

modules/users/client/views/invite.client.view.html
```diff
@@ -26,6 +26,33 @@
...
+              passthrough: {
+                twitter: {
+                  text: 'You have been invited to join Trustroots. Your invite code: ' + code + ' and url:'
+                }
+              }
+            };
+          </script>
+          </div>
+          <div class="addthis_inline_share_toolbox"></div>
+          <script type="text/javascript" src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-591c2040c6459c4a" async="async"></script>
```
simison: Make sure to add `*.addthis.com` to CSP rules: https://github.com/Trustroots/trustroots/blob/13ff4ddf171c09f9224162f9714b85e39e58995f/config/lib/express.js#L270-L271

modules/users/client/controllers/signup.client.controller.js
simison: Also add `&code` to here:

https://github.com/Trustroots/trustroots/blob/master/modules/users/client/config/users.client.routes.js#L261

simison: Maybe also this?
```js
// If invitation code was passed via URL, validate it right away
if (vm.invitationCode) {
  validateInvitationCode();
}
```

modules/users/client/controllers/signup.client.controller.js
```diff
@@ -82,6 +75,14 @@
      */
     function activate() {
 
+      if (! Authentication.user && vm.invitationCode) {
+        validateInvitationCode(
+          appSettings.invitations.key, // inviteKey
+          new Date(), // today
+          vm.invitationCode.toLowerCase() // code
+        );
+      }
```
simison: I changed how this works recently; make sure to fetch the latest master get latest changes.

You can simply do this here and it'll handle the rest:
```js
if (vm.invitationCode) {
  validateInvitationCode();
}
```

Also no need to check for `Authentication.user` as it's done here already:

https://github.com/Trustroots/trustroots/blob/master/modules/users/client/controllers/authentication.client.controller.js#L11-L12

modules/users/client/views/invite.client.view.html
```diff
@@ -15,10 +15,10 @@
 
         <p class="lead">Your invitation code</p>
 
-        <input type="text"
+        <input id="invite" type="text"
                class="form-control text-center input-lg font-brand-semibold text-uppercase"
                tr-select-on-click
-               ng-attr-value="{{ ::invite.invitation.code }}">
+               ng-attr-value="{{ ::invite.invitation.code }}" readonly>
```
simison: `readonly` is a good idea, but just move it to newline to align with others.

modules/users/client/views/invite.client.view.html
```diff
@@ -15,10 +15,10 @@
 
         <p class="lead">Your invitation code</p>
 
-        <input type="text"
+        <input id="invite" type="text"
```
simison: Move `type` to newline to align with others.

id could be something less generic, eg. `tr-invite-code`?

modules/users/client/views/invite.client.view.html
```diff
@@ -27,6 +27,33 @@
...
         </p>
 
+        <!-- Go to www.addthis.com/dashboard to customize your tools -->
+        <p>
+          <div ng-if="::invite.invitation.code">
+          <script type="text/javascript">
+            var code = document.getElementById("invite").value.toUpperCase();
+            var addthis_config = addthis_config||{};
+                addthis_config.ui_email_from = 'help@trustroots.org';
+                addthis_config.ui_email_note = 'Your invite code: ' + code + ' would be shared which can be used for signup with Trustroots at https://www.trustroots.org/signup?code=' + code;
```
simison: There's probably a way with `$location` (https://docs.angularjs.org/api/ng/service/$location) to make that domain dynamic, so that we can test this at dev.trustroots.org as well?

config/lib/express.js
```diff
@@ -279,6 +279,7 @@ module.exports.initHelmetHeaders = function (app) {
         '*.twitter.com',
         '*.google-analytics.com',
         '*.gstatic.com', // Google analytics related
+        '*.addthis.com', // addthis.com social media sharing service
```
simison: Looks like they use also `*.addthisedge.com`. You can see CSP warnings in console log by keeping it open when opening the page.

modules/users/client/views/invite.client.view.html
```diff
@@ -27,6 +27,33 @@
...
+            var addthis_share = {
+              url: 'https://www.trustroots.org/signup?code=' + code,
+              title: 'Invite by ' + user.displayName + ' to join Trustroots',
+              description: 'You have been invited to join Trustroots, an invite only community for travelers. Your invite code to join Trustroots is: ' + code,
+              email_template: 'new_template',
+              email_vars: { 'code': code, 'invitee': user.displayName },
+              passthrough: {
+                twitter: {
+                  text: 'You have been invited to join Trustroots. Your invite code: ' + code + ' and url:'
+                }
```
simison: First two buttons (FB + Twitter) can be removed as they're public shares and we want to have personal sharing.

simison: I just found out where to modify these and went on and made it look like this:

image

simison:
The page could look like this:

image

config/lib/express.js
simison: `*.addthis.com` probably has to be added to `childSrc` rules as well.

Check also other CSP warnings from console.log (at least Chrome shows them up).

PR #528, opened 2017-07-16
Exponent push notifications

bitfasching:
These commits refactor the existing Firebase push implementation to allow for additional cloud providers to be added and implement the Exponent push notification gateway. (#394)

### DIFF
--- a/config/lib/exponent-notifications.js
+++ b/config/lib/exponent-notifications.js
@@ -1,0 +1,33 @@
+'use strict';
+
+var Expo = require('exponent-server-sdk');
+
+// to check if something is a push token
+exports.isPushToken = Expo.isExponentPushToken;
+
+// create a new Expo SDK client
+var expo = new Expo();
+
+// send push notification (returns Promise)
+exports.sendToDevice = function sendToDevice(tokens, notification) {
+
+  // Promises for each token to push to
+  var pushPromises = [];
+
+  // iterate over tokens
+  tokens.forEach(function(token) {
+    // dispatch push notification and save returned Promise
+    pushPromises.push(expo.sendPushNotificationsAsync([{
+      // Exponent target device token
+      to: token,
+      // text to display
+      body: notification.body || '',
+      // action URL
+      data: { url: notification.click_action }
+    }]));
+  });
+
+  // return unified Promise
+  // (resolves when all push actions succeeded, rejects if any of them fails)
+  return Promise.all(pushPromises);
+};
--- a/modules/core/server/jobs/send-push-message.server.job.js
+++ b/modules/core/server/jobs/send-push-message.server.job.js
@@ -18,46 +18,75 @@ module.exports = function(job, done) {
   var jobId = attrs._id.toString();
 
   var userId = data.userId;
-  var tokens = data.tokens;
-  var payload = data.payload;
+  var pushServices = data.pushServices;
+  var notification = data.notification;
 
-  // Log that we're sending an email
+  // Log that we're sending a notification
   log('debug', 'Starting `send push notification` job', { jobId: jobId });
 
-  firebaseMessaging.sendToDevice(tokens, payload)
-    .then(function(response) {
-      var unregisteredTokens = [];
-      response.results.forEach(function(result, idx) {
-        if (result.error) {
-          if (result.error.code === UNREGISTERED_TOKEN_ERROR_CODE) {
-            unregisteredTokens.push(tokens[idx]);
-          }
-        }
-      });
-      if (unregisteredTokens.length > 0) {
-        removeUserPushTokens(userId, unregisteredTokens, onFinish);
-      } else {
-        onFinish();
-      }
-    }).catch(onFinish);
+  // tokens for Firebase and Exponent
+  var firebaseTokens = [],
+      exponentTokens = [];
 
-  function onFinish(err) {
-    process.nextTick(function() {
-      if (err) {
+  // sort push tokens by cloud service
+  pushServices.forEach(function(pushService) {
+    if (pushService.platform === 'expo') {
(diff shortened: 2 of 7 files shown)


simison:
@bitfasching yo thanks for this! Was nice to hack together. 
I'm pushing this to producting tonight and publishing the mobile app to store :tada:

bitfasching:
You're welcome! :) Though maybe I wouldn't merge it to production yet: Outdated Exponent tokens won't be purged right now. It doesn't break anything as long as the app remembers its tokens, but in the long run unused tokens will pile up in the database.

Also, I haven't added tests to [send-push-message.server.job.tests.js](https://github.com/Trustroots/trustroots/blob/master/modules/core/tests/server/jobs/send-push-message.server.job.tests.js) yet. I'd check whether:
- tokens are correctly sorted by platform (Firebase/Expo)
- notifications are sent (perhaps using a proxy function like the one for Firebase)
- stray Exponent tokens are removed

Probably I won't have time to look into it until September. In case somebody else wants to implement the token removal: Expo [replies with receipts](https://docs.expo.io/versions/v14.0.0/guides/push-notifications.html#response-format) for pushed notifications. If it returns a `DeviceNotRegistered` error, the server should delete this token. This is similar to how the push service currently removes Firebase tokens, you can probably borrow some code from [there](https://github.com/Trustroots/trustroots/blob/f61b31a8b3991910ff8de3c6119471b8fe440e30/modules/core/server/jobs/send-push-message.server.job.js#L53-L67) and make it a function.

simison:
@bitfasching yep no worries, I think it's more important to get this rolling and under testing so that we can improve it, get feedback and learn what is important and what is less important. We generally try not to be afraid to break things so that we could push features online faster. ;-)

I monkey tested on dev and locally pretty well and I could see nothing is breaking so far; email notifications, desktop push notifications and FB notifications all roll out fine for unread messages. 

Just tested on prod as well; definitely ok for production, things are never in ideal perfect. ;-)

On mobile phone I got a ton of "push notifications enabled" messages when I'm enabling them at desktop. :-D I have to look into that.

I also want to work on notification settings view at account page and the whole "`login -> notify mobile app that user is available -> send token`" -process is still missing. Now mobile app registers token only on app restart, not when user logs in.

I'll take those tests as a first thing on my todo when I work on this, thanks for the list! :thumbsup: If you can think of more scenarios to test, feel free to list here.

bitfasching:
> On mobile phone I got a ton of "push notifications enabled" messages when I'm enabling them at desktop. :-D
Try to check the push registration user entry in the database. Multiple tokens for the same device will trigger multiple notifications. Ari and me observed this when playing around with Expo.

ilmirajat:
You get many notifications because in dev enviroment the react-native context seems to change build by build. I.e. always when you restart exp build it will generate new context to it (this is what I know for sure). I presume that if you get the app form app store it will have only one context but that is only a theory on basis of source code:

```
InstanceID instanceID = InstanceID.getInstance(this.getReactApplicationContext());
String gcmSenderId = Exponent.getInstance().getGCMSenderId();
if (gcmSenderId == null || gcmSenderId.length() == 0) {
    throw new InvalidParameterException("GCM Sender ID is null/empty");
}
final String token = instanceID.getToken(gcmSenderId, GoogleCloudMessaging.INSTANCE_ID_SCOPE, null);
```

Expo documentation is not exact here and I've not read all the relevant code. I did not check implementation of InstanceID. It's rather hard to track where all classes come form, as some of references seems to refer native Java classes. Hence, it's better to just try and figure out how it really works.

So, it is possible that it will generate new id each time your app starts. Notice that it will run in background because of notification logic, so it does not generate new idea each time you open app by clicking the icon.  Reference implementation for registration takes user name as a param. So, it fully possible that this is how it works - or id designed to work.

Anyway, for testing and dev you probably need to implement feature for removing Expo id's from admin UI. In dev, you cannot avoid registering multiple tokens. 

If the duplicates turn out to be a real issue. We need to implement replace token logic and use sender id as the key. Sender id should be unique (see code above) per device (not per device and app instance). This means two changes to endpoint:

Platform id should come in format Expo::[senderId]. If there is already token with same id, it should be removed so that you have only one active id at the time. New token is registered when app start, and you get new one when you start it again. We could store token to phone memory but if Expo notification service removes it, but we don't know then whether it is still properly registered in Expo. I.e. we cannot expect that Expo store token in future and it won't expire ever (it may store tokens indefinitely or not). It is more safe to ensure token each time app starts even if generated id is not unique. I also suspect that token may change per version even if it works correctly otherwise. Second change relates sending. Currently, it check if token platform is expo, it should be changes so that it checks if the format is the one I mention above.

simison:
I've restarted the app on my phone (the store version) several times now and I don't see extra tokens appearing in `pushRegistration`.

Yesterday:
```json
	"pushRegistration" : [
		{
			"token" : "####wm",
			"platform" : "web",
			"created" : ISODate("2017-07-20T21:32:27.756Z")
		},
		{
			"token" : "ExponentPushToken[####Ea]",
			"platform" : "expo",
			"created" : ISODate("2017-07-20T21:32:55.666Z")
		}
	],
```

Today:
```json
	"pushRegistration" : [
		{
			"token" : "####wm",
			"platform" : "web",
			"created" : ISODate("2017-07-20T21:32:27.756Z")
		},
		{
			"token" : "ExponentPushToken[####Ea]",
			"platform" : "expo",
			"created" : ISODate("2017-07-21T09:53:16.029Z")
		}
	]
```

Note that `created` has fresh timestamp.

simison:
How ever, after starting the app now, I get 3 x "You just enabled Trustroots mobile push notifications." :-) I guess I'll fix that before advertising the app more, any pointers what parts of the code modify welcome!

Issue #530, opened 2017-07-17
Find contact by typing in username or name


simison:
Show a list of members by typing in *username*, *firstName*, *lastName* or *displayName*.

- The API is already implemented at #531
- Markup can be basically the same as at contacts tab on the user profile
- First iteration doesn't need pagination, just hard limit (e.g. 10-30 or something)
- ~First iteration doesn't need to find from the middle of the name, only from the beginning~ We'll not implement partial search at all, see https://github.com/Trustroots/trustroots/pull/531#issuecomment-430018690
- Should not be case sensitive
- In future this page would hold also FB your friends who are also on TR (#392).
- Might need adding another Mongo index to *firstName*, *lastName* and *displayName*.

In navigation - on mobile:
![image](https://user-images.githubusercontent.com/87168/28264027-c2694742-6af1-11e7-8f1c-90c5cc892300.png)

In navigation - on desktop:
![image](https://user-images.githubusercontent.com/87168/28264058-e2440e58-6af1-11e7-842d-9812761e843e.png)

Label could be "Find friends".

The goal of this feature is:
- to find your existing contacts easier from Trustroots so that you can 1) add them as contacts 2) write to them.
- to not to find hosts or members to meet.

danidanitr:
We continue with this tomorrow :)

danidanitr:
So far:

**left side: icon of contacts + plus
right side: search through the contact list (shorter: search your contacts)
contact count (+pending) = centered
results = users appear with the same design**

![contacts_1](https://user-images.githubusercontent.com/43136016/47148191-5d96e200-d2d0-11e8-946d-43419d4edc36.jpg)

Then, when you type in "add new contacts..."
![contacts_2](https://user-images.githubusercontent.com/43136016/47148678-984d4a00-d2d1-11e8-8baf-1482fec77be4.jpg)

danidanitr:
So far:

left side: icon of contacts + plus
right side: search through the contact list (shorter: search your contacts)
contact count (+pending) = centered
results = users appear with the same design
![contacts_1](https://user-images.githubusercontent.com/43136016/47148972-8029fa80-d2d2-11e8-9464-b91ee7889a36.jpg)

Then, when you type in "add new contacts..."
![contacts_2](https://user-images.githubusercontent.com/43136016/47148983-861fdb80-d2d2-11e8-8856-f97d2f11ea2c.jpg)

danidanitr:
So far:

left side: icon of contacts + plus
[...]

PR #531, opened 2017-07-17
Search users: API

mrkvon:
Backend API part for _"searching members you already know"_ -feature #530

The endpoint is `GET /api/users?search=[searchString]`, where replace `[searchString]`.

The endpoint will support search by full names, (given and family) and full usernames. The support for partial search (_search as you type_) will be dropped.

### DIFF
--- a/modules/users/tests/server/search-users.server.routes.tests.js
+++ b/modules/users/tests/server/search-users.server.routes.tests.js
@@ -14,8 +14,7 @@ var request = require('supertest'),
 
 describe('Search users: GET /users?search=string', function () {
 
-  var agent,
-      sandbox;
+  var agent;
 
   var limit = 9;
 
@@ -34,13 +33,11 @@ describe('Search users: GET /users?search=string', function () {
   });
 
   beforeEach(function () {
-    sandbox = sinon.sandbox.create();
-
-    sandbox.stub(config.limits, 'userSearchLimit').value(limit);
+    sinon.stub(config.limits, 'userSearchLimit').value(limit);
   });
 
   afterEach(function () {
-    sandbox.restore();
+    sinon.restore();
   });
 
 
--- a/modules/users/tests/server/search-users.server.routes.tests.js
+++ b/modules/users/tests/server/search-users.server.routes.tests.js
@@ -422,8 +422,9 @@ describe('Search users: GET /users?search=string', function () {
       it('[query string is less than 3 characters long] respond with 400', function (done) {
         agent.get('/api/users?search=aa')
           .expect(400)
-          .end(function (err) {
-            // @TODO test the error message
+          .end(function (err, res) {
+            should(res.body.message).eql('Bad request.');
+            should(res.body.detail).eql('Query string should be at least 3 characters long.');
             done(err);
           });
       });
--- a/modules/users/tests/server/search-users.server.routes.tests.js
+++ b/modules/users/tests/server/search-users.server.routes.tests.js
@@ -10,7 +10,7 @@ var request = require('supertest'),
     User = mongoose.model('User'),
     express = require(path.resolve('./config/lib/express')),
     config = require(path.resolve('./config/config')),
-    userHandler = require(path.resolve('./modules/users/server/controllers/users.server.controller'));
+    userHandler = require(path.resolve('./modules/users/server/controllers/users.profile.server.controller'));
 
 describe('Search users: GET /users?search=string', function () {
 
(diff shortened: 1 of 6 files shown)


mrkvon:
The basic version is finished. Should we return the logged user (self) from the search?

`GET /api/users?search=string`

The query string can use `+` instead of a space.

simison:
Great! I'll have a look today.

>Should we return the logged user (self) from the search?

That's fine I think.

modules/users/server/controllers/users/users.profile.server.controller.js
```diff
@@ -1051,3 +1051,51 @@ exports.validateInviteCode = function(req, res) {
...
+          { username: queryRegexp },
+          { firstName: queryRegexp },
+          { lastName: queryRegexp },
+          { displayName: queryRegexp }
+        ]
+      }
+    ] })
+    // select only the right profile properties
+    .select(exports.userMiniProfileFields + ' -_id')
+    // todo what to choose for sorting?
```
simison: Sorting by `username` is fine. :thumbsup:

modules/users/tests/server/search-users.server.routes.tests.js
```diff
@@ -0,0 +1,432 @@
...
+        ], done);
+      });
+    });
+
+    context('invalid request', function () {
+      it('[query string is less than 3 characters long] respond with 400', function (done) {
+        agent.get('/api/users?search=aa')
+          .expect(400)
+          .end(function (err) {
+            // @TODO test the error message
```
simison:
```
.end(function (err, res) {

    res.body.message.should.equal('...Foo...');

});

mrkvon:
rebased with master

mrkvon:
I was just looking into [indexes to use](https://docs.mongodb.com/manual/reference/operator/query/regex/#index-use) and noticed this: 

> Case insensitive regular expression queries generally cannot use indexes effectively. The `$regex` implementation is not collation-aware and is unable to utilize case-insensitive indexes.

So indexes won't help in this case (unless we create lowercase versions of `username`, `displayName`, `firstName`, `lastName` with indexes and perform case sensitive lowercase search :unamused:).

simison:
This seems like worth exploring:  https://docs.mongodb.com/manual/core/index-case-insensitive/ <https://docs.mongodb.com/manual/core/index-case-insensitive/>

mrkvon:
@simison That seems like the case-insensitive index, which can't be used by regexp search, according to the above quote. o_o

simison:
Right; perhaps we could drop support for partial search but keep case-insensitive with that type of index.

I think MongoDB already treats stuff like `é` and `e` as same with text indices?

At this point, we could probably take the performance hit of regexp queries just fine but I'd rather not open an easy path for DoS attacks via expensive API endpoints.

mrkvon:
Ok. Thank you @simison for the specification. I'd like to leave this to sink in for a while. When the design and (API needs) for the frontend are clear and somebody is ready to implement it, I think I can fix this relatively quickly.

For the future reference, people who will use this endpoint: The support for the _search as you type_ functionality is deprecated. It won't work in production. Only the search by full given names, family names and usernames.

The endpoint is `GET /api/users?search=[searchString]`, where replace `[searchString]`.

simison:
@mrkvon sounds like a godo plan :thumbsup:

Akronix:
I'm up for taking the frontend part :raising_hand_man:
There was also a designer from the Iberian collective willing to do the design part. I'll ping him.

simison:
I'd recommend re-using HTML+CSS from contacts -tab as much as possible.

Regenwasser:
Continuing from #530 (but not quite sure if I'm at the right place here?): 

> Agreed not to have additional UX for this and build it in the existing API.

If we can do that I'd love to add keyword search right in there, even if it takes longer for it to appear on the site. Just like adding something to the URL structure would be sweet: 

`http://trustroots.org/search?keyword=placeholder`

That would be helpful for user research purposes, without having to go into any database.

guaka:
What's happening with this?

I don't think much design was done, but let's push this through asap - after that I'd be very happy to adapt it for users with `admin` (or rather `support`) role so that the search will be useful for this. (Using my hackety cli script for regex search on users all the time, to find someone's profile when they contact us just by email or even FB.)

mrkvon:
Rebased. I'll re-familiarize myself with this and try to update to fit the requirements.

mrkvon:
As of now, we're using a compound text index on `username`, `firstName` and `lastName`.

I just had a conversation with @Akronix how to improve this in a next iteration (pagination, ...). Let's follow up in an issue.

Also we may want another index (not text index) on `username`. I don't know how dots and dashes in usernames are treated now.

PR #534, opened 2017-07-24
Add user info to message thread mobile view

daiyi:
I'd like to expose some information about the person you're about to write to in the mobile view of message threads:

message-thread-mobile-after

The other views show the other person's name, but it's hidden in mobile and quite inconvenient.

Not sure about the design but somehow getting their name there would be excellent!

### DIFF
--- a/modules/messages/client/less/thread.less
+++ b/modules/messages/client/less/thread.less
@@ -16,11 +16,11 @@
 
   @media (min-width: @screen-xs-min) {
     .content-empty {
-      position: absolute;
-      bottom: 50%;
-      margin-top: -65px;
-      left: 0;
-      right: 0;
+      // position: absolute;
+      // bottom: 50%;
+      // margin-top: -65px;
+      // left: 0;
+      // right: 0;
     }
   }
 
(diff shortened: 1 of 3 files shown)


modules/messages/client/less/thread.less
```diff
@@ -16,11 +16,11 @@
 
   @media (min-width: @screen-xs-min) {
     .content-empty {
-      position: absolute;
-      bottom: 50%;
-      margin-top: -65px;
-      left: 0;
-      right: 0;
+      // position: absolute;
```
daiyi: I'm not sure what these styles were originally for. It didn't make a visible different after it was disabled.

simison: I'll pull this and test around to see if I notice any difference.

simison:
Looks & works great! Thanks.

modules/messages/client/less/thread.less
simison: I think it used to place things in the middle of the screen but has since been broken (should really be done in flex now anyway, but that was back in 2014).

daiyi: ah, makes sense!

daiyi:
\o/!

PR #535, opened 2017-07-24
doesn't work with node 8

daiyi:
Running `npm start` with node at latest couple versions of 8 (including the latest, 8.2.1) gives me a variety of compilation errors, so I'm removing the option from the install guide!

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,7 +6,7 @@ _These instructions are for installing locally. If you'd like to have containeri
 ## Prerequisites
 
 Make sure you have installed all these prerequisites:
-* [Node.js](https://nodejs.org/en/download/) v6 or v8 and the NPM package manager (`node --version && npm --version`). You can run multiple Node versions using [NVM](https://github.com/creationix/nvm).
+* [Node.js](https://nodejs.org/en/download/) v6 and the NPM package manager (`node --version && npm --version`). You can run multiple Node versions using [NVM](https://github.com/creationix/nvm).
 * [MongoDB](http://www.mongodb.org/downloads) v3 (`mongod --version`).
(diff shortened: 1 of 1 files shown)


simison:
That's interesting, which `NPM` version and which OS?

At Travis it's passing just fine on Linux.

We run v6 on production.

daiyi:
doesn't work for me: node 8.2.1 with npm 5.3.0 on Linux Mint 18 (Ubuntu 16.04 Xenial LTS)
works for me: node v6.11.1 with npm v3.10.10

simison:
On Travis this passes:
```
Ubuntu 14.04
Node 8.2.1
npm 5.3.0
```

Do you have `g++-4.8` & `gcc-4.8` compilers installed as it seems like Nodejs might require them?
 https://github.com/Trustroots/trustroots/blob/master/.travis.yml#L22-L24

Can you see which package is throwing errors? Might be this could be solved by simple `package.json` update. :-)

simison:
Aight, gave this a spin locally on Node 8 + NPM 5 and it works here.

For some weird reason I had to install `karma-phantomjs-launcher`, apparently some part of our Karma setup required it even for Chrome testing and NPM 3 just installed it previously, while NPM 5 didn't.

I also added `package-lock.json` for the exact versions that worked for me on OSX, Node 8 and NPM 5.

You could try removing node_modules and running `npm install` again, might be that will help?

simison:
FYI I've had lots of problems with NPM 5.3.0 (https://github.com/npm/npm/issues/17858). Downgrading to 5.2.0 helped.

simison:
Been running this quite a lot on Node.js 8 already myself. ;-)

PR #536, opened 2017-07-24
Remove profile

### LINKED ISSUE #273: Removing own profile
A feature for users to remove their own profile.

UI at the [account page](https://www.trustroots.org/profile/edit/account) is already there, just backend missing.

**Before removal:**
- Generate a token with our secret, user ID and timestamp, that we can decrypt (think JWT).
- Send confirmation mail with token-URL in it
- To access this URL you have to be logged in?
- Token expires in 1h
- Record two points to stats: 1) user removal initiated, 2) user removed. See how `statService` e.g. at [auth controller](https://github.com/Trustroots/trustroots/blob/master/modules/users/server/controllers/users/users.authentication.server.controller.js). 
- Ask for feedback why user is leaving. Checkboxes. See [the current form](http://ideas.trustroots.org/account-removal) — questions could be different in the future. Send checkbox data to Stats servers.

**Stuff to remove:**
- Profile
- Profile photo if it's uploaded (it's always `modules/users/img/profile/uploads/USER_ID/avatar/*.jpg`) so just removing that folder will do.
- Hosting offer
- Mark all messages sent _to_ that user as `notified:true` (so that unread-messages doesn't pick them up anymore). Leave messages _from_ that user as is.
- Contacts
- `-1` count for tribes user has joined
- Sessions (not sure how easy/possible/important) **Edit:** old sessions get cleaned out anyway over time, but maybe remove logic could "hit" logout logic? Or maybe it doesn't matter.

**Future development:**
- Send out an email to user after the profile removal.
- Collect freetext feedback, could be sent to our support mail.
- Prevent any user to sign up with that username afterwards? Leaving user object there would work for this, or some special reserved usernames table.
- Security: https://github.com/expressjs/csurf — would be great to have this, but JWT style tokens should be secure enough.

_All this said, feel free to implement this in some other quicker/better/easier way! This is just what we brainstormed so far._

simison:
Closes #273 

I'll do quick testing soon and get this merged. It's causing crazy workload at support with this stuff so wanna get this into production finally.


simison:
Seems to need rebasing.

mrkvon:
Fixed the conflict and rebased.
Removed some console.logs.

Sorry for being silent on this. I simply wasn't sure if/where to proceed. I took over this feature when it seemed (almost) finished. Then unclear what to do. Then forgot. :)

modules/users/tests/server/user-removal.server.routes.tests.js
```diff
@@ -0,0 +1,840 @@
...
+
+  });
+
+  // clear database
+  afterEach(function (done) {
+    var collectionsToClear = [User, Contact, Message, Offer, Tag];
+
+    async.each(collectionsToClear, function (collection, cb) {
+      collection.remove().exec(cb);
+    }, done);
```
simison: Smooth! :thumbsup:

scripts/rate-limit-test.sh
```diff
@@ -0,0 +1,14 @@
...
+
+#DOMAIN = "https://dev.trustroots.org"
+
+echo ""
+echo "Testing rate limiting..."
+
+for i in {0..20}; do
+  echo "Attempt $i..."
+  (curl -Is https://dev.trustroots.org/api/auth/signin | head -n1 &) 2>/dev/null; 
+done
```
simison: Oh, I didn't realise I pushed this here too. :-D

The actual implementation is done at Nginx but it's useful to have this script around, so I'll let it be here.

simison:
No worries! Yeah this is a bit fuzzy feature and I was also offline for a couple days.

Looks good! You did some great additions.

simison:
Ah damn, I was supposed to squash and merge but I just rebased and merged. Made a messy commit history because of that but oh well. :-) Next time.

mrkvon:
Oh, now i see what you meant by rebasing. I made an unclear comment above. Only fixed conflict, not `git rebase -i` with squash. Hmm.

simison:
I meant that I should've squashed it; github interface gives that option nowadays while merging. Back in times everyone had to squash their own commits but now we can be lazy as long as merger isn't the lazy one. ;-)

PR #543, opened 2017-08-16
Add new "meet" offer type

simison:
- New main navigation with "host" and "meet" sections
- Tabbed filter/results panel in search
- Better mobile view for search


daiyi:
So cool!

nicksellen:
Great!

So you create a "meetup" which is saying: _I am Nick, and you can meet me in this area for the next month (or until I delete the meetup) for the purpose of <description>_

simison:
Yep, the intro text is now:
>Traveling? Organising an event? Just making a dinner and would like to invite people over?
>
>Meetups stay visible on map for one month unless removed.

But there might be some other scenarios we just didn't think of and we don't want to guide it too much towards "events" or anything else specifically. Will be interesting to observe how people start using it!

We'll get out a blog post later today or tomorrow to elaborate more.

Issue #343, opened 2016-03-21
Timestamp last password change


simison:
For stats and possible small UX improvements.

trmendes:
Can I try this one?

simison:
Sure!

It'll require adding a new field to user Model: 

https://github.com/Trustroots/trustroots/blob/master/modules/users/server/models/user.server.model.js#L98

It should be type "Date", default to "now" and we'll need to update it here to `new Date()`:

https://github.com/Trustroots/trustroots/blob/master/modules/users/server/controllers/users.password.server.controller.js#L177-L180

...as well add to this list: https://github.com/Trustroots/trustroots/blob/master/modules/users/server/controllers/users.profile.server.controller.js#L380

simison:
Yep, I'm fixing that here: https://github.com/Trustroots/trustroots/pull/603 — once that's merged, you'll have to rebase your own branch (or just update `master` if you didn't create a branch).

PR #605, opened 2018-06-14
server: user models/controllers: Add passwordUpdated

trmendes:
Add the ability to keep the time when the user changed her/his
 password for the last time.

 issue #343


simison:
Tests as described. Thanks!

PR #614, opened 2018-07-09
Messages: do not send back messages that belongs to deleted users

trmendes:
Since we don't have a way to fetch messages from deleted users we won't
return those messages to the user who is checking her/his messages.
With this we don't have to mark every message sent by a deleted user as
read and as soon as we have a way to fetch those messages we can fix it
another way.

Issue #596

### DIFF
--- a/modules/messages/server/controllers/messages.server.controller.js
+++ b/modules/messages/server/controllers/messages.server.controller.js
@@ -69,19 +69,10 @@ function sanitizeMessages(messages) {
  * @returns {Array}
  */
 function sanitizeThreads(threads, authenticatedUserId) {
-
-  if (!threads || !threads.length) {
-    return [];
-  }
-
-  // Sanitize each outgoing thread
-  var threadsCleaned = [];
-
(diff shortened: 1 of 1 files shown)


trmendes:
@simison I'm wondering how is the write way to performe some tests on my code.

This patch fix the message counter for deleted messages and it will also prevent the backend to send messages from deleted users to the frontend. It does not allow the user to read messages from deleted users though :(. And because of that I'm not sure if this patch is a valid one. Like we talked, we might need to split the issue #596 into two. One, at first, to avoid counting and showing messages from deleted users and another one to create a new way to handle messages with no userFrom defined.

trmendes:
Any news here? Looking forward to fix any issue with this code to learn more and get other issues to work on.

simison:
See also https://github.com/Trustroots/trustroots/pull/633

simison:
Thanks @trmendes — I'd like the underlaying issue get fixed instead, but let's keep this open here if it turns out to be too difficult.

Issue #529, opened 2017-07-16
Last seen stats


simison:
This could just be something super simple that calculates average "last seen" data and pushes it to stats in [daily stats job](https://github.com/Trustroots/trustroots/blob/master/modules/statistics/server/jobs/daily-statistics.server.job.js).

Logic itself can be here: https://github.com/Trustroots/trustroots/blob/master/modules/statistics/server/controllers/statistics.server.controller.js

https://momentjs.com/ might be useful here.

_How many users logged last day, week, month, year, ... never),_ (_other ideas?_)

trmendes:
There is a problem with the 'users logged last day' which is:

 let's imagine that a 100 users logged into TR yesterday. So those 100 users will have seen = today - 1
 let's now say that half of the same users who logged into TR yesterday logged into TR today too. 
 Now we will have a statistic saying that we had 50 users logged users yesterday and 50 today which is wrong. 

 Shall we keep the yesterday statistic? Any better idea of how to sum how many users logged a day before today?

simison:
Follow-ups to https://github.com/Trustroots/trustroots/issues/755

PR #616, opened 2018-07-12
Statistics: generate "last seen" snapshot statistics about users

trmendes:
Processing some statistics about the user based on the time she/he
 logged in. With that TR can know how many users have been using the
 system during a period.

Issue #529


Tartesos:
Is that info useful for the user in the short term? I can see displaying that info being more useful in the long term, not sure right now.

My two cents

Carlos

trmendes:
@Tartesos Not sure how it could be useful for the user. Maybe just for those who are curious about how statistics and might be digging into the blog looking for some info. But back to this patch, all we have here are a few statistics about the use of the system by the users. That info might be useful for the system admin to check how the system is being used by its members as the time goes by.

Tartesos:
>  That info might be useful for the system admin to check how the system is being used by its members as the time goes by.
> 
How does that help growth?

Tartesos:
PS Sorry, I meant, how displaying the real time use of TR (if that is what I understood from the statistics that is projected to be displayed) does help the growth of the Community?

Just to make sure I understand. Thank you :-)

trmendes:
@Tartesos No worries! I like to discuss ideas and I like when people have question! :)

I believe those statistics are good in a long term. I was thinking about raise some data about how many users are living TR too. This could help the developers and system administrators to undestrand better how the user is interacting with TR. How many of them are coming back after registration, how many them are not, how many activities member TR has and how many of them are giving up from TR.

It is no big deal. It is just more info to try to find out ways to understand the use of TR better in a long term.

So I don't have a clean vision on how this could help growth in a short term but I see possibilities of analysys in a long term.

Though, this patch is just a small part of it. There are some statistics about the use of TR they are collected already and there are some that are not.

We could discuss more about what TR should have as statistic or not here: #529 (the issue that made me work on this PR).

Tartesos:
I agree completely with you :-) Yes, that data is essential for developers working on TR. Which, btw, it’s most of the job, so really thankful on this side :-)

Not sure if it’s the right time for the active users data display online for everyone. Yes, it’s important data, but, imho, also each data should be release when TR and its Community is ready. The more data out there does not necessarily mean that will help the Community. Only the adequate data in the adequate moment will help more people. But again, that is just my humble opinion.

modules/statistics/server/controllers/statistics.server.controller.js
```diff
@@ -187,6 +188,51 @@ exports.getPushRegistrationCount = function (callback) {
...
+/**
+ * Generate statistics based on the user last seen attribute
+ */
+exports.getLastSeenStatistic = function (type, callback) {
+  var now = moment();
+  var query = { };
+
+  switch (type) {
+    case 'never':
+      query.seen = { $exists: false };
```
simison: This will be a stable number (only very old profiles have this set to false) so we can drop this.

modules/statistics/server/controllers/statistics.server.controller.js
```diff
@@ -187,6 +188,51 @@ exports.getPushRegistrationCount = function (callback) {
...
+      break;
+    case 'sinceSixMonthsAgo':
+      var thisMonth = now.startOf('month');
+      query.seen = { '$gte': thisMonth.subtract(2, 'months') };
+      break;
+    case 'sinceAYearAgo':
+      var thisMonth = now.startOf('month');
+      query.seen = { '$gte': thisMonth.subtract(12, 'months') };
+      break;
+    default:
```
simison: Stats for the current ongoing month/week/day might not give us any real insight as a snapshot stat. Sliding window of "past X days" would be more useful metric and we can work on improving.

How about these sliding windows: 
- past 48h,
- past 7 days,
- past 14 days,
- past 30 days,
- past 6 months,
- past 365 days

modules/statistics/server/jobs/daily-statistics.server.job.js
```diff
@@ -65,6 +65,139 @@ module.exports = function (job, agendaDone) {
...
+          values: {
+            count: parseInt(count)
+          },
+          tags: {
+            access: 'members'
+          }
+        }, done)
+      });
+    },
+
```
simison: There's quite a bit of code-repetition here; would looping them differently work better while ensuring that each task is done before the next one starts so that we don't have lots of expensive queries and writes on the air simultaneously?

simison:
I'll re-open since this is a pretty important snapshot stat to have.

modules/statistics/server/jobs/daily-statistics.server.job.js
trmendes: Yes. There is a code repetition for evey time-stat we have.
If each task is async how could I be sure when to start the next one?

I mean. I could do a look that interates into an array of stats I want to collect. But in that case my poor knowlege of JS would get confused on "what about this return(done)?" What should I have in the case of a loop? I need some help here with the theory...

simison: It indeed gets a bit elaborate with ES5 Javascript and callbacks and how this whole file is structured, so maybe I'd just do this for now:

```
function (done) {
  collectLastSeen(7, done);
},
function (done) {
  collectLastSeen(14, done);
},
```

```js
function collectLastSeen(seenSinceDays, callback) {
  statistics.getLastSeenStatistic(seenSinceDays, function (err, count) {
    if (err) {
      return callback(err);
    }
    writeDailyStat({
      namespace: stat,
      values: {
        count: parseInt(count)
      },
      tags: {
        access: 'members'
      }
    }, callback);
  });
}
```

We can simplify this whole file later on when [we're on ES6 Javascript](https://github.com/Trustroots/trustroots/issues/804) and can get rid of our async helpers completely.

modules/statistics/server/controllers/statistics.server.controller.js
```diff
@@ -187,6 +188,51 @@ exports.getPushRegistrationCount = function (callback) {
...
+      return callback(new Error('Missing last seen statistic type.'));
+  }
+  User.count(query, function (err, count) {
+    if (err) {
+      callback(err);
+      return;
+    }
+    callback(null, parseInt(count) || 0);
```
simison: A nitpick: could you add `radix` number 10 to each `parseInt(count, 10)` ([You can read here  why](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description)).

simison:
Note [linter errors in Travis](https://travis-ci.org/Trustroots/trustroots/builds/450194151)

PR #630, opened 2018-08-19
Allow users change their usernames

simison:
Follow up for https://github.com/Trustroots/trustroots/pull/606 to have this here so that multiple folks can work on this at the same time.

cc @trmendes


simison:
I managed to fix https://github.com/Trustroots/trustroots/pull/606 and stacked my commits there instead. Closing this.

PR #633, opened 2018-08-20
Fix showing messages from deleted users

simison:
Resolves https://github.com/Trustroots/trustroots/issues/596

Allow reading messages from deleted users by:
- Passing user `_id` from the API
- Opening message thread with that id instead of username
- Loading messages with that _id even when `userFrom` does not resolve to any user _(still broken)_

Todo:
- [x] Fix frontend, might need some refactoring
- [x] Update tests

### Testing

- Create three users and send messages from one to two others
- Delete one of the user's profile
- See if you can still open the message thread
- Ensure for other users, `?userId=` bit doesn't appear in the URLs
- No JS errors in the console when looking at inbox and threads

### DIFF
--- a/modules/messages/client/components/Thread.component.js
+++ b/modules/messages/client/components/Thread.component.js
@@ -119,13 +119,12 @@ export default function Thread({ user, profileMinimumLength }) {
     return null; // important to return null to indicate "nothing to render"
   }
 
-  const chatWithUnknownUser = username === 'undefined';
-
   const [nextParams, setNextParams] = useState(null);
   const [isFetching, setIsFetching] = useState(false);
   const [isFetchingMore, setIsFetchingMore] = useState(false);
   const [otherUser, setOtherUser] = useState(null);
   const [doesNotExist, setDoesNotExist] = useState(false);
+  const [removed, setRemoved] = useState(false);
   const [messages, setMessages] = useState([]);
   const cacheKey = `messages.thread.${user._id}-${username}`;
 
@@ -137,8 +136,7 @@ export default function Thread({ user, profileMinimumLength }) {
   const userHasReplied = Boolean(
     messages.find(message => message.userFrom._id === user._id),
   );
-  const showReply =
-    (messages.length > 0 || !hasEmptyProfile) && !chatWithUnknownUser;
+  const showReply = (messages.length > 0 || !hasEmptyProfile) && !removed;
   const showQuickReply = showReply && !userHasReplied;
 
   const isExtraSmall = useMediaQuery({ maxWidth: 768 - 1 });
@@ -173,16 +171,27 @@ export default function Thread({ user, profileMinimumLength }) {
 
   async function fetchData() {
     if (isFetching) return;
-    const username = getRouteParams().username;
     try {
       setIsFetching(true);
       let otherUser;
-      if (username !== 'undefined') {
+      let userRemoved = false;
+      try {
         otherUser = await api.users.fetch(username);
-      } else {
-        const userId = getRouteParams().userId;
-        otherUser = createFakeUserObject(userId);
+      } catch (error) {
+        if (error.response?.status === 404) {
+          const userId = getRouteParams().userId;
+          if (userId !== undefined) {
+            otherUser = createFakeUserObject(userId);
+            userRemoved = true;
+          } else {
+            setDoesNotExist(true);
+            throw error;
+          }
+        } else {
+          throw error;
+        }
       }
+      setRemoved(userRemoved);
 
       const { messages, nextParams } = await api.messages.fetchMessages(
         otherUser._id,
@@ -192,7 +201,7 @@ export default function Thread({ user, profileMinimumLength }) {
         a.created.localeCompare(b.created),
       );
       // TODO should be done at the back-end?
-      if (chatWithUnknownUser) {
+      if (userRemoved) {
         const filledMessages = messages.map(message => {
           if (!message.userTo) {
             message.userTo = { _id: otherUser._id };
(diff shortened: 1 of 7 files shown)


modules/messages/server/controllers/messages.server.controller.js
```diff
@@ -63,57 +63,88 @@ function sanitizeMessages(messages) {
...
-      };
-    }
+/**
+ * Sanitize threads
+ *
+ * @param {Array} threads - list of threads to go trough
+ * @param {ObjectId} authenticatedUserId - ID of currently authenticated user
+ * @returns {Array}
+ */
+function sanitizeThreads(threads, authenticatedUserId, callback) {
```
simison: Refactoring `sanitizeThreads()` is prolly better to do outside this PR for clarity.

modules/messages/client/controllers/thread.client.controller.js
```diff
@@ -78,6 +79,12 @@
         vm.content = locker.driver('session').get(cachePrefix, '');
       }
 
+      userTo.$promise = {
+        then: function (callback) {
+          callback();
+        }
+      };
```
simison: Just a hack to get things rendering.

modules/messages/client/config/messages.client.routes.js
```diff
@@ -31,8 +31,27 @@
...
+          /* @ngInject */
+          userTo: function (UserProfilesService, $q, $stateParams) {
+
+            // Deleted users are passed in by just ID
+            if ($stateParams.userId) {
+              var deferred = $q.defer();
+              deferred.resolve({
+                _id: $stateParams.userId
+              });
+              return deferred.promise;
```
simison: Very much work-in-progress here.

guaka:
This works very well on my localhost with my data, tested twice with my own messages, looks good:

image

trmendes:
I get a console error (a few) saying 
```
Error: message.userFrom is null
messageRead@http://localhost:3000/modules/messages/controllers/thread.client.controller.js?c=58a3ce31283b:339:1
```

guaka:
What are the next steps here? I'd like to get rid of that red 2 :)

guaka:
Weird stuff when going to another unrelated profile where, empty messages from multiple random other profiles showing up when starting a thread:

image

(doing this with copy of prod database on local machine)

guaka:
same happens with http://localhost:3000/messages/mostusers

also seems to be the consistently the same profiles showing up as senders of the empty messages

and as a bonus it also shows for non-existing usernames (that have never existed)

image

hlibbabii:
@simison I guess we should reopen this, right? I can also look into it

simison:
Sure! The API changes might be salvageable here, but client-side work isn't usable (Angular vs React).

hlibbabii:
@simison Could you maybe check out this when you have some time so that I know if I'm moving in the right direction here?

simison:
@hlibbabii yap! sorry I missed this — I'll find time to understand the solution asap. Thanks for the nudge and patience!

simison:
Works pretty well! :-) 

I can open threads from deleted user 👍  Similarly if they have `shadowban` or `suspended` role 👍 

Some prop warnings in console when user profile is deleted:
Screenshot 2021-02-10 at 13 17 49

Also note undefined/missing username and link fields:

missing user

I think best would be on these cases:
- just show default avatar (done)
- don't link from avatar to profile if there's no user
- instead of nothing, show "Member not available" or something similar in place of name,

I think this would be fine as-is too, so I'm happy to get this merged and do follow-up for those.

hlibbabii:
Current ui of the conversation with removed user:

image

modules/messages/client/components/ThreadMessage.js
```diff
@@ -70,30 +70,34 @@ export default function ThreadMessage({ message, user }) {
...
         <div className="message-meta">
           {isMe(message.userFrom) ? (
             <span>{t('You')}</span>
-          ) : (
+          ) : !deletedUser ? (
             <a href={`/profile/${message.userFrom.username}`}>
               {message.userFrom.displayName}
             </a>
+          ) : (
+            <span>Unknown User</span>
```
simison, line 86: What do you think about doing "Unknown member" here instead, to be more in line with language across the site.

simison: Oh and needs translation function.

modules/messages/client/components/Thread.component.js
```diff
@@ -158,24 +159,63 @@ export default function Thread({ user, profileMinimumLength }) {
     }
   }
 
+  function createFakeUserObject(userId) {
+    return {
+      _id: userId,
+      displayName: 'Unknown User',
+      username: null,
+      member: [],
+      languages: [],
```
simison, line 168: I wonder if some of this can be avoided by doing checks like `user?.username`, `user?.languages?.length` in the code instead?

simison:
Tested deleted users, suspended users — all good, messages get marked as read.

When non-deleted user had written last message in the thread, server-side failed for checking with un-existing `_id`, so I fixed that. Also changed "Unknown User" to slightly friendlier "Unknown member". 

Thanks for helping to get this over finish line @hlibbabii ! 🙌

modules/messages/client/components/ThreadMessage.js
simison: Fixed in https://github.com/Trustroots/trustroots/pull/633/commits/756fd6feb7d636e4546f93bc1d0ea20009f32b35

simison:
Except that this doesn't now work for existing users 😁 

image

simison:
Added small notification to frontend when user is not available:

Screenshot 2021-02-28 at 16 51 45
Screenshot 2021-02-28 at 16 51 33
Screenshot 2021-02-28 at 16 51 21

PR #634, opened 2018-08-21
minimal upgrade to mongoose v5

### LINKED ISSUE #580: Upgrade compatibility with MongoDB 3.6
Mostly just aggregate queries that need updating ('cursor' option is required for those).

Mongoose update would be good as well (#539)

mrkvon:
Fixes #580 

todo:
take care of deprecation warnings

### DIFF
--- a/modules/core/tests/server/core.server.config.tests.js
+++ b/modules/core/tests/server/core.server.config.tests.js
@@ -105,7 +105,7 @@ describe('Configuration Tests:', function () {
               }
 
               // The user we just created should be exposed
-              res.text.should.containEql('user = {"_id":"' + userId + '",');
+              res.text.should.match(new RegExp('user = \\{.*"_id":"' + userId + '"'));
 
               return done();
             });
@@ -152,7 +152,7 @@ describe('Configuration Tests:', function () {
                 }
 
                 // The user we just created should be exposed
-                res.text.should.containEql('user = {"_id":"' + userId + '",');
+                res.text.should.match(new RegExp('user = \\{.*"_id":"' + userId + '"'));
 
                 Tribe.remove().exec(done);
               });
--- a/modules/core/tests/server/jobs/send-push-message.server.job.tests.js
+++ b/modules/core/tests/server/jobs/send-push-message.server.job.tests.js
@@ -235,7 +235,8 @@ describe('job: send push message', function () {
           var tokens = updatedUser.pushRegistration.map(function (reg) {
             return reg.token;
           });
-          tokens.should.deepEqual(['123', '456']);
+          // we need to convert CoreMongooseArray to Array
+          [].concat(tokens).should.deepEqual(['123', '456']);
 
           done();
         });
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "mmmagic": "~0.5.0",
     "moment": "~2.22.2",
     "mongodb": "~2.2.36",
-    "mongoose": "~4.9.10",
+    "mongoose": "^5.2.9",
     "mongoose-beautiful-unique-validation": "~5.1.1",
     "mongoose-integer": "~0.1.1",
     "mongoose-paginate": "5.0.3",
--- a/modules/contacts/tests/server/contact.server.model.tests.js
+++ b/modules/contacts/tests/server/contact.server.model.tests.js
@@ -45,13 +45,13 @@ describe('Contact Model Unit Tests:', function () {
     });
 
     // Create users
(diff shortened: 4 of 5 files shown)


simison:
Amazing, thank you for picking this up!

https://github.com/Trustroots/trustroots/pull/539 might have some useful bits (but it indeed has too much going on at the same time).

modules/core/tests/server/core.server.config.tests.js
```diff
@@ -105,7 +105,7 @@ describe('Configuration Tests:', function () {
               }
 
               // The user we just created should be exposed
-              res.text.should.containEql('user = {"_id":"' + userId + '",');
+              res.text.should.match(new RegExp('user = \\{.*"_id":"' + userId + '"'));
```
simison: Did some new fields appear before `_id` or did the order of fields change?

simison:
@mrkvon do you want to take care of deprecation warnings before I merge or fine to deal with them in a separate PR?

modules/core/tests/server/core.server.config.tests.js
mrkvon: Order of the fields seems to have changed.

modules/core/tests/server/jobs/send-push-message.server.job.tests.js
```diff
@@ -235,7 +235,8 @@ describe('job: send push message', function () {
           var tokens = updatedUser.pushRegistration.map(function (reg) {
             return reg.token;
           });
-          tokens.should.deepEqual(['123', '456']);
+          // we need to convert CoreMongooseArray to Array
+          [].concat(tokens).should.deepEqual(['123', '456']);
```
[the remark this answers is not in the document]
mrkvon: Updated with `Array.from()`. The proposed methods seem to not be present.

mrkvon:
The problem with deprecation warnings is they don't have a stack trace by default.
I'm happy with lazily not doing it now.

I fixed the `config/lib/worker.js` so it works with the v5 of mongoose (`npm start` was failing even though tests were passing). Then the tests started requiring `express.init(mongoose.connection)` => `express.init(mongoose)`. Not sure why.

package.json
```diff
@@ -11,7 +11,7 @@
   "engines": {
     "node": ">=8.0.0 <11",
     "npm": ">=5.0.0",
-    "mongodb": "~3.4"
+    "mongodb": ">=3.0 <=4.0"
```
mrkvon: This may need a second check. Not sure what should be here. https://mongoosejs.com/docs/compatibility.html

simison: Mongo tends to break between minor point releases (3.4 vs 3.5) so I'd keep this as strict as possible for now. Let's manually test and target 3.6

simison:
I'm most concerned about Aggregate queries, let's test them all manually. Listed them in https://github.com/Trustroots/trustroots/issues/580

package.json
mrkvon: I'm currently testing against mongodb v 4.0.0.

The #580 queries work. Tested manually.

simison: Excellent! :thumbsup: I'll do some testing as well and attempt to upgrade production to MongoDB v4 as well.

simison:
Have you tested if [`scripts/fillTestData.js`](https://github.com/Trustroots/trustroots/blob/06e5bdc3fe969d8d02d13797031fc017e0c6b036/scripts/fillTestData.js) still works with this? (If it even works in `master`, it's not covered by CI tests).

mrkvon:
Tested the fillTestData now, seems to work.
When creating 1000 users it works.
When creating 10000 users I get `RangeError: Maximum call stack size exceeded`. I believe that's an unrelated issue.

mrkvon:
There is also PR #539. However there are many changes I don't understand. I tried but was unable to rebase and work further with it.

simison:
>There is also PR #539. However there are many changes I don't understand. I tried but was unable to rebase and work further with it.

No worries; lots what was there has already been merged in other PRs.

mrkvon:
I have an issue with tracking the deprecation warnings. I added [`traceDeprecation: true`](https://mochajs.org/#usage) to [gulpfile](https://github.com/Trustroots/trustroots/blob/master/gulpfile.js#L424) and still no trace is shown. Any experience with this?

simison:
Note that our version of Mocha might be older than documentation. Can you describe the problem?

simison:
Also noting that the next steps include updating `mongodb` and `connect-mongo` packages.

Issue #676, opened 2018-09-04
grafana: Hosting offers not working


guaka:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/54728/45043651-0ca79300-b06e-11e8-985f-40da9c3867f9.png">

https://grafana.trustroots.org/d/000000019/offers?orgId=1

PR #678, opened 2018-09-04
Fix statistics collecting daily hosting offer stats

simison:
Fixes https://github.com/Trustroots/trustroots/issues/676

Results from the DB query were like: `{ yes: 1, maybe: 2, no: 3 }` so we had to iterate them in different way than other stats which are mostly arrays.


guaka:
What about using  `{ yes: 2, maybe: 1,  no: 0 }`, seems nicer numbers to handle. You could even sum them to get something meaningful.

PR #688, opened 2018-09-22
Run mongod with the other services

chreekat:
This might conflict with how other people run MongoDB if people are running it as a system service. If so, I still think it's a good idea, but perhaps the local service should be configured to use a non-standard port.

See https://github.com/chreekat/blatherings/blob/master/building-trustroots.md for rationale and other ideas.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,11 +9,11 @@
       "resolved": "https://registry.npmjs.org/@gulp-sourcemaps/identity-map/-/identity-map-1.0.1.tgz",
       "integrity": "sha1-z6I7xYQPkQTOMqZedNt+epdLvuE=",
       "requires": {
-        "acorn": "^5.0.3",
-        "css": "^2.2.1",
-        "normalize-path": "^2.1.1",
-        "source-map": "^0.5.6",
-        "through2": "^2.0.3"
+        "acorn": "5.3.0",
+        "css": "2.2.1",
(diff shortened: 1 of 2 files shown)


chreekat:
I committed the lockfile changes as well. Like I said in the commit message:

Many versions in the lockfile are not concrete. Running 'npm install'
modified the specified versions, making them concrete. That's what this
commit adds.

simison:
This is a good idea but there might indeed be differences how folks run MongoDB; some leave it running all the time in the backgroudn as a service, some prefer running together with the app like this etc.

To stay flexible, we could have a separate command that people can run but we shouldn't enforce the same way for everyone.

Our Docker setup contains MongoDb so that's probably better as a "start all in one command" setup for development; https://github.com/Trustroots/trustroots/blob/ecf0049f25fdb43aef82c480df204800f2c0639b/INSTALL-DOCKER.md

(Docker setup is broken right now, but there are a couple PRs to get it back alive, pending merging)

simison:
>I committed the lockfile changes as well

Yeah, our `package-lock` can be annoying because we define loose versions in our `package.json` (using `~` to get automatic updates for patches). I've been thinking of strict version locking everything but then it's more work to maintain. 🤷‍♂️ 

Running [`npm ci`](https://docs.npmjs.com/cli/ci) does `npm install` but does not update the lockfile.

PR #690, opened 2018-09-30
Update README.md

kbhutani0001:
PR FOR HACTOBERFEST


alxwrd:
@simison Might be worth labelling this as "invalid" so it doesn't count towards Hacktoberfest stats.

Issue #687, opened 2018-09-12
username in "Re: How is it going, Firstname?" emails


guaka:
Would be nice to have the user name in feedback emails, so we can check out their profile when replying, to get a better context.

Tartesos:
Yes, tart would be very nice :-)

PR #692, opened 2018-10-01
Added username to How is it going email.

lucbpz:
Solves issue #687.


guaka:
Hi @lucbpz, your addition is moving in the right direction but it only solves part of the issue.

The variable also needs to be used in the different templates such as https://github.com/Trustroots/trustroots/blob/master/modules/core/server/views/email-templates/welcome-sequence-third.server.view.html

Issue #641, opened 2018-08-24
Additional newline when inserting newline in a message


Anth0rx:
When writing messages to others and inserting a newline it is displayed as two newlines in the message history.

sharunspi:
show details regarding common places .

AbhyudayaSharma:
Hi! I'd like to work on this. Could I please get a code pointer?

jakubkarabula:
Hi @Anth0rx @simison 

I have investigated this issue. It's not showing "two newlines" but rather creates a new `<p>` element with every `enter` and `<br>` with every `shift + enter`.
There was a discussion on medium-editor issues page to allow to add simple `<br>` when using `enter`, you can see it here https://github.com/yabwe/medium-editor/issues/70
But they decided it's the expected behaviour and I don't see any easy way around it.

This second `<p>` element is shown lower because `<p>` in your project have assigned global css `margin: 0 0 10px;`

So to fix the unnecessary margin for the messages you or I can add something like `.panel-body  p { margin: initial; }` to css/less.
What do you think?

guaka:
@Anth0rx can you check if this fixes it for you?

PR #696, opened 2018-10-03
Reset margin for paragraphs in messages

jakubkarabula:
This is a proposed solution for issue https://github.com/Trustroots/trustroots/issues/641

The "new line" style is caused by a global style assigned to `<p>` element

Preview:
screenshot from 2018-10-03 17-16-00

### DIFF
--- a/modules/messages/client/less/thread.less
+++ b/modules/messages/client/less/thread.less
@@ -99,6 +99,10 @@
     }
   }
 
+  .panel-body p {
+    margin: auto;
+  }
+
   // "Yes, I can host!" messages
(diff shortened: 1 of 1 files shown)


guaka:
@Anth0rx I meant this, this is the patch (simple CSS).

petr-hajek:
@jakubkarabula With this change there would be no space between paragraphs when the user makes double enter. Users would need to press shift-enter to make space between paragraphs.
image

simison:
I'll close this one as it doesn't solve the problem in #641 and it's kinda old PR already.

Feel free to re-open if you'd like to continue working on this tho!

PR #615, opened 2018-07-10
work in progress: server/client/contact list: Show location info about your friends

trmendes:
In order to make it easy to find frineds who are living in or were born
in a city/country you are in or have interest to visit TR now lists
those infos from each friend you have so you don't have to check one by
one.

Resolves #612

Issue #395, opened 2016-08-17
Make from/living location fields store actual geo info


simison:
We'll need coordinates.

That said, I think it'd be great to let people fill in freetext like "Mars".

Related to #359

mrkvon:
When filling the names of places to `From` and `Location` fields, the [api.mapbox.com](https://api.mapbox.com/geocoding/v5/mapbox.places/Vaasa.json?access_token=pk.eyJ1IjoidHJ1c3Ryb290cyIsImEiOiJVWFFGa19BIn0.4e59q4-7e8yvgvcd1jzF4g&types=country,region,place,locality,neighborhood) already provides coordinates and everything. So it seems to be a question of storing these data in database. (and maybe some script to update the locations saved earlier).

And if the user doesn't select the place from the provided list, then there can be a little warning for user and storing a boolean indicating that the name is not an actual geolocation (so it would be treated accordingly).

simison:
Yep or just `FromLabel` (String) and `From` (Object) to keep it simple. Mongo would even support mixing String/Object but that's probably not smart.

simison:
Noting here that this is more about the flow/UI than it's about technical solution. As @mrkvon pointed out above, the technical side of this is fairly straightforward (and could indeed be the first step).

Regenwasser:
So, if some wireframes are needed here to bring this further I think we should have a chat or call about it. (This seems to be connected to other location features, etc.).

> and could indeed be the first step 

Yes, good idea! If it's fast and easy, better do this first to get some users onto the map.

Maybe the first step could include: 
- make the text run through some geolocator (filter out all countries and things that weren't found)
- add a higher randomisation, so it's not just clustered around city centers.
- change field to Autocomplete addresses and places with something like [this](https://developers.google.com/maps/documentation/javascript/places-autocomplete). Right now only cities are auto-completed. Streets are not recognised, so I assume not many entered them).

PR #699, opened 2018-10-12
#612 adding location to miniprofiles at /profile/USER/contacts

### LINKED ISSUE #612: Show more info of contacts: location and contact status
It would be very useful - and quite simple to add - to see the place and country where contacts of someone are located.

E.g. now I'm looking for people in Vienna, and I know my friend has lived there, so she must have friends in Vienna, but now I have to open 22 profiles to see whom of her contacts are in Vienna.

https://www.trustroots.org/profile/*/contacts

image

While at it it would be a simple thing to also add whether a person is already a contact.
(This is useful to know if you should still add someone you know. Now I'm not sure whether I added a person and I have to open their profile to check)

guaka:
Resolves #612

### DIFF
--- a/modules/contacts/client/views/directives/tr-contact.client.view.html
+++ b/modules/contacts/client/views/directives/tr-contact.client.view.html
@@ -12,6 +12,11 @@
     <span ng-if="::contactCtrl.contact.confirmed === false">Requested</span>
     <time ng-bind="::contactCtrl.contact.created | date:'mediumDate'"></time>
   </small>
+  <div>
+    <small class="text-muted">
+      <span ng-bind="::contactCtrl.contact.user.locationLiving"></span>
+    </small>
+  </div>
 
   <!-- Authenticated user requested this connection -->
   <div class="contact-confirm"
--- a/modules/contacts/server/controllers/contacts.server.controller.js
+++ b/modules/contacts/server/controllers/contacts.server.controller.js
@@ -476,6 +476,7 @@ exports.contactListByUser = function (req, res, next, listUserId) {
           avatarSource: '$user.avatarSource',
           avatarUploaded: '$user.avatarUploaded',
           emailHash: '$user.emailHash',
+          locationLiving: '$user.locationLiving',
           additionalProvidersData: {
             facebook: {
               id: '$user.additionalProvidersData.facebook.id'
--- a/modules/users/server/controllers/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users.profile.server.controller.js
@@ -80,6 +80,7 @@ exports.userMiniProfileFields = [
   'avatarSource',
   'avatarUploaded',
   'emailHash',
+  'locationLiving',
   'additionalProvidersData.facebook.id' // For FB avatars
 ].join(' ');
 


modules/contacts/client/views/directives/tr-contact.client.view.html
```diff
@@ -12,6 +12,11 @@ <h4>
     <span ng-if="::contactCtrl.contact.confirmed === false">Requested</span>
     <time ng-bind="::contactCtrl.contact.created | date:'mediumDate'"></time>
   </small>
+  <div>
+    <small class="text-muted">
```
mrkvon: The `small` tag could be replaced with css? However the `small` tag is used in multiple places, therefore it perhaps doesn't matter?

modules/contacts/server/controllers/contacts.server.controller.js
```diff
@@ -476,6 +476,7 @@ exports.contactListByUser = function (req, res, next, listUserId) {
           avatarSource: '$user.avatarSource',
           avatarUploaded: '$user.avatarUploaded',
           emailHash: '$user.emailHash',
+          locationLiving: '$user.locationLiving',
```
mrkvon: If I understand correctly, `locationLiving` is an independent information from the hosting offer, isn't it? Somehow related to #395 and the confusion of `locationFrom`, `locationLiving` and `Offer.location`. I don't say that it should be fixed here, just to keep that in mind. :)

modules/contacts/client/views/directives/tr-contact.client.view.html
guaka: yeah it doesn't matter for now, but maybe we can replace all `<small>` tags at some point

modules/contacts/server/controllers/contacts.server.controller.js
guaka: Yes, correct.

There should probably be third location field, `locationCurrent`, which should really only be coordinates (in the database), and which is even more meaningful for `/profile/USER/contacts`
(And then we can make `locationFrom` a bit less prominent because it's not very relevant for the trustroots experience:)

guaka:
@mrkvon I added a test. Wanna merge this?

guaka:
@mrkvon (correction: I fixed the test for #700)

I am trying to add a test for this but adding           `contactsGetRes.body[user1Order].user.locationLiving.should.equal(user1.locationLiving);` doesn't work. Both sides don't actually have `locationLiving`.

mrkvon:
I guess `locationLiving` of the user will need to be set [here](https://github.com/Trustroots/trustroots/blob/master/modules/contacts/tests/server/contact.server.routes.tests.js#L56).

When it comes to merging, I don't have the rights to do so, but @guaka you do, don't you?

modules/contacts/client/views/directives/tr-contact.client.view.html
simison: Could all 3 tags be replaced with just one?

`<small ng-bind="..."></small>`

modules/users/server/controllers/users.profile.server.controller.js
```diff
@@ -80,6 +80,7 @@ exports.userMiniProfileFields = [
   'avatarSource',
   'avatarUploaded',
   'emailHash',
+  'locationLiving',
```
simison: There's an API test somewhere for this endpoint, could you add this to there as well?

simison:
Pretty much the same feature implemented differently in https://github.com/Trustroots/trustroots/pull/615

simison:
Done in https://github.com/Trustroots/trustroots/pull/615 :tada:

PR #708, opened 2018-10-13
Add seen field to test users.

mrkvon:
In `scripts/fillTestData.js` the users get a random `seen` date within last year.
Otherwise they weren't searchable, because they were always filtered out either with `"seen":{"months":24}` or `"seen":{"months":6}`.


simison:
CI failed for an unrelated flakey test.

mrkvon:
This change disappeared in #740

PR #710, opened 2018-10-14
Fix one off error while adding users

Jagjeet:
When seeding the database, N+1 users were being added instead of the expected N.


Akronix:
LGTM

PR #712, opened 2018-10-14
Update Mocha

simison:
Work-in-progress.

Updates `gulp-mocha` to latest.

Hitting issue with how Mongoose is structured currently.

_Feel free to take over if you're looking for something to work on!_ ;-)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -419,30 +419,30 @@ function mocha(done) {
   mongooseService.connect(function (db) {
     // Clean out test database to have clean base
     mongooseService.dropDatabase(db, function () {
-      mongooseService.loadModels();
-
-      // Run the tests
-      gulp.src(testSuites)
-        .pipe(plugins.mocha({
-          reporter: 'spec',
-          timeout: 10000
-        }))
-        .on('error', function (err) {
-          // If an error occurs, save it
-          error = err;
-          console.error(err);
-        })
-        .on('end', function () {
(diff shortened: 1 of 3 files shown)


Jagjeet:
I poked into this and made a little progress on this with regard to loading models, but after that it seems like the configuration is not loading properly. Still trying to investigate what is going on. 

https://github.com/Jagjeet/trustroots/tree/test-framework-deps

simison:
@Jagjeet Nice, I like the approach you took with `file`. Feel free to open a work-in-progress PR!

Here's a bit easier to compare view:

https://github.com/Trustroots/trustroots/compare/master...Jagjeet:test-framework-deps

simison:
@Jagjeet I was looking into this again — did you have a branch running with some changes? Can you open a work-in-progress PR from it? No worries if you don't have time to work on it, others can do it as well but would be ace to have a better starting point than my PR. :-)

mrkvon:
@simison Since we don't use mocha anymore, can this PR be closed? Instead we may remove gulp-mocha completely, i suppose.

nicksellen:
mocha is still used for the server tests. I did look into use jest on the server too, it's not so commonly used for server tests, but it mostly worked to that extent that I tried it. mongoose lib actually [strongly suggests](https://mongoosejs.com/docs/jest.html) that you use mocha instead of jest.

mrkvon:
@nicksellen Thank you for clarifying!

PR #714, opened 2018-10-15
Upgrade sinon

mrkvon:
- upgrade to sinon v7.0.0
- fix useFakeTimers
  https://sinonjs.org/releases/latest/fake-timers/#var-clock--sinonusefaketimersconfig
- use default sandbox (`sinon` object) instead of `sinon.sandbox.create()`
  https://sinonjs.org/releases/latest/sandbox/#default-sandbox

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,7 @@
     "proxyquire": "~2.1.0",
     "semver": "~5.6.0",
     "should": "~13.2.3",
-    "sinon": "^7.0.0",
+    "sinon": "~7.0.0",
     "supertest": "~3.3.0"
   },
   "bugs": {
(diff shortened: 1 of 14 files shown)


mrkvon:
Something I had already done locally.

package.json
```diff
@@ -154,7 +154,7 @@
     "proxyquire": "~2.1.0",
     "semver": "~5.6.0",
     "should": "~13.2.3",
-    "sinon": "~2.3.7",
+    "sinon": "^7.0.0",
```
simison: Nitpick; let's change this to `~` so that we'll auto-update only patch versions, not minor versions.

PR #716, opened 2018-10-16
Added bower to prerequisites

petr-hajek:


petr-hajek:
@simison You are right. There was an error during install and when I subsequently run `npm start` it skipped the `npm install` part, because node_modules was not empty. Also this would happen if an user runs `npm update` before `npm start`

I'd suggest:
a) add note to the install.md that user should run `npm install` manually if there has been some issue (maybe not obvious for newbies).
b) extend the check in "install-if-no-modules" script not only for node_modules folder but also public/lib folder

petr-hajek:
OK, I will do both. Closing this one.

simison:
What we _really_ should do is get rid of Bower completely and install those debs via `npm`, but that's much more work at this point and will get done at any point when we might start rewriting the frontend in something else than Angular v1.

PR #717, opened 2018-10-16
prestart script changed to update bower 

petr-hajek:
& should run npm install also when the install was not finished in previous run (public/lib was not created during post-install)


simison:
Closing in favor of https://github.com/Trustroots/trustroots/pull/784

PR #720, opened 2018-10-17
Show User Not Found in client

### LINKED ISSUE #669: Show "no profile found" for missing & deleted profiles
https://www.trustroots.org/profile/abcdefdoesnotexist
image

it would be nice to have some more verbosity when a profile is not found

mrkvon:
When looking at `/profile/:nonexistent-username` and receiving `404` status from API, angular resolver resolves with empty object and allows error to be displayed.
Before the error lead to resolver not being resolved, and empty page being displayed.

Fixes #669

Coauthored by @Akronix :clap:

### DIFF
--- a/modules/users/client/config/users.client.routes.js
+++ b/modules/users/client/config/users.client.routes.js
@@ -149,15 +149,30 @@
             return SettingsService.get();
           },
 
-          profile: function (UserProfilesService, $stateParams) {
+          profile: function (UserProfilesService, $stateParams, $q) {
             return UserProfilesService.get({
               username: $stateParams.username
-            });
+            }).$promise
+              .catch(function (e) {
+
+                if (e.status === 404) {
+                  // when user was not found, resolving with empty user profile, in order to display the User Not Found error.
+                  return { $promise: $q.resolve({ }), $resolved: true };
+                }
+
+                throw e;
+              });
           },
 
           // Contact is loaded only after profile is loaded, because we need the profile ID
           contact: function (ContactByService, profile, Authentication) {
             return profile.$promise.then(function (profile) {
+
+              // when user doesn't exist, no need to load contact
+              if (!profile._id) {
+                return;
+              }
+
               if (Authentication.user && Authentication.user._id === profile._id) {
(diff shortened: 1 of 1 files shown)


modules/users/client/config/users.client.routes.js
```diff
@@ -149,15 +149,30 @@
...
+                throw e;
+              });
           },
 
           // Contact is loaded only after profile is loaded, because we need the profile ID
           contact: function (ContactByService, profile, Authentication) {
             return profile.$promise.then(function (profile) {
+
+              // when user doesn't exist, no need to load contact
+              if (!profile.id) {
```
simison: Shouldn't this be `profile._id` ?

Akronix:
Should we write a test for this as well?

modules/users/client/config/users.client.routes.js
mrkvon: Very much indeed. This would not load any contacts of any users. :astonished:

mrkvon:
@Akronix I was feeling lazy, but I guess that's a very good idea. :+1:

simison:
A test would be ace, considering this broke earlier. Don't hold it as a blocker though; our frontend tests are very minimal since we've been trying to concentrate covering backend API with tests instead.

This is where you'd write it: https://github.com/Trustroots/trustroots/tree/master/modules/users/tests/client

Akronix:
So I'd say to merge this and write the test at any other time. At least @mrkvon feels motivated to do it now.

mrkvon:
TLDR:
I'm interested to learn writing tests for TR client, but it might take some time.
If you don't mind merging this without tests, go on.

---
Writing tests for TR frontend: Exciting, I haven't done it before.

Ideally I'd like to test: when user enters a path `/profile/:nonexistent-username`, show `User was not found` on the screen.
I imagine this would rather be an integration test or even e2e test, because it involves router, resolvers and component.

If somebody has examples of tests like this, or other advice, I'll be grateful to receive. In the meantime I'll learn, experiment with this, and do other stuff with higher priority. :slightly_smiling_face:

Issue #718, opened 2018-10-16
References


simison:
Members should be able to leave references for each other.

_This is as a high-level thread to keep track of all the plans around this feature. Other issues and PRs should be linked here for reference and detailed feature discussions should go into those instead._

---

# Implementation

UI for references will live inside profile pages:

<img width="300" alt="image" src="https://user-images.githubusercontent.com/87168/47045264-0fcc8d80-d19b-11e8-96ff-e1635d76d841.png">

Technically we'll call these "user references" or just "references", in contrast to "thread references" which are visible next to message threads for bigger screens:

<img width="250" alt="image" src="https://user-images.githubusercontent.com/87168/47040412-dd1c9800-d18e-11e8-857c-9009610d3863.png">

During the 1st implementation, we'll probably want to merge some parts to continue working on other parts. We can hide the feature using "feature flags", which is just an object in `config/env/*.js` files with `true/false` values.

We can divide the work roughly into four parts and probably into smaller sub-parts. Writing a references API+UI and reading references API+UI. Smaller the PRs the easier reviews, and quicker we can work on this and multiple people can work together at the same time.

There's some prior experimental work done in https://github.com/Trustroots/trustroots/pull/649

Preparation work before 1st iteration: https://github.com/Trustroots/trustroots/pull/719

#### User can leave a reference to anyone:
- They're _not_ required to have message exchange first
- Both need to have "email confirmed"; `user.public` has to be `true`

#### References contain:
- _"How do you know them?"_ -info:
    - _"Met in person"_ (level 1)
    - _"I hosted them"_ (level 2)
    - _"They hosted me"_ (level 3)

    They can choose 1-to-many, but they _have_ to choose at least one or they cannot leave a reference.

    <img width="575" alt="screen shot 2018-10-16 at 23 35 48" src="https://user-images.githubusercontent.com/87168/47046287-7488e780-d19d-11e8-95d7-15d2cb6ab58f.png">

- Recommendation: depending what level they choose, question differs on next level:
    - _"Would you recommend others to meet them?"_ (level 1)
    - _"Would you recommend others to host them?"_ (level 2)
    - _"Would you recommend others to stay with them?"_ (level 3)

    In the DB we can still save this just as one `recommend: true|false`.

    <img width="569" alt="screen shot 2018-10-16 at 23 35 56" src="https://user-images.githubusercontent.com/87168/47046337-8e2a2f00-d19d-11e8-82e6-5877501145cd.png">

#### After writing the reference
[...]

PR #723, opened 2018-10-18
Create a reference (API)

mrkvon:
`POST /api/references`
```js
{
  userTo: UserId,
  interactions: {
    met: Boolean,
    hostedMe: Boolean,
    hostedThem: Boolean
  },
  recommend: Enum('yes', 'no', 'unknown')
}
```

Implementing parts of #718
Follow up: #759 #761 

# Behavior

1. A public user (`userFrom`) can create a reference for any other public user (`userTo`).
  - only one reference from `userFrom` to `userTo` can be given
  - references can't be edited or removed
2. The reference is not public. Email and push notifications are sent to the `userTo`. `userTo` has a limited time period (currently [14 days](https://github.com/Trustroots/trustroots/blob/references-create-api/config/env/default.js#L107)) to give a reference back.
3. When the `userTo` gives a reference to `userFrom`, both references become public. `userFrom` is notified that she received a reference and can see it.
4. If `userTo` doesn't give the reference back within the time period, the original reference will become public. (_TODO later_ both users should get notified about this).
5. When the reference is public, a new opposite direction reference can be positive only. `recommend: 'yes'`

Rationale: users should be able to give a neutral or negative reference and not receive a reciprocal negative reference.

# Notes
- configurables:
  - `limits.timeToReplyReference` ([./config/env/default.js](https://github.com/Trustroots/trustroots/blob/references-create-api/config/env/default.js#L107), currently 14 days)
  - period of repeating the `publish old unpublished references` job ([./config/lib/worker.js](https://github.com/Trustroots/trustroots/blob/references-create-api/config/lib/worker.js#L95), currently 23 minutes (very random choice))

- Email notifications will need some :art:, :heart: and :thought_balloon: from a designer. This kind of stuff is frustrating for me and somebody else may enjoy it.
  I would do it as a last resort only.
  - Email notification: First Reference between users was created. [html template](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/core/server/views/email-templates/reference-notification-first.server.view.html), [text template](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/core/server/views/email-templates-text/reference-notification-first.server.view.html)
  - Email notification: Second Reference between users was created. [html template](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/core/server/views/email-templates/reference-notification-second.server.view.html), [text template](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/core/server/views/email-templates-text/reference-notification-second.server.view.html)

- Push notifications' text needs some :heart: too. Currently the text is:
  - First reference: `<username> gave you a new reference. Give a reference back.` ([test](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/references/tests/server/reference-create.server.routes.tests.js#L433))
  - Second (reply) reference: `<username> gave you a new reference. You can see it.` ([test](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/references/tests/server/reference-create.server.routes.tests.js#L676))
  - [implementation here](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/core/server/services/push.server.service.js#L95)

- Implementation of _No duplicate references!_ relies on MongoDB unique compound index on fields `userFrom` and `userTo`. ([code here](https://github.com/Trustroots/trustroots/blob/references-create-api/modules/references/server/models/reference.server.model.js#L77))
  Therefore, this index should be created directly in MongoDB when merged to production, because [mongoose documentation says so](https://mongoosejs.com/docs/faq.html#unique-doesnt-work). Or not. Just make sure it's present.
  Alternatively, the checks could be done before (more reads from database).

- still missing notifications to users when given time to reply passes and reference thus becomes public

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -93,6 +93,8 @@ module.exports = {
     // How long we should wait before trying to reactivate "no" hosts?
     // Moment.js `duration` object literal http://momentjs.com/docs/#/durations/
     timeToReactivateHosts: { days: 90 },
+    // How long should user have for replying a reference before it becomes public?
+    timeToReplyReference: { days: 14 },
     // How long should we wait to update user's seen field since the last update
     timeToUpdateLastSeenUser: { minutes: 5 },
     // when to send reminders about unread messages (since the last unread message was sent)
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -18,6 +18,16 @@ module.exports = {
     description: 'Travellers community for sharing, hosting and getting people together. We want a world that encourages trust and adventure.'
   },
 
+  // Feature flags
+  featureFlags: {
+    // enable references at all?
+    reference: false
+    // Tribe endorsements for user references
+    // referenceTribeEndorsements: false,
+    // Free text feedback in references
+    // referenceFreetextFeedback: false
+  },
+
   // Is site invitation only?
   invitations: {
     enabled: false,
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -20,12 +20,8 @@ module.exports = {
 
   // Feature flags
   featureFlags: {
-    // enable references at all?
+    // enable references?
     reference: false
-    // Tribe endorsements for user references
-    // referenceTribeEndorsements: false,
-    // Free text feedback in references
-    // referenceFreetextFeedback: false
   },
 
   // Is site invitation only?
--- a/config/lib/worker.js
+++ b/config/lib/worker.js
@@ -77,6 +77,12 @@ exports.start = function (options, callback) {
       require(path.resolve('./modules/users/server/jobs/user-welcome-sequence-third.server.job'))
     );
 
+    agenda.define(
+      'publish old unpublished references',
+      { lockLifetime: 10000, concurrency: 1 },
+      require(path.resolve('./modules/references/server/jobs/references-publish.server.job'))
+    );
+
     // Schedule job(s)
 
     agenda.every('5 minutes', 'check unread messages');
@@ -86,6 +92,7 @@ exports.start = function (options, callback) {
     agenda.every('15 minutes', 'welcome sequence first');
     agenda.every('60 minutes', 'welcome sequence second');
     agenda.every('60 minutes', 'welcome sequence third');
+    agenda.every('23 minutes', 'publish old unpublished references');
 
     // Start worker
 
--- a/modules/core/server/services/email.server.service.js
+++ b/modules/core/server/services/email.server.service.js
@@ -410,6 +410,21 @@ exports.sendReferenceNotificationFirst = function (userFrom, userTo, callback) {
 };
 
 /**
+ * Reference Notification (Second reference between users)
+ */
+exports.sendReferenceNotificationSecond = function (userFrom, userTo, reference, callback) {
+
+  var params = exports.addEmailBaseTemplateParams({
+    subject: 'New reference from ' + userFrom.username,
+    email: userTo.email,
+    seeReferencesUrl: url + '/profile/' + userTo.username + '/references',
+    recommend: reference.recommend
+  });
+
+  exports.renderEmailAndSend('reference-notification-second', params, callback);
+};
+
+/**
  * Add several parameters to be used to render transactional emails
  * These variables are used by email base template:
  * `modules/core/server/views/email-templates/email.server.view.html`
--- a/modules/core/server/services/email.server.service.js
+++ b/modules/core/server/services/email.server.service.js
@@ -396,6 +396,20 @@ exports.sendWelcomeSequenceThird = function (user, callback) {
 };
 
 /**
+ * Reference Notification (First between users)
+ */
+exports.sendReferenceNotificationFirst = function (userFrom, userTo, callback) {
+
+  var params = exports.addEmailBaseTemplateParams({
+    subject: 'New reference from ' + userFrom.username,
+    email: userTo.email,
+    giveReferenceUrl: url + '/profile/' + userFrom.username + '/references/new'
+  });
+
+  exports.renderEmailAndSend('reference-notification-first', params, callback);
+};
+
+/**
  * Add several parameters to be used to render transactional emails
  * These variables are used by email base template:
  * `modules/core/server/views/email-templates/email.server.view.html`
--- a/modules/core/server/services/push.server.service.js
+++ b/modules/core/server/services/push.server.service.js
@@ -85,6 +85,35 @@ exports.notifyMessagesUnread = function (userFrom, userTo, data, callback) {
   exports.sendUserNotification(userTo, notification, callback);
 };
 
+/**
+ * Send a push notification about a new reference, to the receiver of the reference
+ * @param {User} userFrom - user who gave the reference
+ * @param {User} userTo - user who received the reference
+ * @param {Object} data - notification config
+ * @param {boolean} data.isFirst - is it the first reference between users?
+ */
+exports.notifyNewReference = function (userFrom, userTo, data, callback) {
+  var giveReferenceUrl = url + '/profile/' + userFrom.username + '/references/new';
+  var readReferencesUrl = url + '/profile/' + userTo.username + '/references';
+
+  // When the reference is first, reply reference can be given.
+  // Otherwise both references are public now and can be seen.
+  var actionText = (data.isFirst) ? 'Give a reference back.' : 'You can see it.';
+  var actionUrl = (data.isFirst) ? giveReferenceUrl : readReferencesUrl;
+
+  var notification = {
+    title: 'Trustroots',
+    body: userFrom.username + ' gave you a new reference. ' + actionText,
+    click_action: analyticsHandler.appendUTMParams(actionUrl, {
+      source: 'push-notification',
+      medium: 'fcm',
+      campaign: 'new-reference',
+      content: 'reply-to' // @TODO what are the correct parameters here? What do they mean?
+    })
+  };
+  exports.sendUserNotification(userTo, notification, callback);
+};
+
 exports.sendUserNotification = function (user, notification, callback) {
 
   var data = {
--- a/modules/core/server/views/email-templates-text/reference-notification-first.server.view.html
+++ b/modules/core/server/views/email-templates-text/reference-notification-first.server.view.html
@@ -1,0 +1,6 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+give the reference, too: {{giveReferenceUrl}}
+And give more explanation about time limit etc.
+{% endblock content %}
--- a/modules/core/server/views/email-templates-text/reference-notification-second.server.view.html
+++ b/modules/core/server/views/email-templates-text/reference-notification-second.server.view.html
@@ -1,0 +1,7 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+see my own references: {{seeReferencesUrl}}
+the recommendation was {{recommend}}
+and we can show other reference data, too.
+{% endblock content %}
--- a/modules/core/server/views/email-templates/reference-notification-first.server.view.html
+++ b/modules/core/server/views/email-templates/reference-notification-first.server.view.html
@@ -1,0 +1,5 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+<a href="{{giveReferenceUrl}}">give a reference</a>
+{% endblock content %}
--- a/modules/core/server/views/email-templates/reference-notification-second.server.view.html
+++ b/modules/core/server/views/email-templates/reference-notification-second.server.view.html
@@ -1,0 +1,7 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+<a href="{{seeReferencesUrl}}">see my references</a>
+the recommendation was {{recommend}} <!-- yes, no, unknown -->
+and we can show other reference data, too.
+{% endblock content %}
--- a/modules/references/server/controllers/reference.server.controller.js
+++ b/modules/references/server/controllers/reference.server.controller.js
@@ -10,6 +10,10 @@ var mongoose = require('mongoose'),
     Reference = mongoose.model('Reference'),
     User = mongoose.model('User');
 
+/**
+ * Validate the request body and data consistency
+ * of Create a reference
+ */
 function validateCreate(req) {
   var valid = true;
   var details = [];
@@ -62,7 +66,10 @@ function validateCreate(req) {
   return { valid: valid, details: details };
 }
 
-function create(req, res, next) {
+/**
+ * Create a reference - express middleware
+ */
+exports.create = function (req, res, next) {
 
   var userTo; // not to have to pass found user in callbacks
 
@@ -147,7 +154,7 @@ function create(req, res, next) {
       });
     },
     // ...and if this is a reference reply, make the other reference public, too
-    function (savedReference, otherReference, cb) {
+    function publishOtherReference(savedReference, otherReference, cb) {
       if (otherReference && !otherReference.public) {
         otherReference.set({ public: true });
         return otherReference.save(function (err) {
@@ -158,7 +165,7 @@ function create(req, res, next) {
       return cb(null, savedReference, otherReference);
     },
     // send email notification
-    function (savedReference, otherReference, cb) {
+    function sendEmailNotification(savedReference, otherReference, cb) {
       if (!otherReference) {
         return emailService.sendReferenceNotificationFirst(req.user, userTo, function (err) {
           cb(err, savedReference, otherReference);
@@ -170,13 +177,13 @@ function create(req, res, next) {
       }
     },
     // send push notification
-    function (savedReference, otherReference, cb) {
+    function sendPushNotification(savedReference, otherReference, cb) {
       return pushService.notifyNewReference(req.user, userTo, { isFirst: !otherReference }, function (err) {
         cb(err, savedReference);
       });
     },
     // finally, respond
-    function (savedReference, cb) {
+    function respond(savedReference, cb) {
       return cb({
         status: 201,
         body: {
@@ -193,6 +200,7 @@ function create(req, res, next) {
       });
     }
   ], function (err) {
+    // send error responses
     if (err && err.status && err.body) {
       if (err.body.errType) {
         err.body.message = errorService.getErrorMessageByKey(err.body.errType);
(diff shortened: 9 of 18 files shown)


mrkvon:
I think this piece is close to finished.
Will look through it tomorrow. It's close to be ready for review. It's also quite huge. :unamused:

config/env/default.js
```diff
@@ -18,6 +18,16 @@ module.exports = {
     description: 'Travellers community for sharing, hosting and getting people together. We want a world that encourages trust and adventure.'
   },
 
+  // Feature flags
+  featureFlags: {
+    // enable references at all?
+    reference: false
```
simison: This flag is probably enough for now, detailed ones below we can start implementing/using once we publish the first version of this.

config/lib/worker.js
```diff
@@ -77,6 +77,12 @@ exports.start = function (options, callback) {
       require(path.resolve('./modules/users/server/jobs/user-welcome-sequence-third.server.job'))
     );
 
+    agenda.define(
+      'publish old unpublished references',
+      { lockLifetime: 10000, concurrency: 1 },
```
mrkvon: What does this mean? What values should be here?

config/lib/worker.js
```diff
@@ -86,6 +92,7 @@ exports.start = function (options, callback) {
     agenda.every('15 minutes', 'welcome sequence first');
     agenda.every('60 minutes', 'welcome sequence second');
     agenda.every('60 minutes', 'welcome sequence third');
+    agenda.every('23 minutes', 'publish old unpublished references');
```
mrkvon: 23 minutes may be a different value.

modules/core/server/services/email.server.service.js
```diff
@@ -395,6 +395,35 @@ exports.sendWelcomeSequenceThird = function (user, callback) {
   exports.renderEmailAndSend('welcome-sequence-third', params, callback);
 };
 
+/**
+ * Reference Notification (First between users)
+ */
+exports.sendReferenceNotificationFirst = function (userFrom, userTo, callback) {
+
+  var params = exports.addEmailBaseTemplateParams({
+    subject: 'New reference from ' + userFrom.username,
```
mrkvon: The text may want to be improved.

modules/core/server/services/email.server.service.js
```diff
@@ -395,6 +395,35 @@ exports.sendWelcomeSequenceThird = function (user, callback) {
...
+  exports.renderEmailAndSend('reference-notification-first', params, callback);
+};
+
+/**
+ * Reference Notification (Second reference between users)
+ */
+exports.sendReferenceNotificationSecond = function (userFrom, userTo, reference, callback) {
+
+  var params = exports.addEmailBaseTemplateParams({
+    subject: 'New reference from ' + userFrom.username,
```
mrkvon: The text may want to be improved.

modules/core/server/services/push.server.service.js
```diff
@@ -85,6 +85,35 @@ exports.notifyMessagesUnread = function (userFrom, userTo, data, callback) {
...
+  var readReferencesUrl = url + '/profile/' + userTo.username + '/references';
+
+  // When the reference is first, reply reference can be given.
+  // Otherwise both references are public now and can be seen.
+  var actionText = (data.isFirst) ? 'Give a reference back.' : 'You can see it.';
+  var actionUrl = (data.isFirst) ? giveReferenceUrl : readReferencesUrl;
+
+  var notification = {
+    title: 'Trustroots',
+    body: userFrom.username + ' gave you a new reference. ' + actionText,
```
mrkvon: The text here including `actionText` defined above may want to change.

modules/core/server/services/push.server.service.js
```diff
@@ -85,6 +85,35 @@ exports.notifyMessagesUnread = function (userFrom, userTo, data, callback) {
...
+    click_action: analyticsHandler.appendUTMParams(actionUrl, {
+      source: 'push-notification',
+      medium: 'fcm',
```
mrkvon: What values should be here?

modules/core/server/views/email-templates-text/reference-notification-first.server.view.html
```diff
@@ -0,0 +1,6 @@
+{% extends './email.server.view.html' %}
```
mrkvon: This email template needs a design.

modules/core/server/views/email-templates-text/reference-notification-second.server.view.html
```diff
@@ -0,0 +1,7 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+see my own references: {{seeReferencesUrl}}
```
mrkvon: This email template needs a design.

modules/core/server/views/email-templates/reference-notification-first.server.view.html
```diff
@@ -0,0 +1,5 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+<a href="{{giveReferenceUrl}}">give a reference</a>
```
mrkvon: This email template needs a design.

modules/core/server/views/email-templates/reference-notification-second.server.view.html
```diff
@@ -0,0 +1,7 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+<a href="{{seeReferencesUrl}}">see my references</a>
+the recommendation was {{recommend}} <!-- yes, no, unknown -->
```
mrkvon: This email template needs a design.

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -0,0 +1,215 @@
...
+    // Check if the receiver of the reference exists and is public
+    function isUserToPublic(cb) {
+      User.findOne({ _id: req.body.userTo }).exec(function (err, foundUser) {
+        if (err) return cb(err);
+
+        userTo = foundUser;
+
+        // Can't create a reference to a nonexistent user
+        // Can't create a reference to a nonpublic user
+        if (!userTo || !userTo.public) {
```
mrkvon: this is equivalent `!(userTo && userTo.public)`

modules/core/server/views/email-templates/reference-notification-second.server.view.html
simison: I'd say it needs a copy; email templates already have a design by default:

image

You can see how the button is used from other email templates.

mrkvon: Thank you. :+1: I don't know what those terms mean, therefore the confusion.

simison: Call for copy in https://github.com/Trustroots/trustroots/issues/718#issuecomment-431699368

/cc @Regenwasser @aoifehegarty

modules/references/server/policies/references.server.policy.js
```diff
@@ -0,0 +1,37 @@
...
+  acl.allow([{
+    roles: ['user', 'admin'],
+    allows: [{
+      resources: '/api/references',
+      permissions: ['post']
+    }]
+  }]);
+};
+
+exports.isAllowed = function (req, res, next) {
```
mrkvon: very similar code is repeated over and over again

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -0,0 +1,215 @@
...
+    emailService = require(path.resolve('./modules/core/server/services/email.server.service')),
+    pushService = require(path.resolve('./modules/core/server/services/push.server.service')),
+    Reference = mongoose.model('Reference'),
+    User = mongoose.model('User');
+
+/**
+ * Validate the request body and data consistency
+ * of Create a reference
+ */
+function validateCreate(req) {
```
mrkvon: the request validation could use some abstraction or some library.
_disclaimer:_ i'm not fully aware what _abstraction_ means. i just have _ideas_ about it. :)

modules/references/tests/server/reference.server.jobs.tests.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  var referenceData = {
+    get userFrom() { return generateRandomId(); },
+    get userTo() { return generateRandomId(); },
+    // here we don't care if all interactions are false
+    get met() { return generateRandomBoolean(); },
+    get hostedMe() { return generateRandomBoolean(); },
+    get hostedThem() { return generateRandomBoolean(); },
+    get recommend() { return generateRandomRecommend(); }
```
nicksellen: This caught my eye as it looked _fancy_ to use object getters.

I can imagine this causing some confusion as repeated calls (e.g. `referenceData.userFrom`) will keep returning different values.

I would find something like this less confusing:
```js
function generateReferenceData () {
  return {
    userFrom: generateRandomId(),
    // ... etc
  }
}
```

modules/references/server/policies/references.server.policy.js
nicksellen: I suspect this could be massively simplified, but perhaps not the right time in this PR.

modules/references/server/models/reference.server.model.js
```diff
@@ -0,0 +1,79 @@
...
+  hostedMe: {
+    type: Boolean,
+    default: false,
+    required: true
+  },
+  hostedThem: {
+    type: Boolean,
+    default: false,
+    required: true
+  },
```
nicksellen: I wonder if this is worth putting into a sub schema of `interaction`, so:

```js
{
  interaction: {
    met: true,
    hostedMe: false,
    hostedThem: false
  }
}
```

Advantages:
* validation error for "no interaction selected" can have a field name
* it more clearly shows the intention (and where to add new ones), given mongodb is a document store it's good at nested data

... or an array like is used elsewhere for other things?

modules/references/tests/server/reference-create.server.routes.tests.js
```diff
@@ -0,0 +1,871 @@
...
+  });
+
+  afterEach(function (done) {
+    Reference.deleteMany().exec(function () {
+      User.deleteMany().exec(done);
+    });
+  });
+
+  /**
+   * @TODO this can be refactored. Sign in may be moved to some test utils
```
nicksellen: Maybe can do it? ;)

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -0,0 +1,215 @@
...
+
+        // manage errors
+        if (err) {
+
+          // conflict
+          var isConflict = err.errors && err.errors.userFrom && err.errors.userTo &&
+            err.errors.userFrom.kind === 'unique' && err.errors.userTo.kind === 'unique';
+          if (isConflict) {
+            return cb({
+              status: 409,
```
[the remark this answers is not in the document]
mrkvon: I'm in the 409 camp, although I was in agreement that 400 would be consistent with other code; then I found [409 in duplicate contacts](https://github.com/Trustroots/trustroots/blob/e5595852fdc550eab5c3390057d24c99e39650c6/modules/contacts/tests/server/contact.server.routes.tests.js#L436) and even [200 on signup validation/duplicate username errors](https://github.com/Trustroots/trustroots/blob/master/modules/users/server/controllers/users.authentication.server.controller.js#L208). :confused:

I will implement the duplicate check separately as @nicksellen suggest.

modules/references/tests/server/reference-create.server.routes.tests.js
mrkvon: It is done in a later PR. I refactored this when needed to reuse.

https://github.com/Trustroots/trustroots/pull/759/files#diff-4166014cc218ad965962004b2f8b9f31R136

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -0,0 +1,215 @@
...
+  if (req.user._id.toString() === req.body.userTo) {
+    valid = false;
+    details.push('Reference to self.');
+  }
+
+  // Some interaction must have happened
+  var isInteraction = req.body.met || req.body.hostedMe || req.body.hostedThem;
+  if (!isInteraction) {
+    valid = false;
+    details.push('No interaction.');
```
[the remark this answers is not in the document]
mrkvon: Error details improved as suggested.

modules/references/server/policies/references.server.policy.js
mrkvon: Opened a new issue for this #802

modules/references/tests/server/reference.server.jobs.tests.js
simison: Agreed here; these were simplified in https://github.com/Trustroots/trustroots/pull/723/commits/3c523d1a42fabde58aa07fdf39674ed29f66ebf8

modules/references/tests/server/reference.server.jobs.tests.js
```diff
@@ -44,21 +44,23 @@ describe('Job: Set reference to public after a given period of time', function (
...
-  };
+  function generateReferenceData() {
+    return {
+      userFrom: generateRandomId(),
+      userTo: generateRandomId(),
+      // here we don't care if all interactions are false
+      met: generateRandomBoolean(),
+      hostedMe: generateRandomBoolean(),
+      hostedThem: generateRandomBoolean(),
+      recommend: generateRandomRecommend()
```
simison: If these are random, they could probably simply be `false` instead since we're not testing for their values anyway.

modules/references/tests/server/reference.server.jobs.tests.js
```diff
@@ -44,21 +44,23 @@ describe('Job: Set reference to public after a given period of time', function (
...
-    // here we don't care if all interactions are false
-    get met() { return generateRandomBoolean(); },
-    get hostedMe() { return generateRandomBoolean(); },
-    get hostedThem() { return generateRandomBoolean(); },
-    get recommend() { return generateRandomRecommend(); }
-  };
+  function generateReferenceData() {
+    return {
+      userFrom: generateRandomId(),
+      userTo: generateRandomId(),
```
simison: Something like this should work:
```js
userFrom: mongoose.Types.ObjectId(),
```

modules/references/tests/server/reference.server.model.tests.js
```diff
@@ -0,0 +1,181 @@
...
+          should.not.exist(err);
+        } catch (e) {
+          return done(e);
+        }
+        reference2.save(function (err) {
+          try {
+            should.not.exist(err);
+          } catch (e) {
+            return done(e);
+          }
```
simison: I don't think we have try/catch in our other assertion tests. Not opposed to this but I'm curious to hear rationale here? Somehow I remember earlier discussions about this but it escapes me now...

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -0,0 +1,221 @@
...
+
+  // Some interaction must have happened
+  var isInteraction = req.body.interactions && (req.body.interactions.met || req.body.interactions.hostedMe || req.body.interactions.hostedThem);
+  if (!isInteraction) {
+    valid = false;
+    interactionErrors.any = 'missing';
+  }
+
+  // Value of 'recommend' must be valid ('yes', 'no', 'unknown')
+  if (req.body.recommend && !['yes', 'no', 'unknown'].includes(req.body.recommend)) {
```
simison: Hehe, `includes` is pretty new JS but since this is backend and we're at the brink of moving to ES6, let's roll with it. :thumbsup:

modules/references/tests/server/reference.server.model.tests.js
mrkvon: If the assertion fails, it will lead to test timeout (uncaught error in asynchronous code). Indeed it's missing elsewhere, but I need this when practicing test first development. Can't always wait for the timeout. :)

This is a theory, I'll check it.

modules/references/server/controllers/reference.server.controller.js
mrkvon: :smile: Indeed. This drops support for [node 5 and less](https://node.green/#ES2016-features-Array-prototype-includes).
Hopefully this will be replaced with some validation library in the future. Reinventing the wheel here. :)

simison: Yap, let's add validation library to the todo and loop back to it later. Would you mind opening a quick issue about it and add it to backlog at project board I just made: https://github.com/Trustroots/trustroots/projects/4 ?

mrkvon: #807

modules/references/tests/server/reference.server.model.tests.js
mrkvon: Ok, it's actually caught and no timeout happens. Magic. I can simplify them, then.

:+1:

simison:
Nice! I was trying to understand this better from Mocha docs but didn't see
need for this there neither. Thanks for checking.

modules/references/tests/server/reference.server.model.tests.js
mrkvon: > Nice! I was trying to understand this better from Mocha docs but didn't see need for this there neither.

The only difference is that the report says: `Uncaught Error` instead of `Error`.

With Promises/(async functions), this stuff will be a breeze. :slightly_smiling_face:

modules/references/tests/server/reference.server.jobs.tests.js
mrkvon: resolved with ba25d1c

modules/references/tests/server/reference.server.jobs.tests.js
mrkvon: resolved with ba25d1c

modules/references/server/controllers/reference.server.controller.js
mrkvon: #807

modules/references/tests/server/reference.server.model.tests.js
mrkvon: I'm confused, because now I encountered a case which timeouts without try catch.

Remove `try catch` [here](https://github.com/Trustroots/trustroots/pull/723/files#diff-6af870a19c75bac875ded8cb33478262R208) and change `length(1)` to `length(0)`.

I'm tempted to leave this as it is, because it doesn't cause any harm, keeps things predictable and is likely to be refactored as part of #804.

simison: Sounds good! :thumbsup:

simison: Nah, Github email reply thew my reply out of the thread: https://github.com/Trustroots/trustroots/pull/723#issuecomment-435608095

mrkvon: I'm marking this as resolved, then. :slightly_smiling_face:

Issue #722, opened 2018-10-17
Update the GIBS call - NASA fallback not working


petr-hajek:
When mapbox is off, the nasa gibs fallback is not working.
Reason: GIBS have changed the API, hence it needs to be updated as the fallback option.
https://wiki.earthdata.nasa.gov/display/GIBS/GIBS+API+for+Developers#GIBSAPIforDevelopers-ServiceEndpointsandGetCapabilities
https://github.com/nasa-gibs/gibs-web-examples/blob/master/examples/leaflet/webmercator-epsg3857.js

PR #724, opened 2018-10-18
#722 Updated the NASA GIBS fallback call 

petr-hajek:
also changed the layer to avoid displaying clouds

Fixes https://github.com/Trustroots/trustroots/issues/722


simison:
Thanks! :thumbsup:

Issue #462, opened 2016-11-27
Better DB seeding


simison:
(_Now, this isn't an urgent task by any means but it's a really good one for getting to know the codebase and DB structures, and some nodejs._) ;-)

Now we have quite outdated [DB seeding script](https://github.com/Trustroots/trustroots/blob/master/scripts/fillTestData.js) which seeds only users + hosting offers.

It should add a bunch of messages and tribes, too. Actually it probably makes sense to just start writing that script from scratch.

Mean.js (our boilerplate upstream) has some functionality (and even tests!) written for this, but I'm not sure if their strategy is the way to go: https://github.com/meanjs/mean/search?utf8=%E2%9C%93&q=seed — it's pretty close to core and it might make sense to have this a bit more separate.

@kalpaitch wrote:
>I actually tend to take it that we should be able to take the master db and sanitise it periodically. Can be helpful identifying issues with real uses cases.

For this we have a python script elsewhere to be able to analyse data quickly (written by @guaka). I'll give you access to that repo.

That said, for learning purposes it probably makes sense to write this in JS. These scripts might have useful patterns to look into:

https://github.com/Trustroots/trustroots/tree/master/migrations
https://github.com/Trustroots/trustroots/tree/master/scripts/db-maintenance
https://github.com/Trustroots/trustroots/tree/master/scripts/influxdb

[yargs](https://github.com/yargs/yargs) package is already available in `devDependencies` and would probably come handy.

Related: https://github.com/Trustroots/trustroots/issues/414

simison:
@mrkvon did you have any useful code for populating the messages DB?

mrkvon:
[tr-influx-fake-messages](/mrkvon/tr-influx-fake-messages) is a very naive script for populating the database with messages which i've used for testing message statistics. It doesn't create `Threads` and it drops the database at the beginning.
I believe it is not very useful for the general case unless it is edited substantially. Have a look if you wish. :)

simison:
https://github.com/meanjs/mean/pull/1808

Jagjeet:
I was looking at this as part of Hacktoberfest. Are there any updated thoughts on how to do this? I noted the original request is almost two years old so just figured I would check in.

simison:
[...]

PR #740, opened 2018-10-20
Better db seeding 462 part2

Jagjeet:
This PR does the following relating to issue #462 :

-Restructures the seeding scripts to separate CLI commands from the database seeding
-Adds a script for creating message threads with the following usage: node fillTestMessageData <number of threads to add> <max messages per thread>
-Moves the previous 'User' functionality into a script called 'fillTestUserData.js'
-Combines the logic for creating users and admins into a single function
-Moves the individual seeding scripts into a 'db-seeding' directory
-Creates a master 'fillUserData.js' script that is preconfigured with values that should hopefully be good enough for most cases.

Further issues will be addressed more individually to avoid giant PRs. These will potentially include:
-Adding mock messaging between admin users
-Adding mock images to Tribes
-Improving randomization of message times and read/unread flags
-Soliciting feedback to see what else is valuable to make the mock data more useful. (eg. We could separate out user creation and offer creation, but if nobody needs it there is no point)

In any case, hopefully it will be easier for me or others to make smaller changes individually now that the code is slightly reorged. Let me know if there are any questions.


scripts/db-seeding/fillTestMessagesData.js
```diff
@@ -0,0 +1,151 @@
...
+
+      }).catch(function (err) {
+        console.log(err);
+      });
+    });
+  });
+};
+
+// Number of threads and max number of messages is required
+if (process.argv[2] == null || process.argv[2] < 1 || process.argv[3]== null || process.argv[3] < 1) {
```
simison: As a future improvement, we could build this stuff using [Yargs](http://yargs.js.org/), but that's for another PR.

simison:
What do you think if we'd move these scripts to `/bin` folder and access them via `npm` scripts configured in [`package.json`](https://github.com/Trustroots/trustroots/blob/9574266d8d1a04bde96f1093b12e208af2205126/package.json#L23-L55)?

For example:
```
npm run test-data
```
```
npm run test-data:tribes 20
```
would run:
```json
{
  "test-data": "npm run test-data:*",
  "test-data:tribes": "node ./bin/fillTestData/tribes.js",
  "test-data:messages": "node ./bin/fillTestData/messages.js",
}
```

...or something like that.

Jagjeet:
That sounds good! It's a quick change and I am happy to do it as part of this request. Coming from outside the JS world, I tend to forget to centralize things through package.json, but it is definitely better that way. :)

Jagjeet:
This is all ready to go now. Going to address adding images to the tribes as another PR.

simison:
Works like a charm 👌 

As a future iteration, we could allow setting Mongoose's debug -mode off during the script's run because now the console output is quite jarring.

https://github.com/Trustroots/trustroots/blob/bf775bbb08dd65f3cea2ea1b924868e6c4333d1e/config/env/development.js#L23-L24

Environment variables could work for that.

Issue #746, opened 2018-10-20
About and Overview are both highlighted when visiting a profile in mobile view


Regenwasser:
#### Issue description

Both "About" and "Overview" are highlighted, so it looks like one item (which they maybe could be anyway).

#### Steps to reproduce the issue

1.  Use map to find user
2.  Open a user profile from map

#### What's the expected result?

Just one tab should be highlighted. 

#### Screenshot:

![image](https://user-images.githubusercontent.com/1645834/47261011-7537bb80-d4c6-11e8-8758-122c3b9d72a4.png)

PR #756, opened 2018-10-21
#746 - Give time for highlight to change

LukaszKokot:
This simple fix gives time for the highlight to switch from "about" to "overview".


guaka:
@simison I guess we can merge this?

Issue #261, opened 2015-04-29
Add notes about discussion to contributing.md


chmac:
GitHub issues are not a discussion forum. This is a forum for technical discussion about how to implement features, questions of a technical nature, reporting bugs, and so on.

It is not a place for policy discussion. Personal opinion about which features are important, which features should or should not be implemented, and so on, does not belong here.

We operate a 3 strike policy. 1st warning, 2nd warning, blocked.

Refer to the [Ubuntu Code of Conduct](http://www.ubuntu.com/about/about-ubuntu/conduct) for guidance on how to contribute. Particularly the following points:

* Be considerate
* Be respectful
* Take responsibility for our words and our actions
* Be collaborative
* Value decisiveness, clarity and consensus
* Ask for help when unsure
* Step down considerately

Our code of conduct might be described as follows.

> Be nice to people. Show respect for those who do the work. When in doubt, defer to those with more experience, who have contributed for longer, and support their decisions.

chmac:
@guaka @simison What's missing from the above? Feel free to edit away. Should we move it to a wiki page so we can see a history of edits?

simison:
I'd turn the whole thing reverse — last (positive, encouraging) bits first and first bits (no/don'ts) to the bottom.

Wikipage would be better. On the other hand Github shows a big link to Contributing-file each time when creating and issue, so def. add a link to contributing file.

chmac:
Awesome, look forward to reading your edits! :-)

Yeah, I thought maybe wiki page to edit it, and then copy it into contributing once it's ready.

simison:
;-) I might. It's very good already anyway. Thanks!

guaka:
No need for a wiki page, I copied and remixed it directly at https://github.com/Trustroots/trustroots/blob/master/CONTRIBUTING.md - please edit there, also works quite well from GitHub's web interface.

simison:
Thanks, LGTM.

guaka:
Just added this: "We highly value real life interaction. Both within the network as well as among technical contributors. If possible, try to meet up with other Trustroots contributors in real life."

We should do something cool with the GitHub connection on TR profiles soon :)

simison:
BTW I quite like "be pragmatic" bit at https://owncloud.org/community/code-of-conduct/ (ping @jancborchardt)

jancborchardt:
Yeah – that’s very important to fend off the »power users« who tend to be very loud on the issue tracker. ;)

[...]

PR #758, opened 2018-10-22
code of conduct

guaka:
straight copy from contributor covenant v1.4

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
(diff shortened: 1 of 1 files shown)


guaka:
possible improvments:
- [ ] We probably want to refer to https://www.trustroots.org/rules
- [ ] We want to contributors of significant contributions to have a profile on Trustroots, and we (as in board/organisation) probably want to know which profile that is

guaka:
Ah we already had https://github.com/Trustroots/trustroots/blob/master/CONTRIBUTING.md - see #261

simison:
> We want to contributors of significant contributions to have a profile on Trustroots, and we (as in board/organisation) probably want to know which profile that is

That's maybe less code-of-conduct and rather just a protocol / "heads up, please do this" -type of thing somewhere.

I'd suggest we work on _volunteer onboarding_ which is:

- Ask them to read and abide by the code of conduct
- a checklist for things like "make sure to create TR profile and add it to your slack profile", "write an introduction at the slack channel"
- Pointers for people, channels, repositories: where to ask what, who does what, etc
- Some good reading
 
The world is probably full of bots for this purpose (like https://greetbot.io/) that can send a private message to folks who join Slack.

Jagjeet:
> "make sure to create TR profile and add it to your slack profile"

Is it possible to add a field for this in the Slack profile? Not sure how much customization is possible.

guaka:
https://github.com/Trustroots/trustroots/community looks good
image

Issue #753, opened 2018-10-21
Update native MongoDB driver


simison:
[`mongodb`](https://www.npmjs.com/package/mongodb) driver needs updating from v2 to v3.

https://github.com/Trustroots/trustroots/blob/9574266d8d1a04bde96f1093b12e208af2205126/package.json#L114

See breaking changes:

https://github.com/mongodb/node-mongodb-native/blob/HEAD/CHANGES_3.0.0.md

It's used by Worker script, here: 
https://github.com/Trustroots/trustroots/blob/15a8bcebee8286eca1591f950a120420b2e08dd3/config/lib/worker.js

Shouldn't affect the main app script (app and worker run separately in parallel with `npm start`), since the app script depends on Mongoose handling the connection directly.

Alternatively we could refactor Worker depend on the Mongoose connection just like the main app script does. That would require changes to Angeda configuration: https://github.com/agenda/agenda#mongomongoclientinstance

steaward:
Hey, just caught wind of this project. I'd love to help contribute. 
I can attempt to tackle this issue for you.

simison:
Yay! That would be amazing, @steaward

steaward:
@simison it seems that I am able to use mongoDB v 3.1.8 (this is the version I see on NPM) with Trustroots. 

I ran the script `scripts/fillTestData.js` as well as `scripts/fillTestTribesData.js` and things seemed to work without problems. 

Something did pop up though:
When I ran `npm test`, 4 tests failed. However, the `master` is also failing these same 4 (they are to due with changing a user's profile picture to jpeg, gif, ect.). Is this just me failing somewhere, or is this a known issue?

Just a little concerned before I make a PR. 
Thanks :)

PR #762, opened 2018-10-23
Issue 753: Updating MongoDB

steaward:
Hey, 

This PR is in respect to [issue 753](https://github.com/Trustroots/trustroots/issues/753)
I've updated the code to fit with MongoDB v3. 
Only two things that I noticed that needed changing was the fact the DB is now passed back to the client during the initial connection. I don't know what Trustroots DB name is, so it's left blank by default. The collection `'agendaJobs'` is then grabbed. 

Ran all tests as well as the scripts to populate the DB with fake data and it's failing on the profile picture tests. However, the `master` branch from my fork fails on this too?


simison:
Thanks!

Looks like `package.json` and `package-lock.json` are still out of sync and thus `npm ci` fails tests in Travis.

Running `npm install` should update the lock file and you can push changes to it here.

simison:
>Only two things that I noticed that needed changing was the fact the DB is now passed back to the client during the initial connection. I don't know what Trustroots DB name is, so it's left blank by default. The collection 'agendaJobs' is then grabbed.

simison:
Thanks for working on this!

I know this might've been a confusingly small task but I like to leave these out there for the community during the _Hactoberfest_.

>Only two things that I noticed that needed changing was the fact the DB is now passed back to the client during the initial connection. 

Yap, only the background job worker script uses native driver directly for Agenda's connection (job runner library) — the server app itself uses Mongoose ODM at all levels.

Agenda supports re-using Mongoose connection so we should harmonise the setup as a future iteration.

> I don't know what Trustroots DB name is, so it's left blank by default. The collection 'agendaJobs' is then grabbed.

Trustroots DB name depends on environment config and is different for prod, dev and test envs. It's the last bit in these URIs:

https://github.com/Trustroots/trustroots/blob/3281c5c0a21cd87812b66040ddca2d1507b4a4ca/config/env/development.js#L15
https://github.com/Trustroots/trustroots/blob/3281c5c0a21cd87812b66040ddca2d1507b4a4ca/config/env/production.js#L15
https://github.com/Trustroots/trustroots/blob/3281c5c0a21cd87812b66040ddca2d1507b4a4ca/config/env/test.js#L15

Worker script gets this via `config.db.uri` so we're all good. 👌

simison:
>Ran all tests as well as the scripts to populate the DB with fake data and it's failing on the profile picture tests.

This could be due missing `imagemagic` in your local setup?

They don't seem to fail on Travis CI currently (this branch [shows up green](https://travis-ci.org/Trustroots/trustroots/builds/445735818?utm_source=github_status&utm_medium=notification) as well)/

PR #774, opened 2018-10-26
Add button to open location in app for mobile devices

### LINKED ISSUE #704: open locations with other apps
When I'm using TR on my phone I'd like to open locations with mapping applications.

See https://en.wikipedia.org/wiki/Geo_URI_scheme

topalavlad:
#### Proposed Changes

Add "Open in device" button that has a Geo URI with the fuzzy location of the host;
Whitelisted "geo:" URIs;
Button is only visible on mobile devices

#### Testing Instructions

Test that the pressing the button opens a maps app or browser on mobile devices;
Test that button is not visible on desktop browsers and is visible in mobile browsers

Fixes #704

### DIFF
--- a/modules/core/client/app/init.js
+++ b/modules/core/client/app/init.js
@@ -53,6 +53,7 @@
     // @link https://docs.angularjs.org/guide/production#disable-comment-and-css-class-directives
     $compileProvider.commentDirectivesEnabled(false);
     $compileProvider.cssClassDirectivesEnabled(false);
+    $compileProvider.aHrefSanitizationWhitelist(/^geo:.*/);
 
     // By default tooltips and popovers are appended to
     // '$body' instead of the parent element
--- a/modules/offers/client/directives/tr-offer-host-view.client.directive.js
+++ b/modules/offers/client/directives/tr-offer-host-view.client.directive.js
@@ -22,7 +22,7 @@
     return directive;
 
     /* @ngInject */
-    function trOfferHostViewDirectiveController($scope, OffersByService) {
+    function trOfferHostViewDirectiveController($scope, $window, OffersByService) {
 
       // ViewModel
       var vm = this;
@@ -35,7 +35,7 @@
       vm.isUserPublic = false;
       vm.hostingDropdown = false;
       vm.hostingStatusLabel = hostingStatusLabel;
-      vm.isMobile = navigator.userAgent.toLowerCase().indexOf('mobile') >= 0;
+      vm.isMobile = $window.navigator.userAgent.toLowerCase().indexOf('mobile') >= 0 || $window.isNativeMobileApp;
       activate();
 
       /**
--- a/modules/offers/client/directives/tr-offer-host-view.client.directive.js
+++ b/modules/offers/client/directives/tr-offer-host-view.client.directive.js
@@ -35,7 +35,7 @@
       vm.isUserPublic = false;
       vm.hostingDropdown = false;
       vm.hostingStatusLabel = hostingStatusLabel;
-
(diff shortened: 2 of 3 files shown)


modules/offers/client/views/directives/tr-offer-host-view.client.view.html
```diff
@@ -198,6 +198,11 @@ <h4 ng-if="!trOfferHost.isOwnOffer">
            class="btn btn-sm btn-inverse-primary">
           Bigger map
         </a>
+        <a ng-if="trOfferHost.isMobile"
```
simison: Let's try one-time-binding (`::`) here so that `isMobile` wouldn't be evaluated more than once (and not on every scope-render): https://code.angularjs.org/1.5.11/docs/guide/expression#one-time-binding

modules/offers/client/views/directives/tr-offer-host-view.client.view.html
```diff
@@ -198,6 +198,11 @@ <h4 ng-if="!trOfferHost.isOwnOffer">
...
+          ng-href="geo:{{trOfferHost.offer.location[0]}},{{trOfferHost.offer.location[1]}};u=200"
+           class="btn btn-sm btn-inverse-primary">
+          Open on device
```
simison: Feels very technical; would there be any alternative ways to say the same?

Not a blocker though, we can very well merge with this. :-)

modules/offers/client/directives/tr-offer-host-view.client.directive.js
```diff
@@ -35,7 +35,7 @@
       vm.isUserPublic = false;
       vm.hostingDropdown = false;
       vm.hostingStatusLabel = hostingStatusLabel;
-
+      vm.isMobile = navigator.userAgent.toLowerCase().indexOf('mobile') >= 0;
```
simison: Let's access `navigator` via `$window.navigator` Angular service and add `$window` to the list on line 25. Not a biggie but it's just more Angular way of accessing globals.

While at it, let's also a check if `$window.isNativeMobileApp` is true. That's set by our native mobile app bridge (http://android.trustroots.org).

modules/offers/client/views/directives/tr-offer-host-view.client.view.html
topalavlad: `See location in app` maybe? Although I think I prefer `Open on device`

modules/offers/client/views/directives/tr-offer-host-view.client.view.html
topalavlad: Done

topalavlad:
Added the changes, not sure about the text on the button

modules/offers/client/views/directives/tr-offer-host-view.client.view.html
simison: We can think about it and update later on. 👍

modules/core/client/app/init.js
```diff
@@ -53,6 +53,7 @@
     // @link https://docs.angularjs.org/guide/production#disable-comment-and-css-class-directives
     $compileProvider.commentDirectivesEnabled(false);
     $compileProvider.cssClassDirectivesEnabled(false);
+    $compileProvider.aHrefSanitizationWhitelist(/^geo:.*/);
```
simison: I believe this has made every other protocol show up as "unsafe". 😁  

I'll revert this change, let's get a fix in.

topalavlad: Oops, I'll check it out this week

simison: Fixed in https://github.com/Trustroots/trustroots/pull/846

PR #776, opened 2018-10-26
Set autocapitalize off for password field

### LINKED ISSUE #775: Password-visible field should have autocapitalize false
**Describe the bug**
When the user wants to see what he types in the password field, the field should not autocapitalize the password since it's easy to miss and error prone. Behavior should be the same regardless if password is hidden or not.

**To Reproduce**
1. Go to Sign in screen
2. Click on the `Show password` button on the right of the password field
3. Start typing 
4. First letter is capitalized

**Expected behavior**
First letter is not capitalized

- Does not apply

**Smartphone (please complete the following information):**
- All smartphones

topalavlad:
#### Proposed Changes

Add `autocapitalize="off"` for password field because it's error prone on mobile devices when user sets password to be visible

#### Testing Instructions

Test that the first typed letter in the password field is not capitalized when the password is visible

Fixes #775

### DIFF
--- a/modules/users/client/views/authentication/signin.client.view.html
+++ b/modules/users/client/views/authentication/signin.client.view.html
@@ -31,6 +31,7 @@
                          placeholder="Password"
                          class="form-control input-lg"
                          ng-attr-type="{{ showPassword ? 'text' : 'password' }}"
+                         autocapitalize="off"
                          ng-disabled="auth.isLoading"
                          ng-model="auth.credentials.password"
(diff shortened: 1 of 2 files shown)


simison:
Nice catch! 
Could you also check if this should be added to fields where user can change their password?

topalavlad:
The change password fields are always of password type. The text cannot be made visible, but I'll update the register field which does have the visibility toggle.

PR #778, opened 2018-10-27
Test framework deps

Jagjeet:
#### Proposed Changes

* Update testing framework to the latest Mocha version

#### Testing Instructions

* Run the test suite and verify there are no errors. 

Fixes #712

### DIFF
--- a/config/lib/mongoose.js
+++ b/config/lib/mongoose.js
@@ -39,10 +39,7 @@ module.exports.connect = function (callback) {
   // https://mongodb.github.io/node-mongodb-native/2.1/api/Server.html
   // http://mongoosejs.com/docs/connections.html
   var mongoConnectionOptions = {
-    server: {
-      // Never stop reconnecting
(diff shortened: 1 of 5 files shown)


Jagjeet:
I was able to finally get the tests running, but Mocha hangs after they are complete. This issue is documented here, but I'm having issues debugging it. 

https://github.com/mochajs/mocha/issues/3044

There is an exit flag that can be used with Mocha to behave like v3, but even with that on, the hanging behavior still occurs. 

Also currently, the database connection is not being terminated. In testing locally, even when I terminate the connection, mocha still hangs. The documentation suggests using the --recursive flag and an AfterAll hook, but I couldn't get it to work with gulp-mocha. 

https://github.com/mochajs/mocha/issues/2507

One option might be to use the Mocha API directly, but I haven't went down that road yet.

simison:
Hmm. I wonder if eliminating all tests and adding them back set by set
would reveal anything?

PR #783, opened 2018-10-29
Remove bower

simison:
Removes Bower package-manager dependency by installing everything via `npm` and requiring scripts directly from `node_modules`.

Todo:

Do something about `/lib/*` assets (mainly [image files](https://github.com/Trustroots/trustroots/blob/d1695a416294854baf7d0dc528bd5131f73e83a5/modules/core/client/less/modules/chosen.less#L13) referenced at core-module and [JS files for Firebase](https://github.com/Trustroots/trustroots/blob/f4bd051562b112e28b4bbeb920374b7a7713dbb2/public/push-messaging-sw.js#L6-L7)) in a few JS files and CSS files.

We could either copy these over to `public` folder with Gulp or use express static-serve handle them.

#### Testing Instructions

- Start the app both in dev (`npm start`) and prod (`NODE_ENV=production npm start`) modes
- Click around and see if you see 404s in console or anything off/missing

Issue #779, opened 2018-10-28
Crop and cancel new upload for profile pictures


Akronix:
**Is your feature request related to a problem? Please describe.**
When I wanted to change my profile picture, I selected a vertical picture that it is too tall for the square ratio for Trustroots profile picture. I could crop it and select the area I might use for that picture, but I coudn't go back to my former one either, because the picture was already set as my new profile picture.

**Describe the solution you'd like**
I would like to be able to select the area in a square or circle shape as it is going to be displayed in Trustroots, similarly as it is in other social networks.
Also, it should be possible to cancel the action and go back to your previous profile picture.

Add links to discussion on Slack (or Discourse)
https://trustroots.slack.com/archives/C0A3Q15SS/p1540642721001100

Issue #638, opened 2018-08-22
Is MEAN still active?


guaka:
I looked around a bit, and MEAN.io looks like an inactive project these days. Is my assessment correct?

If not, what does this mean for Trustroots development?
Also, how come it's not active anymore?

If it is, where is the activity taking place?

simison:
Short answer: _nope_.

Note that http://meanjs.org/ (inactive now) and http://mean.io/ are two separate projects. The former was a fork of mean.io in 2014: http://blog.meanjs.org/post/76726660228/forking-out-of-an-open-source-conflict

Trustroots was built on meanjs.org version.

[Riess.js](https://github.com/lirantal/Riess.js) was an attempt to move it forward but it seems to have stalled, too.

Anyway, they’re written in Angular v1 which is old and outdated and there wasn’t clear update path to v2 and onwards (think Drupal style major breaking updates), so we got stuck.

We’ll eventually need to look into replacing Trustroots frontend bit-by-bit with something else, e.g.  React. It'll be a good change to decouple backend and frontend codebases a bit further, too.

I’d say regarding TR's framework, a lower hanging fruit would be to turn at least backend code from ES5 to ES6, I wrote a bit about it in chat, copypasting some bits:

>I’ve been playing with codemodding in another project and it was fun, I’m planning to do the same at least for the backend codebase using 5to6-codemods: https://github.com/5to6/5to6-codemod

>here’s a nice list of further reading and list of examples:  https://github.com/sejoker/awesome-jscodeshift

>In theory porting the backend is pretty straightforward;  needs separate eslint rules for client/server folders, `eslint --fix`, some codemodding and manual sorting out.
>

guaka:
I'm still curious, what does it mean that meanjs.org is not active anymore? I.e. in terms of upgrades of stuff, and documentation, should we just forget it's "MEAN"?

simison:
>I'm still curious, what does it mean that meanjs.org is not active anymore? I.e. in terms of upgrades of stuff, and documentation, 

We're on our own now. :-) Docs and stuff are still out there so it doesn't hurt to mention the fact, but going forward there aren't any "free" framework level updates trickling down for us.

>should we just forget it's "MEAN"?

Basically yes.

PR #784, opened 2018-10-29
Switch from bower to webpack

nicksellen:
#### Proposed Changes

* switching from bower to webpack - [bower](https://bower.io/) says: _...psst! While Bower is maintained, we recommend using Yarn and Webpack or Parcel for front-end projects read how to migrate!_
* also prep for moving away from angular1 to react (see https://github.com/Trustroots/trustroots/issues/779#issuecomment-433696445 and https://github.com/Trustroots/trustroots/issues/638#issuecomment-415192512)

#### Testing Instructions

* run dev setup as normal
* ... which should run `npm install` and `npm run build` as part of it, if not, run those
* visit the site
* you should not notice anything changed (my favourite kind of contribution)

#### TODO

- [x] put bower in fire by integrating / rebasing with https://github.com/Trustroots/trustroots/pull/783
- [x] make tests work
- [x] build service worker scripts via webpack too
- [x] production build setup
- [x] check/fix profile networks (`Error: dst is undefined`)
- [x] check it works in docker
- [x] merge/adapt/rewrite dev scripts to use webpack build (and remove `build:webpack` npm script)

Some time later, or sooner, depending on motivation, but not essential for now, probably best in another PR:
- [ ] put gulp in fire
- [ ] style (css/less) via webpack
- [ ] add hot (module) reloading
- [ ] maybe use some cooler webpack options (splitting, etc) to make it suitable for production deployment
- [ ] reimplement `uib-templates.js` / `templates.js` building via webpack loader
- [ ] consider extracting and abstracting dependencies for legacy libs ([example from fs.de](https://gist.github.com/nicksellen/569a54762e9c2343cf27ffe15682a1aa))
- [ ] move away from massive lists of files in `config/assets/*`
- [ ] actually use js modules (currently all the actual `modules/*` files are written as if they are global) - can probably use some [webpack contexts](https://webpack.js.org/guides/dependency-management/) for globbing
- [ ] add es6 loader to webpack
- [ ] add at least 300 webpack plugins so computer can be used as heater in the coming winter


simison:
>maybe use some cooler webpack options (splitting, etc) to make it suitable for production deployment

We could leave most optimizations for another PR, especially since we don't declare any sort of dependency tree, so it makes sense to build just one bundle for now.

>style (css/less) via webpack

If it helps, this could easily be in another PR as well.

nicksellen:
> We could leave most optimizations for another PR, especially since we don't declare any sort of dependency tree, so it makes sense to build just one bundle for now.

You get most of it out-the-box with mode=production. The webpack peeps simplified it a lot since the old days. Development build cannot be used on production site, it's crazy and enormous.

simison:
Yap, I meant to leave fancier optimizations like code-splitting and tree-shaking for later tinkering. :thumbsup:

nicksellen:
This is rebased on top of #783 now :) ... and the errors from there fixed too.

config/webpack/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+const shims = require('./webpack.shims');
+const basedir = join(__dirname, '../..');
+
+const config = require('../config');
+
+const prod = process.env.NODE_ENV === 'production';
+
+module.exports = merge(shims, {
+  mode: prod ? 'production' : 'development',
+  entry: require.resolve('./entries/main'),
```
simison: Should we keep entry files as part of the app modules structure?

Something like `/modules/core/client/app/index.js` / `boot.js` / `main.js`

config/webpack/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+
+const shims = require('./webpack.shims');
+const basedir = join(__dirname, '../..');
+
+const config = require('../config');
+
+const prod = process.env.NODE_ENV === 'production';
+
+module.exports = merge(shims, {
+  mode: prod ? 'production' : 'development',
```
simison: nit: `prod` could be `isProduction` for clarity?

config/webpack/webpack.config.js
nicksellen: That would then output one file per client module. I think it's better to leave webpack to split the files. What is common/easy with vue-router (and others I'm sure) is to use webpack dynamic loading so the resources for each page are loaded on demand without ever having to explicitly think about it.

The current approach here is basically a hack to leave the `assets/*` files as they are.

config/webpack/webpack.config.js
nicksellen: Sure! Have a free :horse: too while I'm here :)

config/webpack/webpack.config.js
simison: Gotcha, yeah Angular1 + ng-ui-router + Webpack should be popular enough combo that I'm sure someone has figured it out.

> The current approach here is basically a hack to leave the assets/* files as they are.

Makes sense! :thumbsup:

nicksellen:
I'm not too sure about the production build right now. `npm run start:prod` will bring up the homepage ok, but it seems to point to a different db and I didn't go further for now.

There is quite a lot of complexity and cruft in the scripts but I'd rather not address that in this PR. But there might be some improvements worth making.

Total bundled size is up a bit, from about 1.3mb (minified, no compression) to about 1.5mb. Seems not too bad.

nicksellen:
So, would be great to get it reviewed/tried out, I suspect there are a few things that can be made a bit smoother too, or clarified, etc.

nicksellen:
The profile network connections are broken, but don't know if that's just because it's local dev. Better go and prepare some food now :)

simison:
This is really neat work! :tada:

I've tested both development and production builds.

> I'm not too sure about the production build right now. npm run start:prod will bring up the homepage ok, but it seems to point to a different db and I didn't go further for now.

Yep, dev and prod use different collections via config. The actual deployment is built using `npm run build:prod`.

Production seems to build `push-messaging-sw.js` and other files nicely:

image


>There is quite a lot of complexity and cruft in the scripts but I'd rather not address that in this PR. >But there might be some improvements worth making.

Separate PRs better indeed. 👌 


>Total bundled size is up a bit, from about 1.3mb (minified, no compression) to about 1.5mb. Seems not too bad.

Yep, totally fine. 👍 New 0.44 mb and old 0.39 mb gzipped; at least we can optimize this much easier now! 💪 

>The profile network connections are broken, but don't know if that's just because it's local dev

Yeah they're not meant to work in `localhost` just like that. I connected via GitHub oAuth with some `/etc/hosts` juggling and it worked fine.


Great work!

I found one small issue with assets we could address before merging:

On profile edit page (http://localhost:3000/profile/edit), when modifying languages, the select directive is missing some images:

image

Here:
https://github.com/Trustroots/trustroots/blob/15a8bcebee8286eca1591f950a120420b2e08dd3/modules/core/client/less/modules/chosen.less#L6-L25

We could either use `file-loader` to load them from `node_modules` or alternatively just move them to [`/modules/core/client/img`](https://github.com/Trustroots/trustroots/tree/15a8bcebee8286eca1591f950a120420b2e08dd3/modules/core/client/img) since we've copied over the CSS from the module as well.


So looking forward to merge this!

simison:
> check it works in docker

FYI, I didn't test Docker setup yet.

Docker might need adding `bower` back?

nicksellen:
Ok, fixed up images (put them in `img/` folder as suggested, can't really use `file-loader` properly until we have the style going through webpack), and a couple of other minor things.

Didn't try docker, but should probably wait until someone has.

simison:
Greets from the Docker land:

```
> message-center@1.2.7 postinstall /trustroots/node_modules/message-center
> bower install

bower ESUDO         Cannot be run with sudo

Additional error details:
Since bower is a user command, there is no need to execute it with superuser permissions.
If you're having permission errors when using bower without sudo, please spend a few minutes learning more about how your system should work and make any necessary repairs.

http://www.joyent.com/blog/installing-node-and-npm
https://gist.github.com/isaacs/579814

You can however run a command with sudo using "--allow-root" option
npm WARN karma-webpack@3.0.5 requires a peer of webpack@^2.0.0 || ^3.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN ajv-keywords@3.2.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN The package firebase is included as both a dev and production dependency.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! message-center@1.2.7 postinstall: `bower install`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the message-center@1.2.7 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2018-10-29T22_33_26_841Z-debug.log
ERROR: Service 'trustroots' failed to build: The command '/bin/sh -c npm install --quiet' returned a non-zero code: 1
```

What a bummer.

I just went ahead and forget that library and removed the `postinstall` script: https://github.com/Trustroots/message-center

Let's use that one instead.

nicksellen:
Oh the other thing to consider is the dev experience. At the moment I guess people change a js file and wait for livereload to reload browser or do it manually. 

I didn't want to put too much time into it here as would prefer to do it after gulp is gone with a proper hot reload setup. But it might be a worse experience until then. Maybe a minimal step is to at least run webpack in watch mode during dev.

nicksellen:
Ok, fixed a couple of things:
* switched to fork of message-centre, removed bower entirely
* fixed up webpack watch mode in dev (will rebuild and trigger browser reload)
* checked `npm run test:client:watch` works fine, it does

So, should be good to go from my perspective. Didn't run the docker setup.

simison:
Nice, I'll give this another spin later tonight. :thumbsup:

nicksellen:
Cool, just rebased on top of master again.

.gitignore
```diff
@@ -35,8 +34,9 @@ scipts/docker/nginx-conf/*
 
 # App
 build
-public/lib
```
simison: Should we add this line back to ensure folks have a smooth transition?

This folder will exist with 1500 files and they'll appear in their `git status` the moment this goes to `master`.

simison:
Docker spins up now nicely! :thumbsup:

Live-reloading works smoothly as well.

I'll put back that previously ignored `public/lib` for a while so that other devs have it easier and merge.

🍻

simison:
Moved todo to https://github.com/Trustroots/trustroots/issues/791

.gitignore
```diff
@@ -37,6 +36,8 @@ scipts/docker/nginx-conf/*
 build
 public/lib
```
mrkvon: `public/lib` can be removed

simison: Yap, I put it back to be there for a little while so that simply pulling latest master wouldn't cause 1500 files to appear in `git status` when  `public/lib` disappears from gitignore.

Let's loop back to it and remove in a couple weeks.

Meanwhile, run `rm -fr public/lib` :-)

PR #787, opened 2018-10-29
Test data: optimize mock content for less emails

simison:
I have a bunch of generated mock data and the app decided to send tons of "you've got message!" notifications. 😆 

screen shot 2018-10-29 at 22 27 23

### Changes
- `read:false` so that the app won't send "new unread message" emails
- `welcomeSequenceStep:3` so that welcome sequence emails are never sent

### Testing Instructions

* Run `npm run test-data` to generate some content
* Wait for some 15 mins and run the app, no "new email' notifications flying out

### DIFF
--- a/bin/fillTestData/Users.js
+++ b/bin/fillTestData/Users.js
@@ -116,6 +116,7 @@ var addUsers = function (max, adminUsers) {
           user.public = true;
           user.avatarUploaded = false;
           user.avatarSource = 'none';
+          user.welcomeSequenceStep = 3;
 
           if (admin !== undefined) {
             // admin user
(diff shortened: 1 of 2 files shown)


bin/fillTestData/Users.js
```diff
@@ -116,6 +116,7 @@ var addUsers = function (max, adminUsers) {
           user.public = true;
           user.avatarUploaded = false;
           user.avatarSource = 'none';
+          user.welcomeSequenceStep = 3;
```
Jagjeet: I am assuming that is the last step in welcome emails, but I couldn't find elsewhere in the code this value getting set to 3. On that note when we switch to ES6, it would be helpful to use consts so we can enumerate what each step is.

simison: Yep it's the last one; It's set in this job: https://github.com/Trustroots/trustroots/blob/15a8bcebee8286eca1591f950a120420b2e08dd3/modules/users/server/jobs/user-welcome-sequence-third.server.job.js

Using `$inc` https://docs.mongodb.com/manual/reference/operator/update/inc/

Thanks for looking into it!

PR #790, opened 2018-10-30
Add React and an example component

nicksellen:
#### Proposed Changes

* add react just to see how it works/feels

#### Testing Instructions

* run up as normal
* look on profile page
* the "Member since..." and "Online..." boxes on the left are rendered with a react component!

It's just a proof of concept to see how it integrates. That particular component has a load of other stuff too, and maybe not worth switching it over, but yeah, just gives an example. Seems quite nice!


mrkvon:
Exciting! Is this something to consider or use in the new client PRs already? Is the idea of using React (instead|on top) of AngularJS mature?

simison:
>Is the idea of using React (instead\|on top) of AngularJS mature?

Kinda as this PR well proves; that said, there are still unknowns regarding routing and data fetching — especially the latter one will be important to figure out sooner than later but we might be able to work our way around it for a while by using Angular app as an app-shell and limiting react only to components inside the app, basically having tons of render-trees all over the place.

> Is this something to consider or use in the new client PRs already?

Definitely. It would be good to get some "framework" for doing this in place first with tiny changes like this.

References list could be an interesting experiment to try with React indeed, but only if it doesn't significantly slow us down shipping references.

nicksellen:
I added a import helper so all react components are automatically imported, from the comment in main.js:

```
/*
 *  Imports all react components from modules/ and register them as angular components
 *
 *  So if you defined modules/users/client/components/TrustrootsGreeting.component.js as:
 *
 *    export default function TrustrootsGreeting({ name }) {
 *      return <p>Hello {name}, from Trustroots!</p>;
 *    };
 *
 *  It will be available to use in (any) angular  as:
 *
 *    <trustroots-greeting name="'Donald'"></trustroots-greeting>
 *
 *  Uses a webpack require context
 *  See https://webpack.js.org/guides/dependency-management/#require-context
 */
```

simison:
>I added a import helper so all react components are automatically imported, from the comment in main.js:

Good idea, although the more explicit imports we can have (even at module level) and less magic around, the clearer dependencies are going to be and easier everything is to grok.

This time convention over configuration (that the `modules/users/client/config/users.client.config.js` file is) totally makes sense though.

How do you see the path forward later on when we might have components that are very module (or "section") specific, as opposed to being more generic and re-usable?

At what point we could have completely explicit imports and remove `importComponents`?

nicksellen:
Tests weren't working before when using react2angular, but ngreact works fine too (and has more github stars) and was easy to swap in.

I would suggest this is _ready_, except the example component I created is not complete, or very useful. but also probably not a good idea to merge it without _any_ react components in use, so perhaps either:

* @mrkvon or someone else wants to start building a real react component and can build on top of this branch (I would favour this option)
* someone picks a really simple component to implement and we can merge with that one

Both cases would involve removing the one I created for this proof of concept.

mrkvon:
@nicksellen How shall we talk with api? Within angular, or react? Do you have any guidelines to share?

I'm very new to react, done the [tutorial](https://reactjs.org/tutorial/tutorial.html) and bits of docs only. Also read a [nice article about data fetching](https://www.robinwieruch.de/react-fetching-data/), which seems to assume full react application.

What is the long-term future picture of frontend? Everything i.e. react + redux? @nicksellen @simison

nicksellen:
Personally, I would keep it simple/incremental. Try with just fetch (+ [polyfill](https://www.npmjs.com/package/whatwg-fetch)), and if that doesn't have enough features then [axios](https://www.npmjs.com/package/axios). Basically like that second link.

Although I would suggest to add one additional layer - an API service that abstracts any of the details about using fetch/axios. So inside a component you can use it something like:

```js
// login
await api.auth.login({ username, password })
// fetch contacts
const contacts = await api.contacts.list()
```

That kind of thing...

At some point it might make sense to use redux, but probably not worth it until it seems useful. The API service will still be used then, as will the components.

If the presentation is getting too mixed up with logic you can always write two components:
- one stateful component that does API calls, or whatever, passes down props and callback handler functions (e.g. `onAddContact`)
- one that purely does presentation, renders only based on props, and calls handler functions to _do stuff_

mrkvon:
I'd like to create a link (like `ui-sref`) from react component; maybe a redirect. I've been struggling with this most of the day. Most recently trying with [@uirouter/react-hybrid](https://github.com/ui-router/react-hybrid), facing errors.

@nicksellen @simison Do you have any experience with wiring up routing between react and angular-ui-router? Or ideas?

I'm thinking about using `href` and forcing a page reload... :unamused:

simison:
>I'm thinking about using href and forcing a page reload... 😒

Yeah, depending on where these are it should be safe and best we can have when living with two frameworks.
E.g. moving between tabs on profile page using `href` can be problematic, but moving from "module to module" it's quite safe.

mrkvon:
rebased

nicksellen:
@mrkvon you can access all the angular services from outside angular, e.g.:

```js
var $state = angular.element(document).injector().get('$state');
// Go to messages page ("inbox" state)
$state.go('inbox')
// Get URL for messages page
$state.href('inbox')
// ... or one with params
$state.href('profile', { username: 'peter' })
```

And there is an example where I use some angular services outside of angular in `photos.services.js` (to call `$emit`).

If you do anything more fancy outside of angular you might need to consider the digest scope to ensure things update in the angular world, but I think it should be ok for these uses.

mrkvon:
Rebased, moved the react profile-view-basics component and fixed a link.

I'll merge this now. It is the first merge I do into Trustroots master. Feel free to revert if there are any issues.
@simison @guaka

mrkvon:
`npm run build:prod` is failing.

mrkvon:
@nicksellen 
> It's quite easy to start dropping browser support with some of these things if we don't get the babel options right
Is this something you took care of, or is it yet to be done?

PR #806, opened 2018-11-03
Revert "Statistics: generate "last seen" snapshot statistics about users"

trmendes:
Reverts Trustroots/trustroots#616


simison:
Looking

simison:
We're good without a merge.

#616 was accidental merge but tests are passing and the code looks good. :-)

PR #816, opened 2018-11-07
Replying "no" to push notification question does not persist between sessions

### LINKED ISSUE #585: Replying "no" to push notification question does not persist between browser resets
Either my "No, Thanks" is stored in my cookie which is deleted with closing the browser or it is not stored at all. It is annoying. Please make it stop.

topalavlad:
#### Proposed Changes

Save `tr.push.asked` key on a different namespace so that it doesn't get cleared on sign out

#### Testing Instructions

The testing should be done on a desktop browser if tested before #811

1. Login, answer question `Can we send you push notifications?` with either yes or no.
2. Logout
3. Login again and check that the question popup is not shown again

Fixes #585

### DIFF
--- a/modules/core/client/services/push.client.service.js
+++ b/modules/core/client/services/push.client.service.js
@@ -17,6 +17,7 @@
     $q) {
 
     var LOCKER_KEY = 'tr.push';
+    var NOT_CLEARED_ON_SIGN_OUT_NAMESPACE = 'notClearedOnSignOut';
 
     var push = {
       isSupported: getIsSupported(),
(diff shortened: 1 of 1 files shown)


simison:
I believe this was designed to behave so that it clears it out on each logout, so that you wouldn't subscribe/unsubscribe to notifications on other people's behalf if multiple people use the same computer.

An in generally just to be good data citizen.

The answer should be remembered between sessions if member does not logout, though. So the problem someone is experiencing is that it gets cleared even if they don't logout.

I'm suspecting their browser is configured to pretty aggressively clear out localstorage since we don't really hear from this issue otherwise.

simison:
Do you think this feature should behave differently or is there a bug that this change solves?

topalavlad:
> I believe this was designed to behave so that it clears it out on each logout, so that you wouldn't subscribe/unsubscribe to notifications on other people's behalf if multiple people use the same computer.

After I reread the reported issue, I think you're right and I couldn't actually reproduce the question on new tab without logout. I'll check it again and comment here if I find anything. Thanks for the review.

PR #817, opened 2018-11-07
welcome emails sender change: Mikael => Dario

guaka:
As discussed with Mikael.
Dario is actually responding to answers to these.

#### Proposed Changes

* "Mikael" => "Dario"


simison:
:ship: it :-)

PR #819, opened 2018-11-07
Search: exclude suspended users from search

simison:
Instead of manually setting their hosting activity to "no", we can just exclude them in the search.

Work in progress

### DIFF
--- a/modules/offers/server/controllers/offers.server.controller.js
+++ b/modules/offers/server/controllers/offers.server.controller.js
@@ -527,6 +527,16 @@ exports.list = function (req, res) {
     },
   });
 
+  // Check for suspended users
+  query.push({
+    $match: {
+      'user.public': true,
+      // We could do this, but since suspended users
+      // are always `public:false`, we don't have to.
+      // 'user.roles': { $ne: 'suspended' }
+    },
+  });
+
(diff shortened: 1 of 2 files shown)


modules/offers/server/controllers/offers.server.controller.js
```diff
@@ -485,21 +485,20 @@ exports.list = function (req, res) {
...
-        as: 'user'
-      }
-    });
-    // Because above `$lookup` returns an array with one user
-    // `[{userObject}]`, we have to unwind it back to `{userObject}`
-    query.push({
-      $unwind: '$user'
-    });
-  }
+  query.push({
```
simison: Noting that these days we're always sending `seen` in queries, so we're anyway doing a lookup to `users` collection.

simison:
rebased

simison:
rebased.  There's a failing test that needs investigation.

simison:
Prolly best to just get rid of the action at this point :-)

simison:
Did some smoke testing. We have some server-side test setup issues so I'm not worried about previously failing test specifically; it's disabled for now and needs returning to (issue https://github.com/Trustroots/trustroots/issues/2524)

PR #820, opened 2018-11-07
Prevent api call when rendering tr-location.client directive

### LINKED ISSUE #814: Question mark as a living location crashes the app

```
```

mrtmrtmlck:
#### Proposed Changes

* Preventing api calls when rendering tr-location.client directive

#### Testing Instructions

* Go Edit Profile -> Locations then observe network whether api call occurs or not

Fixes #814

### DIFF
--- a/modules/core/client/directives/tr-location.client.directive.js
+++ b/modules/core/client/directives/tr-location.client.directive.js
@@ -85,19 +85,6 @@
         vm.searchSuggestions = searchSuggestions;
         vm.onSelect = onSelect;
 
-        // Initialize controller
-        activate();
-
-        function activate() {
-
-          // If directive has init value, use it to search a location
-          if (angular.isDefined($scope.value) && angular.isString($scope.value) && $scope.value) {
-            // If location is found, don't show suggestions list but activate first found result
-            $scope.skipSuggestions = true;
-            searchSuggestions($scope.value);
-          }
-        }
-
         /**
          * Get geolocation suggestions
          */
--- a/modules/search/client/controllers/search.client.controller.js
+++ b/modules/search/client/controllers/search.client.controller.js
@@ -6,7 +6,7 @@
     .controller('SearchController', SearchController);
 
   /* @ngInject */
-  function SearchController($scope, $window, $analytics, $stateParams, $timeout, offer, tribe, Authentication, FiltersService, messageCenterService) {
+  function SearchController($scope, $window, $analytics, $stateParams, $timeout, offer, tribe, Authentication, FiltersService, messageCenterService, LocationService) {
 
     // ViewModel
     var vm = this;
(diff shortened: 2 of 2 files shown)


simison:
Seems like it was there to make this kind of URL query work at search page:

`http://localhost:3000/search?location=London`

Compare what happens in this branch vs at production site.

So this logic just should be moved to search controller(s) in https://github.com/Trustroots/trustroots/tree/master/modules/search/client/controllers

mrtmrtmlck:
> Seems like it was there to make this kind of URL query work at search page:
> 
> `http://localhost:3000/search?location=London`
> 
> Compare what happens in this branch vs at production site.
> 
> So this logic just should be moved to search controller(s) in https://github.com/Trustroots/trustroots/tree/master/modules/search/client/controllers

I've fixed this problem. I hope it is fine. I did my tests and I've some questions as below:

Question 1 (search-map.client.controller.js)
image

Question 2 (tr-location.client.directive.js)
image

simison:
1) `center` seems safer since that's how it has been, although I'd be inclined to test both and see how it behaves for different countries and cities on different zoom levels. I'm assuming the idea of using center by default was to preserve zoom level.

2) I'm not sure if I follow the 2nd question. 


---

BTW you can link lines of code changes like so:

https://help.github.com/articles/getting-permanent-links-to-files/
https://help.github.com/articles/creating-a-permanent-link-to-a-code-snippet/

https://github.com/Trustroots/trustroots/blob/9a3dd5a309e033ff87c1aeb192d785e98867c650/modules/search/client/controllers/search.client.controller.js#L54-L62

Might be easier than screenshots. :-)

mrtmrtmlck:
> 1. `center` seems safer since that's how it has been, although I'd be inclined to test both and see how it behaves for different countries and cities on different zoom levels. I'm assuming the idea of using center by default was to preserve zoom level.
> 2. I'm not sure if I follow the 2nd question.
> 
> BTW you can link lines of code changes like so:
> 
> https://help.github.com/articles/getting-permanent-links-to-files/
> https://help.github.com/articles/creating-a-permanent-link-to-a-code-snippet/
> 
> [trustroots/modules/search/client/controllers/search.client.controller.js](https://github.com/Trustroots/trustroots/blob/9a3dd5a309e033ff87c1aeb192d785e98867c650/modules/search/client/controllers/search.client.controller.js#L54-L62)
> 
> Lines 54 to 62 in [9a3dd5a](/Trustroots/trustroots/commit/9a3dd5a309e033ff87c1aeb192d785e98867c650)
> 
>  if (angular.isDefined(vm.searchQuery) && angular.isString(vm.searchQuery) && vm.searchQuery) { 
>    LocationService.suggestions(vm.searchQuery).then(function (suggestions) { 
>      if (suggestions.length) { 
>        var bounds = LocationService.getBounds(suggestions[0]); 
>        onPlaceSearch(bounds, 'bounds'); 
>        vm.searchQuery = suggestions[0].trTitle; 
>      } 
>    }); 
>  } 
> Might be easier than screenshots. :-)

Thank you :) Below line : https://github.com/Trustroots/trustroots/blob/0ee124f19fa5183039efdc93a0b7a88e541dd9cf/modules/core/client/directives/tr-location.client.directive.js#L98

PR #824, opened 2018-11-08
Validate the length of location value before mapbox api call

### LINKED ISSUE #814: Question mark as a living location crashes the app

mrtmrtmlck:
#### Proposed Changes

* Validate the length of location value before mapbox api call 

#### Testing Instructions

* Type one character then check if there is a request to mapbox api

Fixes #814

### DIFF
--- a/modules/core/client/services/location.client.service.js
+++ b/modules/core/client/services/location.client.service.js
@@ -125,11 +125,11 @@
     function suggestions(val, types) {
       if (!appSettings.mapbox || !appSettings.mapbox.publicKey) {
         $log.warn('No Mapbox settings found; cannot do geocoding!');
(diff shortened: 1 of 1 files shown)


simison:
Seems like the app is expecting a promise instead of an array from this service.

Now if I try to modify location info for my profile,  I see this in console:

image

https://docs.angularjs.org/api/ng/service/$q

mrtmrtmlck:
> Seems like the app is expecting a promise instead of an array from this service.
> 
> Now if I try to modify location info for my profile, I see this in console:
> 
> image
> 
> https://docs.angularjs.org/api/ng/service/$q

I'll take care of it 👍

PR #825, opened 2018-11-08
Prevent httpProvider interceptor work for external domains

### LINKED ISSUE #814: Question mark as a living location crashes the app

mrtmrtmlck:
#### Proposed Changes

* Prevent httpProvider interceptor work for external domains

Fixes #814

### DIFF
--- a/modules/users/client/config/users.client.config.js
+++ b/modules/users/client/config/users.client.config.js
@@ -14,19 +14,20 @@
       function ($q, $location, Authentication) {
         return {
           responseError: function (rejection) {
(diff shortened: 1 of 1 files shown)


simison:
Looks good!

In case of Mapbox API, `rejection.config.url` would've contained the whole URL starting with `https://...`, right?

mrtmrtmlck:
> Looks good!
> 
> In case of Mapbox API, `rejection.config.url` would've contained the whole URL starting with `https://...`, right?

Yes, that's right

PR #649, opened 2018-08-24
References

simison:
Work-in-progress

### Testing

Adding reference
- Go to another user's profile
- Press "add reference"
- Click through the form

Reading reference
- Go to other user's profile
- Press "references"

Reading your own references
- Go to your profile
- Press "references"

PR #827, opened 2018-11-12
Create Reference React

mrkvon:
#### Proposed Changes

* User interface for creating a new reference.
* ~Will need a replacement for [angular-ui-bootstrap](https://www.npmjs.com/package/angular-ui-bootstrap) library.~ (react-bootstrap)
* It depends on ~#790~ #912 by @nicksellen.
* It reuses major parts of code from #649 by @simison.
* ~Currently work in progress.~
* React tests written with `enzyme` (react test helper) and `jasmine` (current client-side test framework).

#### TODO
- [x] replace `angular-ui-bootstrap` nicely
- ~split `ReferencesNew.component` to smart and dummy component~ (wont do now)
- [x] hide behind flags
- ~show `Write a reference` link only if  the reference doesn't exist already~ (probably will do in different PR)
- [x] design looks and info
- fix issues from feedback
  - [x] add some space between the checkbox and text in How do you know them? and next to report checkbox
  - [x] Either make the tabs clickable like the next/back button or not display them as links
  - [x] `Sorry, you cannot leave them a reference if you didn't have any previous interraction.`
typo in the end (_completely reworded_)
  - ~Display only as tooltip for Next button. It is confusing for me, it seems like I am not allowed to access this page~ (just changed the wording to positive one)
  - [x] Make the text next to checkboxes black not gray. It looks like it is disabled.
  - [x] Move the link to your reference in the success display to Your reference for x y is public now. It will look more clear.

For the next iterations, design improvement points by @petr-hajek (I don't want to decide or figure out this) (may also require API changes):
- > Recommendation does not make much sense to me. That does not help me much if I see it in someone's profile. CS also dropped the neutral reference.
- > Met in person is little unclear to me. If it is for cases when I know someone from elsewhere than hosting it maybe should have little different name

##### Automated
* `npm run test:client`

##### Manual
* `npm start`
* Go to `/profile/:username/references/new`

Continued work on #718 
Based and dependent on #912


modules/references/client/components/references.api.js
```diff
@@ -0,0 +1,10 @@
+export async function create(reference) {
+  await fetch('/api/references', {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json; charset=utf-8'
+      // "Content-Type": "application/x-www-form-urlencoded",
+    },
+    body: JSON.stringify(reference) // eslint-disable-line angular/json-functions
+  });
```
nicksellen: Needs a call to `.json()` to fetch the response body, see https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch.

modules/references/client/components/ReferencesNew.component.js
```diff
@@ -0,0 +1,349 @@
...
+            ...this.state.reference.interactions,
+            ...interaction
+          }
+        }
+      });
+    }
+
+    if (field === 'recommend') {
+      this.setState({
+        ...this.state,
```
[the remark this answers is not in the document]
mrkvon: :+1: :white_check_mark: :bowing_woman: 1807d9a

modules/users/client/config/users.client.routes.js
```diff
@@ -358,6 +358,38 @@
           pageTitle: 'Remove profile'
         }
       });
+
+    if (AppConfig.appEnv !== 'production') {
```
mrkvon: I'd really like to replace this with the `appSettings.referencesEnabled`, but I don't know how to provide it in here. I tried, failed and got frustrated.

@simison Do you perhaps know how to provide `appSettings` or `SettingsFactory` here?

simison: Does adding it to `UsersRoutes` -function work?

```
function UsersRoutes($stateProvider, SettingsService) {
  var settings = SettingsService.get();
```

mrkvon: It is failing with error (Firefox) (see the line 3):
```
Error: [$injector:modulerr] Failed to instantiate module trustroots due to:
[$injector:modulerr] Failed to instantiate module users due to:
[$injector:unpr] Unknown provider: SettingsService
http://errors.angularjs.org/1.5.11/$injector/unpr?p0=SettingsService
minErr/<@webpack:///./node_modules/angular/angular.js?:68:12
createInjector/providerCache.$injector<@webpack:///./node_modules/angular/angular.js?:4563:19
getService@webpack:///./node_modules/angular/angular.js?:4716:32
injectionArgs@webpack:///./node_modules/angular/angular.js?:4741:58
invoke@webpack:///./node_modules/angular/angular.js?:4763:18
runInvokeQueue@webpack:///./node_modules/angular/angular.js?:4663:11
...blabla
```

mrkvon: Fixing this will require understanding of inner workings of angularjs and meanjs and I'm not motivated to learn it deeply. I'd rather switch things to React asap.
So unless it's easy, maybe it can stay as it is? When the references are finished, this will be removed anyways. Probably not worth the effort, is it?

petr-hajek:
My comments looking at it at dev:
Cosmetic (can be later iteration): 
- add some space between the checkbox and text in How do you know them?
- Either make the tabs clickable like the next/back button or not display them as links

Logical:
- I don't know recommendation does not make much sense to me. That does not help me much if I see it in someone's profile. CS also dropped the neutral reference.
- Met in person is little unclear to me. If it is for cases when I know someone from elsewhere than hosting it maybe should have little different name

petr-hajek:
`Sorry, you cannot leave them a reference if you didn't have any previous interraction.`
* typo in the end
* Display only as tooltip for Next button. It is confusing for me, it seems like I am not allowed to access this page
* Make the text next to checkboxes black not gray. It looks like it is disabled.
* Move the link to your reference in the success display to <a>Your reference</a> for x y is public now. It will look more clear.

mrkvon:
@petr-hajek I fixed the comments I was able to fix. Thank you for the feedback! :green_heart: 

Noted the rest for further discussion and iterations. Can be added to #718 and meta.trustroots.org.

_Note about feelings:_ I'm finding out that I'm most comfortable with implementing the functionality. I don't want to decide about UX, texts and appearance and I don't derive much pleasure from discussing it. I also suffer when dealing with css. It would be wonderful if there was somebody who enjoys doing these things.

Do we have a trustroots style guide?

mrkvon:
Can we merge this? @simison @guaka

mrkvon:
> can we merge this? @simison @guaka

Sorry, false alert. There is still stuff to do. (axios, wrapping strings for translation)

modules/references/client/components/Interaction.js
```diff
@@ -0,0 +1,59 @@
...
+                type="checkbox"
+                checked={props.reference.interactions.hostedMe}
+                onChange={() => props.onChange('hostedMe')}
+              />
+              They hosted me
+            </label>
+          </div>
+        </div>
+        {(!isInteraction) ? (
+          <div className="alert alert-warning reference-new-tabs-alert" role="alert">
```
simison: Let's remove this badge, it feels like something was broken and it's already obvious from the page without it that user should make a choice at this page:

image

modules/references/client/components/Navigation.js
```diff
@@ -0,0 +1,61 @@
...
+      aria-label="Submit reference"
+      onClick={props.onSubmit}
+      disabled={props.tabDone < props.tabs - 1 || props.disabled}>
+      Submit
+    </button>
+  );
+
+  return (
+    <div className="text-center">
+      {(props.tab > 0) ? back : null}
```
simison: This shorter form should work:
```js
{props.tab > 0 && back}
```

For clarity, let's call these components "backButton".

The same in other places in this PR where ternary + `null` is used.

modules/references/client/components/Interaction.js
```diff
@@ -0,0 +1,59 @@
...
+      <div className="panel-heading">
+        <h4 id="howDoYouKnowThemQuestion">How do you know them?</h4>
+      </div>
+      <div className="panel-body">
+        <div role="group" aria-labelledby="howDoYouKnowThemQuestion">
+          <div className="checkbox">
+            <label>
+              <input
+                type="checkbox"
+                checked={props.reference.interactions.met}
```
simison: `interactions` was extracted from props earlier so we could use the shorter form here too for consistency?

modules/references/client/components/Interaction.js
simison: I guess it was originally meant to be there if user comes back to this view using "back" button and then tries to press next without choosing anything from here?

modules/references/client/components/Info.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+// TODO we can't use the config/config for some reason
+import * as config from '@/config/env/default';
```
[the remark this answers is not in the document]
simison: We could perhaps make API service which simply exposes `window.apiSettings`, much like `settings` Angular service does:

https://github.com/Trustroots/trustroots/blob/f6893632a696d56ecfe64696313ac025112e0169/modules/core/client/services/settings.client.service.js#L48-L49

In future, this might then load settings via API and store in app state somehow.

modules/references/client/components/Navigation.js
mrkvon: :ok_hand: I thought this syntax was unclear, but @agatatalita already explained it to me.

modules/references/client/components/ReferencesNew.component.js
```diff
@@ -0,0 +1,180 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import * as references from './references.api';
+import Navigation from './Navigation';
+import Interaction from './Interaction';
+import Recommend from './Recommend';
```
simison: These component names feel kinda generic if we imagine that in future there will be also "read references" components under this section. I'd love to keep things in separate subfolders for clarity and if that isn't possible, maybe prefix them somehow, maybe just "NewReference"?

mrkvon:
Moved translating to #980, especially to solve issues with testing HOC.

modules/references/client/components/ReferencesNew.component.js
```diff
@@ -0,0 +1,180 @@
...
+    this.setState({
+      reference: {
+        ...this.state.reference,
+        interactions: {
+          ...this.state.reference.interactions,
+          ...interaction
+        }
+      }
+    });
+  }
```
simison: This `handleChangeInteraction` seems pretty complicated and hard to understand what's going on here.

Its toggling one of the values in `interactions`, right?
```
      reference: {
        interactions: {
          met: false,
          hostedThem: false,
          hostedMe: false
        },
        recommend: null
      },
```

...which generally seems like a great use case for [Lodash `update`](https://lodash.com/docs/4.17.11#update).

That said, [nested states are not very React'y](https://stackoverflow.com/a/51136076) and don't provide great performance.

I know we've just inherited them from Angular but ideally, these would just be:
```js
state = {
  interractionMet: false,
  interractionhostedThem: false,
  interractionhostedMe: false,
  recommend: true,
}
```

I think this is already showing signs that we'll need some sort of data layer to handle all this.

At the current state (pun intended) of the application, it'll be messy to deal with using manually building back the object for API calls because we want the data being sent to API in the same format to API as we store it in the DB.

We'd want to store the data in the same format in client side as well, and for that, we'll need a data layer. Components, on the other hand, shouldn't deal with mongoose document structures — they should be super simple and just care about their own state and props.

All this is to say that we'll probably make our future life easier by:
- keeping component state shallow and simple
- manipulating the state back to correct object before sending it to API

modules/references/tests/client/ReferencesNew.client.component.tests.js
```diff
@@ -0,0 +1,192 @@
+'use strict';
+
+import ReferencesNew from '../../client/components/ReferencesNew.component';
+import { Self, Loading, Duplicate, Submitted } from '../../client/components/Info';
```
simison: These component names are pretty compact and don't necessarily describe well what they're doing if just looking at them. :-) Could we rename them for more clarity?

modules/users/client/views/profile/profile-view.client.view.html
```diff
@@ -20,6 +20,11 @@
           Send a message
         </a>
       </li>
+      <li>
```
simison: These navigation items should be protected from showing up with a feature flag.

simison:
>Moved translating to #980, especially to solve issues with testing HOC

Yep, it's okay if we merge first without translations and add translation wrappers later on.

mrkvon:
@simison Thank you for the detailed review! I really appreciate it. :heart: Learning in the process.

simison:
>Learning in the process.

❤️ ❤️ ⚡️

modules/references/client/components/Navigation.js
simison: I just love the simplicity of:

```
{ isEnabled && <Component> }
```
vs
```
{ isEnabled ? <Component> : null }
```

🙃

modules/references/client/components/Info.js
```diff
@@ -0,0 +1,62 @@
...
+export function Self() {
+  return (<div className="alert alert-warning">Sorry, you can&apos;t give a reference to yourself.</div>);
+}
+
+export function Loading() {
+  return (<div className="alert alert-warning">Loading</div>);
+}
+
+export function Duplicate({ userTo }) {
+  return (<div className="alert alert-warning">You&apos;ve already given a reference to <UserLink user={userTo} />.</div>);
```
simison: Translation functions will help with littering `&apos;` in HTML, right? So we’ll be able to:

`{ t( "What's up" ) }`

modules/references/tests/client/ReferencesNew.client.component.tests.js
```diff
@@ -0,0 +1,192 @@
...
+      const userTo = { _id: '222222', username: 'userto' };
+      const stub = sinon.stub(api, 'read');
+      stub.withArgs({ userFrom: userFrom._id, userTo: userTo._id }).resolves([{
+        userFrom, userTo, public: false
+      }]);
+
+      const wrapper = shallow(<ReferencesNew userFrom={userFrom} userTo={userTo} />);
+
+      expect(wrapper.find(Duplicate)).not.toExist();
+      await null;
```
simison: What do these `await null`s do? 🙃

modules/references/client/components/Info.js
```diff
@@ -0,0 +1,62 @@
...
+      <div><a href={`/profile/${userTo.username}/references`}>Your reference</a> for <UserLink user={userTo} /> is public now.</div>
+      <div><a href={`/profile/${userFrom.username}/references`}>See the reference from {name} to you.</a></div>
+      </>
+    ) :
+    (
+      <div>Your reference will become public when <UserLink user={userTo} /> gives you a reference back, or in {config.limits.timeToReplyReference.days} days.</div>
+    );
+
+  return (
+    <div className="alert alert-success">
```
simison: I'll just leave a note for myself to make a pretty design for this step. :-)

This is fine for 1st iteration tho!

modules/references/client/components/Info.js
```diff
@@ -0,0 +1,62 @@
...
+UserLink.propTypes = {
+  user: PropTypes.object.isRequired
+};
+
+export function Self() {
+  return (<div className="alert alert-warning">Sorry, you can&apos;t give a reference to yourself.</div>);
+}
+
+export function Loading() {
+  return (<div className="alert alert-warning">Loading</div>);
```
simison: Let's change this "loading" text to less technical _"Wait a moment..."_ and remove the alert box around it so that it doesn't seem like an error. :-)

modules/references/tests/client/ReferencesNew.client.component.tests.js
mrkvon: It's waiting for the next tick to resolve the `await api.references.read()` in `componentDidMount`. It's similar to `setTimeout(callback, 0)`. 
I could add a comment or figure out a clearer solution. (Any ideas?)

modules/references/client/components/references.api.js
```diff
@@ -0,0 +1,27 @@
+export async function create(reference) {
```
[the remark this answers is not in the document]
mrkvon: I prefer `./api` for now, because `./services` will get mixed up with angular code. Always possible to move it later.

modules/references/client/components/Info.js
mrkvon: Yes, it will.
I'm not sure about the text within [react-i18next's `<Trans>` component](https://react.i18next.com/components/trans-component), though.

modules/references/client/components/references.api.js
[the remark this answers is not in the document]
mrkvon: moved to `./references/client/api/` in 7d8866312d7dd0dfd0f5c4b7e771fc860dffd24a

modules/references/client/components/references.api.js
```diff
@@ -0,0 +1,27 @@
...
+  return await response.json();
+}
+
+export async function read({ userFrom, userTo }) {
+  const response = await fetch(`/api/references?userFrom=${userFrom}&userTo=${userTo}`);
+  return await response.json();
+}
+
+export async function report(user, message) {
+  await fetch('/api/support', {
```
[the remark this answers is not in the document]
mrkvon: 7d8866312d7dd0dfd0f5c4b7e771fc860dffd24a

modules/references/client/components/ReferencesNew.component.js
mrkvon: 4b9a8a585ece7cec53cd399ea79a41a7967bb617 shallowing the component state and manipulating the state from and to API format.

modules/references/client/components/Info.js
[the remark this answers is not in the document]
mrkvon: Do you mean we could have `GET /api/settings?fields=timeToReplyReference,...` will give the needed config value(s) from API?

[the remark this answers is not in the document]
mrkvon: I think I'll hardcode the value for now with a TODO.
Creating the API endpoint later should be easy. I'm attracted by that idea. It makes the app more decoupled from API.

modules/references/client/components/Navigation.js
mrkvon: done

modules/references/client/components/Interaction.js
mrkvon: Removed badge from Interaction and Recommendation components in 85bb82bce1e976251efe4811c6fdc97ee9a096e3.

modules/references/client/components/Info.js
mrkvon: Change the text and remove alert box.

modules/references/client/components/Info.js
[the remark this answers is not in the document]
simison: Good call to hardcode it for now.

API would probably just return all the settings at once and store it in the app state to avoid multiple round-trips.

modules/references/tests/client/ReferencesNew.client.component.tests.js
mrkvon: added comments to explain this

modules/references/client/less/reference-new.less
```diff
@@ -0,0 +1,116 @@
...
+        }
+        &:first-child a:before {
+          display: none;
+        }
+      }
+    }
+  }
+}
+
+.btn-group {
```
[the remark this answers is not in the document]
mrkvon: We're using a select instead of buttons. I guess we can change it.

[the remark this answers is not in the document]
simison: You meant `radio`? I'm happy to look into styling in separate PR as well.

modules/references/tests/client/ReferencesNew.client.component.tests.js
simison: Thanks, yeah I think the comment is warranted. I didn't know you can `await` pretty much anything. 😁 

Reading about this, folks seem to be using `await 'next tick';` sometimes but `await` + comment is very clear now.

This was an interesting read on topic; https://github.com/nodejs/node/issues/19617

modules/references/client/api/references.api.js
```diff
@@ -0,0 +1,70 @@
+import axios from 'axios';
+import set from 'lodash/set';
+import get from 'lodash/get';
+import has from 'lodash/has';
```
simison: Flagging for future work to make  `import { set, get, has } from 'lodash'` not to explode bundle size — as I expect this is a workaround for that?

mrkvon: I found it in a text about bundle size. But I used it, because importing lodash directly was breaking tests including angular.

simison: Oh, funky!

We'll need to look into this more eventually.

modules/references/client/less/reference-new.less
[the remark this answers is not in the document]
mrkvon: I tried to replace this with `ToggleButton` from `react-bootstrap`. Looks similar, just the underline on the selected button is not there.

mrkvon:
I'm merging. Still missing: the logic when only (Recommend: Yes) is allowed. That is - after the other direction reference is published already.

PR #829, opened 2018-11-13
Add user.seen to test data

mrkvon:
Reintroducing changes from #708 removed in #740 
- adds a random `user.seen` date within last year to every created user via `./bin/fillTestData/Users.js`
- additionally fixes script name letter case in `package.json` because `npm run seen` fails on system which cares about filename case

Motivation: without `user.seen` the users aren't displayed in `/search`.


simison:
Good catch!

PR #847, opened 2018-11-18
Improve developer experience with Node based environment checks

### LINKED ISSUE #844: is `npm install` required or not?
According to INSTALL.md in master it's not needed. And that was my experience also, IIRC.
But now when setting up some ansible scripts #828 on fresh servers it seems that it is necessary.to 
Do we need to adapt INSTALL.md?

- Also good to know, what's the difference between `npm ci` and `npm install`?

simison:
Previously "install deps if they're missing" and "ensure config exists" checks weren't very robust.

Now checks are done is Node so they aren't limited to specific OS' and terminals. 

Changes:
- Allow ES6 JS in `/bin/*`
- Using Node.js (previously in bash)
  - Install dependencies if `./node_modules` is missing
  - Copy config file from a template file if it doesn't exist
- Update dependencies if they're out of date compared to package-lock: this is useful when dependencies are different between branches, dependencies got updated in pull etc

These scripts are super well battle-tested at https://github.com/Automattic/wp-calypso/

Resolves #844

### Testing
Ensure:

- npm start / test / build / seed etc commands work as usual.
- `config/env/local.js` is created when it doesn't exist
- `config/env/local.js` symlink is enough to stop config file being copied
- changes in package-lock will cause dependencies being updated
- Missing node_modules folder will cause deps being installed

cc @guaka


simison:
FYI @mrkvon let's observe how this behaves and make changes if something feels off.

PR #854, opened 2018-11-18
Simplify gulp by removing separate prod/dev scripts

simison:
_Work in progress._

The idea is that instead of:

```bash
npm run build:prod
npm run build
```

we would have:
```bash
NODE_ENV=production npm run build
npm run build # defaulting to "development"
```

and the same for:
```bash
NODE_ENV=production npm run start
npm run start # defaulting to "development"
```

It's just a bit annoying how tests are currently set up with their own `NODE_ENV=test` env so this will need some further work to play nicely. Might be I can chop this into two smaller PRs: tests and the rest.

## Testing
- Scripts should default to `NODE_ENV=development` if none is set
- `npm start` and `npm build` should produce different builds with `NODE_ENV=production` and `NODE_ENV=development`
- Different `npm run test:*` scripts should work and not wipe out development database


mrkvon:
Is this still a work in progress? Is it ready for a review?

I don't feel comfortable with gulp, but I can try. :)

simison:
I'd say it's 80% done and up for grabs if you would like to get familiar with it. :-)  Gulp is quite simple.

The problem is that our tests expect `NODE_ENV=test` and manually switch to `development` environment when running the test. That's fine though, we don't need to run tests in `NODE_ENV=production` — just might require shuffling some bits and bobs around to make everything run smoothly. :-)

Issue #828, opened 2018-11-12
ansible: autodeployment, dev staging prod


guaka:
## Good situation
- [x] http://dev.trustroots.org - `master` branch - test database - automatic deployment
- [ ] https://staging.trustroots.org - `master` branch - (copy of) production database? - automatic deploymentn
- [ ] https://www.trustroots.org - `production` branch - production database -  deployment with one manual step

## Action
- Write `ansible` scripts to deploy to a server

## Goals
- Environments where we can quickly show and test things, for devs and non-devs
- Automate more of the process
- Move to cheaper servers (digitalocean is expensive)

-------

## Doesn't require NDA

- [x] InfluxDB installed
- [x] InfluxDB config file modifications and check if working
- [ ] Grafana installed and working
- [ ] Monitoring: disk/Memory/CPU etc monitoring, `monit`, `munin`
- [ ] Auto deploy dev from Travis CI
- [x] `authorized_keys_NDA-signed` and `authorized_keys_no-NDA` (or something like this)
- [ ] shop.trustroots.org https://github.com/Trustroots/shop `prod-conf/nginx-shop.conf`
- [x] migrate node things to PM2 

### Nice to have

- [x] `branch` param for cli
- [ ] run server as separate user (with just enough perms to do this)
- [x] different users on server
- [ ] https://matomo.org/ formerly piwik - this may be a way out of using GA

## Requires NDA access to finish

- [ ] Secrets/configs with https://docs.ansible.com/ansible/2.4/vault.html#what-can-be-encrypted-with-vault
- [ ] Use encrypted variables
- [x] https for staging
- [ ] Think more about using (copy of) production database on staging
- [ ] Deploy prod server configs to staging
- [ ] Production Nginx configs
- [ ] Automatically refreshing certs
- [ ] Weekly and daily backups of production data and assets

guaka:
dev is getting there, main todo:
- [ ] `node` through `passenger`, see https://www.phusionpassenger.com/library/walkthroughs/deploy/nodejs/ownserver/nginx/oss/bionic/install_passenger.html

guaka:
I changed DNS, http://dev.trustroots.org and http://staging.trustroots.org are now working. Still some things:

- [ ] auto deploy dev from CI
- [ ] https for staging, needed?
- [ ] think more about using (copy of) production database on staging

guaka:
[simison wrote](https://github.com/Trustroots/trustroots/pull/863#issuecomment-441380019)

[...]

PR #863, opened 2018-11-24
ansible stuff

guaka:
#### Proposed Changes

* basic ansible to deploy master to dev and staging

Part of #828


simison:
By the looks, this could be merged already and we can start adding features and changes in separate PRs.

simison:
Super quickie list but I think to be at the same state with current production scripts, we’ll still need:

- Production Nginx configs
- Automatically refreshing certs
- InfluxDB installed
- Grafana installed
- Something that does disk/Memory/CPU etc monitoring (currently we use DOs internal tools)
- Backups (currently we use a combination of DOs internal tools and pushing backups)
- Secrets/configs with https://docs.ansible.com/ansible/2.4/vault.html#what-can-be-encrypted-with-vault

Issue #865, opened 2018-11-24
more public stats 


guaka:
**Is your feature request related to a problem? Please describe.**
Some questions I seem to hear a lot:
- How many members are there in [country]?
- How many people are signing up? (e.g. per week)
- How many people are logging in? (e.g. per week)

**Describe the solution you'd like**
Add some more data to https://www.trustroots.org/statistics that can answer these questions.

E.g. 
- graph with number of signups by week
- graph with number of logins by week
- list of countries by number of members (I'm also thinking about SEO here, it's good if we have more content and keywords on the site:)

Is it possible to get this from our grafana?

simison:
I just made https://grafana.trustroots.org/ publicly so that we can link there from `/statistics`

guaka:
mostly good now, let's also add some of this:
- [ ] KPI
- [ ] message stats
- [ ] retention data

PR #873, opened 2018-11-25
adding grafana to stats page

guaka:
part of #865 

- still needs nginx tweaking on prod

### DIFF
--- a/modules/statistics/client/views/statistics.client.view.html
+++ b/modules/statistics/client/views/statistics.client.view.html
@@ -92,7 +92,7 @@
           </div>
           <div class="panel-body">
             <a href="https://grafana.trustroots.org/d/000000002/members?orgId=1">
-              <img class="img-responsive" src="https://grafana.trustroots.org/render/d-solo/000000002/members?orgId=1&from=1419289200000&to=1543142370362&panelId=1&width=800&height=300" width="100%" alt="Trustroots member growth">
+              <img class="img-responsive" src="https://grafana.trustroots.org/render/d-solo/000000002/members?orgId=1&from=1419289200000&to=1543142370362&panelId=1&width=800&height=300&theme=light" width="100%" alt="Trustroots member growth">
             </a>
           </div>
         </div>
--- a/modules/statistics/client/views/statistics.client.view.html
+++ b/modules/statistics/client/views/statistics.client.view.html
@@ -6,6 +6,10 @@
       <h2>Trustroots Statistics</h2>
       <br><br><br>
     </div>
+
+    <!-- this doesn't work yet, nginx cors settings - keeping this up here for now so it's quicker to test -->
+    <img src="https://grafana.trustroots.org/render/d-solo/000000002/members?orgId=1&from=1419289200000&to=1543142370362&panelId=1&width=1000&height=500&tz=Europe%2FMadrid">
+
     <div class="col-xs-12 col-sm-8 col-sm-offset-2 col-md-6 col-md-offset-3 stats">
 
       <div ng-if="!stats.statisticsData.$resolved"
@@ -85,6 +89,13 @@
 
         <hr>
 
+
+        Are you an (aspiring) data scientist? Check <a href="https://grafana.trustroots.org/">our grafana</a> for stats galore.
+
+        Wanna help get better numbers? Consider <a ui-sref="volunteering">joining us</a>.
+
+
+        <hr>
         <h3>Enabling the latent trust between humans since</h3>
         <time class="lead" ng-bind="stats.launchDate | date:mediumDate"></time>
       </section>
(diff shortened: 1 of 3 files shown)


modules/statistics/client/views/statistics.client.view.html
```diff
@@ -6,6 +6,10 @@
       <h2>Trustroots Statistics</h2>
       <br><br><br>
     </div>
+
+    <!-- this doesn't work yet, nginx cors settings - keeping this up here for now so it's quicker to test -->
+    <img src="https://grafana.trustroots.org/render/d-solo/000000002/members?orgId=1&from=1419289200000&to=1543142370362&panelId=1&width=1000&height=500&tz=Europe%2FMadrid">
```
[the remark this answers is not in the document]
guaka: I actually want to show from 2014-12-23, but before that I want it to work.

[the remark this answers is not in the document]
simison: I don't think we have numbers from that early.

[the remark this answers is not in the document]
simison: And folks can always dive in details at Grafana.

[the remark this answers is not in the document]
guaka: We have those numbers.

The CSP headers aren't enough:

image

But also a simple `public/test.html` with this `img` tag doesn't work for me. So I think it's something in nginx.

simison: Maybe something like:
```html
<div class="panel panel-default">
  <div class="panel-heading">
    <h3 class="panel-title">Member growth</h3>
  </div>
  <div class="panel-body">
    <img class="img-responsive center-block" src="https://grafana.trustroots.org/render/d-solo/000000002/members?orgId=1&from=1419289200000&to=1543142370362&panelId=1&width=1000&height=500&tz=Europe%2FMadrid&theme=light" alt="Member growth">
  </div>
</div>
```

- Added wrapping html panel for prettiness
- Added `&theme=light` to url
- Made image responsive so that it plays nicely with mobile screens

simison: Will probably still need to be moved out from the wrapper of stats page since it's quite narrow for an image like this.

guaka:
Total use case for having dev.trustroots.org - ideally it'd be easy to switch to other branches on dev #828

guaka:
Now we can test this at http://dev.trustroots.org/statistics, courtesy of @zrthstr #828

modules/statistics/client/views/statistics.client.view.html
```diff
@@ -85,6 +86,22 @@ <h5 class="text-muted text-uppercase">{{ stats.newsletter | number }} subscriber
 
         <hr>
 
+        <div class="panel panel-default">
+          <div class="panel-heading">
+            <h3 class="panel-title">Member growth</h3>
+          </div>
+          <div class="panel-body">
+            <a href="https://grafana.trustroots.org/d/000000002/members?orgId=1">
+              <img src="https://grafana.trustroots.org/render/d-solo/000000002/members?orgId=1&from=1419289200000&to=1543142370362&panelId=1&width=800&height=300" width="100%>
```
simison: We should use responsive image class `img-responsive` here to make sure layout works nicely with all sizes &devices: https://getbootstrap.com/docs/3.3/css/#images-responsive

public/test.html
```diff
@@ -0,0 +1,9 @@
+Testing CORS.
+
+This file can go once we have this figured out.
+
+<img src="https://upload.wikimedia.org/wikipedia/commons/e/e2/Southern_right_whale.jpg">
+
+<img src="http://hitchwiki.org/en/images/en/thumb/3/39/Trustroots.png/100px-Trustroots.png">
+
+<img src="https://grafana.trustroots.org/render/d-solo/000000002/members?orgId=1&from=1419289200000&to=1543142370362&panelId=1&width=1000&height=500&tz=Europe%2FMadrid">
```
simison: This tested well 👍 let's remember to remove the test file.

modules/statistics/client/views/statistics.client.view.html
simison: Could we add `&theme=light` to this URL to make graph fit better to layout?

guaka:
I think this is good to go.

I don't like the looks of the current KPI board and there's not enough retention data yet.

guaka:
See http://dev.trustroots.org/statistics

simison:
This looks great!

I'd still question fixed `to` values: `&to=1543142370362` — won't that freeze the graph to that time? Does leaving it out or using `now` as a value work?

Sadly docs don't mention it:
http://docs.grafana.org/reference/sharing/#direct-link-rendered-image

http://docs.grafana.org/reference/timerange/#time-units

guaka:
> I'd still question fixed `to` values: `&to=1543142370362` — won't that freeze the graph to that time? Does leaving it out or using `now` as a value work?

I tried `now`, didn't work. I think we can ship this now and we'll find a solution for the `to` param a in the coming 2 months, ideally something in grafana, otherwise we'll generate the current timestamp from angular.

PR #877, opened 2018-11-25
handling 301 redirects inside the code base

guaka:
#### Proposed Changes

* express-redirect

#### Testing Instructions

* go to /invite, end up at /signup (locally)

### DIFF
--- a/modules/core/server/routes/core.server.routes.js
+++ b/modules/core/server/routes/core.server.routes.js
@@ -12,11 +12,11 @@ var path = require('path'),
 
 module.exports = function (app) {
 
-  var redirect = function(src, dst) {
+  var redirect = function (src, dst) {
     app.route(src).get(function (req, res) {
       res.redirect(301, dst);
     });
-  }
+  };
 
   redirect('/invite', '/signup');
   redirect('/tribes/lgbt', '/tribes/lgbtq');
--- a/modules/core/server/routes/core.server.routes.js
+++ b/modules/core/server/routes/core.server.routes.js
@@ -15,8 +15,9 @@ module.exports = function (app) {
 
   redirect(app);
 
+  app.redirect('/invite', '/signup');
   app.redirect('/tribes/lgbt', '/tribes/lgbtq');
-  app.redirect('/tribes/invite', '/signup');
+  app.redirect('/tribes/vegans-vegetarians', '/tribes/veg');
 
   // Gives the service worker access to any config it needs
   app.route('/config/sw.js').get(core.renderServiceWorkerConfig);
--- a/modules/core/server/routes/core.server.routes.js
+++ b/modules/core/server/routes/core.server.routes.js
@@ -4,7 +4,6 @@
  * Module dependencies.
  */
 var path = require('path'),
-    redirect = require("redirective"),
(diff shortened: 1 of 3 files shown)


modules/core/server/routes/core.server.routes.js
```diff
@@ -4,13 +4,20 @@
  * Module dependencies.
  */
 var path = require('path'),
+    redirect = require("express-redirect"),
```
simison: https://www.npmjs.com/package/express-redirect

> _This package isn't actively maintained anymore. It was written for express v3.x, but still seems to work with express v4.x. Use it at your own risk._

> _last publish 3 years ago_

Neat library but I don't think it's a good idea to add it. :-/

modules/core/server/routes/core.server.routes.js
```diff
@@ -4,13 +4,21 @@
...
     tribes = require(path.resolve('./modules/tribes/server/controllers/tribes.server.controller'));
 
+
 module.exports = function (app) {
 
+  redirect(app);
+
+  app.redirect('/invite', '/signup');
+  app.redirect('/tribes/lgbt', '/tribes/lgbtq');
+  app.redirect('/tribes/vegans-vegetarians', '/tribes/veg');
```
simison: We could  handle routes in route configs so that they'll stay in their respective modules.

So on top of [`modules/tribes/server/routes/tribes.server.routes.js`](https://github.com/Trustroots/trustroots/blob/ae4d7a16b5b1ec021c020841dddbc8325f936cc1/modules/tribes/server/routes/tribes.server.routes.js) we can do this:

```js
  app.route('/tribes/lgbt').get(function (req, res) {
    res.redirect(301, '/tribes/lgbtq');
  });
```

`/invite` redirect could go to `modules/users/server/routes/users.server.routes.js`

What do you think?

modules/core/server/routes/core.server.routes.js
guaka: Plugged in another one.

simison: I'm not really sure if saving 3-6 lines of code is really worth external dependency?

https://github.com/Trustroots/trustroots/pull/877#discussion_r236080391

guaka: Well that first one showed up when I googled redirect and express.
If it's 3-6 lines we should not use an external dep.

modules/core/server/routes/core.server.routes.js
guaka: Yeah makes sense. But I'm not yet sure how to do this. Where does the `redirect` function go?

simison: `res.redirect` is built in Express.js: https://expressjs.com/en/4x/api.html#res.redirect

Just adding these three lines to routes.js should be enough, no need to do anything at `express.js`

simison: Yeah ideally tribe ones would be coming from the DB. Next iteration!

simison: On the other hand, having code related to a feature outside the feature's bulk of code itself will make someone else slower because they need to now look for that code. ;-)

This one is fine now, let's 🚢

PR #884, opened 2018-11-26
removing new from links that have been there a while

### LINKED ISSUE #881: Remove "new!" from the Meet link in the menu

guaka:
#### Proposed Changes

* removing 'New!' from menu links to stuff we've had for quite a while

Fixes #881

### DIFF
--- a/modules/pages/client/views/navigation.client.view.html
+++ b/modules/pages/client/views/navigation.client.view.html
@@ -16,7 +16,6 @@
 
   <a ui-sref="offer.meet.list" class="list-group-item">
     Meet
-    <span class="label label-primary">New!</span>
   </a>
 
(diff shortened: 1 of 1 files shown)


modules/pages/client/views/navigation.client.view.html
```diff
@@ -31,7 +30,6 @@ <h4 class="list-group-item-heading" ng-bind="::app.user.displayName"></h4>
      ui-sref="invite"
      class="list-group-item">
     Invite Friends
-    <span class="label label-primary">New!</span>
   </a>
```
simison: While at it, should we remove this whole "invite friends" link since invitations are now disabled anyway?

guaka: good call!

PR #902, opened 2018-12-01
Add command line options for seeding

### LINKED ISSUE #843: npm run seed --options
For deployment #828 it would be great to have these options for `npm run seed`:

- [ ] `--quiet`, limit the output from the command to the essentials
- [ ] `--only-if-empty`, only seed if there's no data yet

(maybe other names are better:)

Jagjeet:
#### Proposed Changes

* Switch to using 'yargs' with the same usage for all seeding scripts
* Adds the following command line options naming previous command line options if required
  * --verbose -enables database output which by default is now turned off. 
  * --limit -only seeds up to the number of items if there is already data in the database. 

#### Testing Instructions

* TBD

Fixes #843

### DIFF
--- a/bin/fillTestData/Messages.js
+++ b/bin/fillTestData/Messages.js
@@ -5,13 +5,17 @@ var _ = require('lodash'),
     mongooseService = require(path.resolve('./config/lib/mongoose')),
     chalk = require('chalk'),
     argv = require('yargs')
-      .usage('Usage: $0 <number of threads to add> <max messages per thread>')
+      .usage('Usage: $0 <number of threads to add> <max messages per thread> {options}')
+      .boolean('verbose')
+      .describe('verbose', 'Enable extra database output (default=false)')
       .demandCommand(2)
+      .example('node $0 100 10', 'Adds 100 random threads wth up to 10 messages per thread to the database')
+      .example('node $0 100 10 --verbose', 'Adds 100 random threads wth up to 10 messages per thread to the database with verbose database output')
       .check(function (argv) {
         if (argv._[0] < 1) {
           throw new Error('Error: Number of threads should be greater than 0');
         }
-        else if (argv._[1]) {
+        else if (argv._[1] < 1) {
           throw new Error('Error: Max messages per thread should be greater than 0');
         }
         return true;
@@ -19,7 +23,8 @@ var _ = require('lodash'),
       .strict()
       .argv,
     faker = require('faker'),
-    mongoose = require('mongoose');
+    mongoose = require('mongoose'),
+    config = require(path.resolve('./config/config'));
 
 var random = function (max) {
   return Math.floor(Math.random() * max);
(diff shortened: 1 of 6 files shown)


bin/fillTestData/Messages.js
```diff
@@ -4,8 +4,27 @@ var _ = require('lodash'),
     path = require('path'),
     mongooseService = require(path.resolve('./config/lib/mongoose')),
     chalk = require('chalk'),
+    argv = require('yargs')
```
simison: For  clarity, let's separate requiring the module and configuring it;
```js
argv = require('yargs');
```

and further down in the app:
```js
argv
  .usage()
  // ...etc
```

bin/fillTestData/Messages.js
```diff
@@ -17,17 +36,25 @@ var addDays = function addDays(date, days) {
   return result;
 };
 
-var addThreads = function (numThreads, maxMessages) {
+var addThreads = function () {
   var index = 0;
+  var numThreads = argv._[0];
+  var maxMessages= argv._[1];
```
simison: I believe in Yargs these can be grabbed with `positional` args: http://yargs.js.org/docs/#api-positionalkey-opt rather than the accessing internal object.

bin/fillTestData/Tribes.js
```diff
@@ -36,66 +55,90 @@ var tribeImageUUIDs = [
...
   }
 
-  console.log(chalk.white('--'));
-  console.log(chalk.green('Trustroots test tribes data'));
-  console.log(chalk.white('--'));
+  if (verbose) {
+    console.log(chalk.white('--'));
+    console.log(chalk.green('Trustroots test tribes data'));
+    console.log(chalk.white('--'));
+  }
```
simison: I feel that this we could include regardless of `verbose` setting? Since the script takes a while, would be great to know that it has started.

bin/fillTestData/Tribes.js
```diff
@@ -36,66 +55,90 @@ var tribeImageUUIDs = [
...
+
+  // Override debug mode to use the option set by the user
+  config.db.debug = verbose;
```
simison: Would it work better to call the `--verbose` `--debug` instead?

bin/fillTestData/Messages.js
Jagjeet: Thanks!  I was looking for that.

bin/fillTestData/Tribes.js
Jagjeet: Thanks for all the feedback! I am totally fine with '--debug'. I just wanted to make sure it was something that expressed 'loud' since 'quiet' is now the default.

Jagjeet:
The debug/limit modes are working though I still have a few outstanding issues to work on with this: 

-With the debug data not in the way, it has made the save failures more obvious and sometimes far less items are actually added to the database than expected so I am going to try and fix those. 
-It might make sense to print a '.' for each added item similar to the way some test suites work to convey status and a summary of what actually was added (eg. If you use the limit flag, depending on how many items are in the database, you'll only add the remaining ones). Let me know if anyone has other thoughts.

guaka:
Nice! This will plug nicely into the work @zrthstr has been doing on ansible.

Jagjeet:
I am back to working on this today. The Tribes script is more or less ready to go. I am doing the same changes to Messages and then Users so let me know if you have any thoughts related to usage or otherwise. I haven't tested with a live version of Trustroots due to my build issues, but didn't change much specific to the data so think things should be okay. That said, definitely want to test it out before pulling in the changes.

Jagjeet:
@simison - All the scripts are working now. I fixed the various hangs and collisions I was seeing so all the items are actually added now. That said, the performance for the User script could use work especially when there is a large numbers of tribes and adding large numbers of users. Also, it could do a better job displaying errors when there are name collisions if passed in by the user. I can try and look into the performance more this week or as part of another PR. Let me know if you have other feedback or still see any errors.

Jagjeet:
Awesome! I'll do the rebase this evening. Anything I need to know for merging?

simison:
FYI @guaka @zrthstr when this merges, you can change seeding data in the Ansible scripts, too.

Jagjeet:
@simison Easy enough! Thanks! :)

PR #905, opened 2018-12-03
Add sign-up: how did you hear about Trustroots? #879

RGHuynh:
WIP

* Add input box "how the you hear about Trustroots" to signup
* Send user responses to InfluxDB

Fixes #
issue #879


petr-hajek:
@RGHuynh Hey, planning to continue working on this?

modules/users/server/controllers/users.authentication.server.controller.js
```diff
@@ -84,6 +84,20 @@ exports.signup = function (req, res) {
 
     },
 
+    // Report how did you hear about us to stats
+    function () {
```
guaka: I'm no async wizard ✨  either but it Looks like this function should receive a callback var `done`, which then should be called in `statService.stat`'s callback, I'm testing this now in #970.

RGHuynh:
> @RGHuynh Hey, planning to continue working on this?

Yeah, I'm just stuck where i can't figure out how to send the user response to influxDB

guaka:
continue with #970

Issue #910, opened 2018-12-04
CI should include `npm run build:prod`


guaka:
If `npm run build:prod` doesn't succeed CI shouldn't give green light.

Example #790 passed, but ends up being a problem when building for prod.

This relates to #828, as ideally CI would simply build a complete new server from scratch.

simison:
>as ideally CI would simply build a complete new server from scratch.

it does :-)

PR #911, opened 2018-12-04
Revert "Add React and an example component"

mrkvon:
Reverts Trustroots/trustroots#790


package.json
```diff
@@ -170,18 +170,11 @@
...
-    "@babel/preset-env": "^7.1.0",
-    "@babel/preset-react": "^7.0.0",
     "agendash": "~0.4.0",
-    "babel-loader": "^8.0.4",
     "concurrently": "~4.1.0",
-    "eslint-plugin-react": "^7.11.1",
     "exports-loader": "^0.7.0",
     "expose-loader": "^0.7.5",
     "faker": "~4.1.0",
-    "firebase": "^3.9.0",
```
simison: just noting that this looks a bit off, but might've been in original PR as well. I remember fixing having this in both `devDependencies` and `dependencies`

guaka:
`npm run build:prod` also passed :) #910

simison:
Just confirmed that this was removed `firebase` and thus broke notifications:

https://github.com/Trustroots/trustroots/pull/911#discussion_r238818182

Let's loop back to add it there again.

PR #912, opened 2018-12-04
React

guaka:
Continuation of #790 after #911 

`npm run build:prod` broke #910 

Nice to have merged before this: #913


package.json
```diff
@@ -170,11 +170,18 @@
...
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-react": "^7.0.0",
     "agendash": "~0.4.0",
+    "babel-loader": "^8.0.4",
     "concurrently": "~4.1.0",
+    "eslint-plugin-react": "^7.11.1",
     "exports-loader": "^0.7.0",
     "expose-loader": "^0.7.5",
     "faker": "~4.1.0",
+    "firebase": "^3.9.0",
```
simison: Let's add this back to `master` in a separate PR.

https://github.com/Trustroots/trustroots/pull/911#issuecomment-444243454

config/webpack/entries/main.js
```diff
@@ -6,7 +6,58 @@
...
+ *  Uses a webpack require context
+ *  See https://webpack.js.org/guides/dependency-management/#require-context
+ */
+importComponents(require.context('../../../modules/', true, /\.component\.js$/));
+
+function importComponents(r) {
+  r.keys().forEach(path => {
+    const Component = r(path).default;
+    const name = extractComponentNameFromPath(path);
+    if (name !== Component.name) {
```
mrkvon: What is the motivation for this check @nicksellen? The application running via `start:prod` fails here and the check doesn't seem to be required. Maybe we could disable this in production env?

modules/core/client/services/photos.service.js
```diff
@@ -0,0 +1,176 @@
...
+
+function getApplicationScope() {
+  return angular.element(document.querySelector('html')).scope();
+}
+
+export function selectPhoto(name) {
+  const photo = photos[name];
+  const file = (window.innerWidth <= 480 && photo.file_mobile) ? photo.file_mobile : photo.file;
+  const imageUrl = `/img/board/${file}`;
+  getApplicationScope().$emit('photoCreditsUpdated', { [name]: photo });
```
mrkvon: What is the motivation for having this @nicksellen? It doesn't seem to be required for the application to function and the application fails when it's run with `start:prod` script.
(I also changed the ng-annotate-loader in `config/webpack/webpack.config.js` to babel-plugin-angularjs-annotate, so maybe these two fail when together.)

mrkvon: I changed this in 68dece5. To see the original, see the branch at 5efdfa6.

config/webpack/entries/main.js
mrkvon: I changed this in 68dece5. To see the original, see the branch at 5efdfa6.

mrkvon:
Merged. Happy React!

simison:
Happy React to everyone! 🎉 

Huge kudos to both @mrkvon @nicksellen 👏

PR #921, opened 2018-12-06
Internationalization (i18n) with react-i18next

### LINKED ISSUE #492: Support multiple languages (i18n)
Might get a lot more support if people can use the site in their native language.

[angular-translate](https://angular-translate.github.io/) is a great, mature library for angularjs.

For managing the translations I just started trialling [crowdin](https://crowdin.com) for a blog. Seems great! And they are into open source [1].

Another option is Transifex which we use for [foodsaving.world](https://foodsaving.world) - see [our dashboard](https://www.transifex.com/yunity-1/foodsaving-tool/dashboard/). But they have gone a bit cold on open source projects (there used to be a whole page on their website [2]), but now just a bit on the [pricing page](https://www.transifex.com/pricing/), and a few more criteria than just open source:

> We offer Transifex for free to Open Source projects that have no funding, revenue, and/or commercialization model. 

[1] https://crowdin.com/page/open-source-project-setup-request
[2] https://www.transifex.com/customers/open-source/

mrkvon:
#### Proposed Changes

* Testing an react-i18next library. So far works fine.

Largely [following](https://react.i18next.com/guides/quick-start) [tutorials](https://react.i18next.com/guides/step-by-step-guide) on https://react.i18next.com.

#### Testing Instructions

* `git checkout react-i18next`
* `npm install`
* `npm start`
* go to http://localhost:3000/volunteering
* try to switch language from `eng` to `cze` and observe the magic

Fixes #492

TODO:
- [x] hide behind a feature flag
- [ ] test translation extraction tools and translating tools


modules/core/client/components/LanguageSwitch.component.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import i18n from './i18n';
+
+export default function LanguageSwitch() {
+  const changeLanguage = (lng) => {
+    i18n.changeLanguage(lng);
```
simison: Let's avoid shortening things when possible to make code easier to read? `code` or `language` would read nicely :-)

modules/core/client/components/i18n.js
```diff
@@ -0,0 +1,30 @@
+import i18n from 'i18next';
+import { reactI18nextModule } from 'react-i18next';
+import * as translationEN from '../../../../public/locales/en/translation';
+import * as translationCS from '../../../../public/locales/cs/translation';
```
simison: We probably better off starting with some ISO standard codes; perhaps the same as we use for [profile languages](https://github.com/Trustroots/trustroots/tree/15a8bcebee8286eca1591f950a120420b2e08dd3/config/languages)? (ISO 639 2)

simison:
What do you think if we instead do this:

```jsx
const Volunteering = () => {
 return <Something />;
}

export default withNamespaces()(Volunteering);
```

That way if function/component is returning multiple times, we wouldn't need to wrap each return with `NamespacesConsumer`:

```jsx
export default const Volunteering = () => {
  if (something) {
    return (
      <NamespacesConsumer>
        <Something />
      </NamespacesConsumer>
    );
  } 

  return (
    <NamespacesConsumer>
      <Something />
    </NamespacesConsumer>
  );
}
```

mrkvon:
> use withNamespaces higher order component

It's nice, yes. There is just [a check](https://github.com/Trustroots/trustroots/blob/master/config/webpack/entries/main.js#L37) in the code for importing react components to angular, which is failing with this. It could be possible to drop that check, too.

modules/core/client/components/i18n.js
mrkvon: Ok, let's use the ISO 639-2 instead of 639-1. It includes more languages and we already use it.

mrkvon:
It's very important that we figure out what localization tool we can use. (A tool that non-programmers can use to translate the files in `/public/locales/**/translation.json`.) It would preferably be free and open source and run on different operating systems.

It makes a difference between choosing this library and not.

Thank you @Akronix for pointing this out to me.

modules/core/client/components/i18n.js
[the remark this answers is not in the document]
mrkvon: Done as described in [an official tutorial](https://react.i18next.com/guides/step-by-step-guide#a-adding-lazy-loading-for-translations).

simison:
>It's very important that we figure out what localization tool we can use.

Good point! I don't see us being limited to anything specific, because i18n-next has utils for to/from Gettext and other formats as well even a Gettext Webpack loader:

https://www.i18next.com/overview/plugins-and-utils#utils

Gettext is about as free and open as you can get so we can either use self-hosted software, SAAS software to even local tools such as [Poedit](https://poedit.net/).

i18n-next seems great also because it has support for such a variety of frameworks, among others Express.js ([`i18next-express-middleware`](https://github.com/i18next/i18next-express-middleware)) which we use at the backend so we can translate our emails and other backend served stuff.

mrkvon:
A design of language switch button would be nice.
How should it behave, where should it be placed?

simison:
>A design of language switch button would be nice.
>How should it behave, where should it be placed?

For non-authenticated users, it could always be visible in the header. We can get quickly going with a simple Bootstrap navigation dropdown:

image

It should fit also in mobile header nicely:

image

(and if it doesn't, using a shortening "FI" for Finnish etc should work.


For authenticated users we don't need to show it all the time (as they really need to change it just once). Instead, we can add it to their "account" page as `<select>`:

image

>This is the language of the interface you see across the site.
>
> Thanks to all our community members who helped translate! You can help us out!


Probably good to hide these selections behind a feature flag until we have something to show?

mrkvon:
> What were you thinking specifically when asking "How should it behave"?

@simison You answered precisely what I was asking. Thank you!

modules/core/client/components/i18n.js
simison: Ahhh bummer, I think I've looked up something wrong here. 

I picked the ISO standard from this sentence:

>The `languages.json` is a custom made file which contains languages from `languages_orig.json` that has `iso_639_2` standard defined.

But it just states that we pick only those languages from the whole massive list and then actually in the database we use codes like `fi_FI`:

https://github.com/Trustroots/trustroots/blob/15a8bcebee8286eca1591f950a120420b2e08dd3/modules/core/client/directives/tr-languages.client.directive.js

mrkvon: Would you like to revert to ISO_639-1, then? It should be as easy as dropping a commit.

Actually, ISO_639-1 will be compatible with `moment.locale()`, i18next documentation etc.
And when we need to use languages out of the scope of ISO_639-1, we can use ISO_639-3 just for them.

I would prefer not to use `fi_FI` though - not a standard, redundant information and hard to reuse in localization library configs without custom parsing.

simison: Sounds reasonable 👍  I'll leave it up to you to decide since you have better insight. :-) We can very easily refactor at any point if we hit issues.

Do ISO_639-1 or ISO_639-2 make difference between localized languages, such as English GB/US or French CA/FR etc? In volunteering space those are often pretty fast to crop up.

mrkvon: Ah, I made some false assumptions (i.e. `fi_FI` is _not_ a redundant information). I'll research more.

mrkvon:
We can save the preferred language in a client (set a cookie) or in a database (`PUT /api/users`, `{ "localization": "en_GB", ... }`). Which one?
Probably a separate PR.

simison:
>We can save the preferred language in a client (set a cookie) or in a database (PUT /api/users, { "localization": "en_GB", ... })

Cookie sounds good to me as a starting point since that's what unregistered users need.

For authenticated users, we can store it in their user table just like any other account content.

> Probably a separate PR.

Totally 👍

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,16 +1,19 @@
 import React from 'react';
+import { NamespacesConsumer } from 'react-i18next';
+import '@/modules/core/client/components/i18n';
 import PropTypes from 'prop-types';
-import moment from 'moment';
 
 export default function ProfileViewBasics({ profile }) {
-  return <>
+  return (<NamespacesConsumer ns="profile-view-basics">{ t => (<>
```
simison: I'd imagine we want to bundle all the translations of `user` module into one to keep thing simple?

modules/core/client/components/LanguageSwitch.component.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import i18n from './i18n';
+import PropTypes from 'prop-types';
+import { Dropdown, MenuItem } from 'react-bootstrap';
```
simison: Does this import Bootstrap v3 or v4 HTML?

We're still using v3 stylesheets so we'd need v3 HTML.

mrkvon: :+1: 

I'm in the process of doing that now.

`react-bootstrap` uses [bootstrap v3](https://react-bootstrap.github.io/).

modules/core/server/views/partials/header.server.view.html
```diff
@@ -196,4 +196,8 @@
   </div>
   <!-- /Header for authenticated users -->
 
+  <div class="container">
+    <language-switch></language-switch>
+  </div>
```
simison: Shall we have this at the beginning behind feature flag and make it visible once the experience is more pleasant? Now there won't really be anything to look at and it can be a bit confusing.

modules/core/client/components/LanguageSwitch.component.js
mrkvon: moved to #957

modules/users/client/components/ProfileViewBasics.component.js
mrkvon: Changed to `user-profile`. Generally an easy fix.

mrkvon:
Wants to be merged. Somebody should succeed with using the tools for extracting translations, and tools for translating. I'll try tomorrow.
We can continue with this only when we know both programmers and non-programmers can create translations frustration free.

mrkvon:
@guaka 
[Exctracting translations](https://react.i18next.com/guides/extracting-translations) - [react-i18next documentation](https://react.i18next.com/)

Also we should decide whether we want to use [natural, or keybased catalog](https://react.i18next.com/guides/step-by-step-guide#3-sidequest-natural-vs-keybased-catalog). Currently we use natural ones (translation keys equal English values).

[The i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter) would be nice to have plugged to make many translation tools available. Ideally we wouldn't need the i18next json files at all.

simison:
> we should decide whether we want to use natural, or keybased catalog. Currently we use natural ones (translation keys equal English values).

I'd go for natural like it's now — it removes some mental complexity while developing and moves the overhead of "changing strings" to translation work instead. I think that's fine since we aren't changing strings _that often_.

simison:
This seems to be a bit broken now on development version:

image

Let's put the flag off by default for all environments and folks working on this can enable locally?

mrkvon:
Ok, let's make a leap of trust that the localization tools will work.
Merging.

simison:
As long as things are behind a feature flag and not breaking anything or making the build enormous, we're safe to iterate and refactor. 👍

PR #925, opened 2018-12-07
Remove extra ellipsis. It is handled by css

### LINKED ISSUE #920: Ellipsis is displayed for non-truncated message in message preview
**Describe the bug**
When the message is longer, it gets truncated with ellipsis `...` in the message preview. But the ellipsis is shown even for the messages, that are not long enough to be truncated.

**To Reproduce**
1. Send short message, e.g. Hello
2. Click on Messages
3. The message preview says Hello ...

**Expected behavior**
The message preview says Hello (no ellipsis)

Linux Mint - Firefox 63.0.3

petr-hajek:
+ renamed textServiceService to textService

Fixes #920 No need to add ellipsis here, it is handled by css by text-overflow.

### DIFF
--- a/modules/messages/server/controllers/messages.server.controller.js
+++ b/modules/messages/server/controllers/messages.server.controller.js
@@ -14,7 +14,7 @@ var _ = require('lodash'),
     messageToStatsService = require(path.resolve('./modules/messages/server/services/message-to-stats.server.service')),
     messageStatService = require(path.resolve('./modules/messages/server/services/message-stat.server.service')),
     errorService = require(path.resolve('./modules/core/server/services/error.server.service')),
-    textServiceService = require(path.resolve('./modules/core/server/services/text.server.service')),
+    textService = require(path.resolve('./modules/core/server/services/text.server.service')),
     userProfile = require(path.resolve('./modules/users/server/controllers/users.profile.server.controller')),
     Message = mongoose.model('Message'),
     Thread = mongoose.model('Thread'),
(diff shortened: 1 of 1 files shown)


modules/messages/server/controllers/messages.server.controller.js
```diff
@@ -14,7 +14,7 @@ var _ = require('lodash'),
     messageToStatsService = require(path.resolve('./modules/messages/server/services/message-to-stats.server.service')),
     messageStatService = require(path.resolve('./modules/messages/server/services/message-stat.server.service')),
     errorService = require(path.resolve('./modules/core/server/services/error.server.service')),
-    textServiceService = require(path.resolve('./modules/core/server/services/text.server.service')),
```
simison: haha, nice catch!

modules/messages/server/controllers/messages.server.controller.js
```diff
@@ -84,7 +84,7 @@ function sanitizeThreads(threads, authenticatedUserId) {
 
     // Clean message content from html + clean all whitespace + shorten
     if (thread.message) {
-      thread.message.excerpt = thread.message.content ? textServiceService.plainText(thread.message.content, true).substring(0, 100).trim() + ' …' : '…';
+      thread.message.excerpt = thread.message.content ? textService.plainText(thread.message.content, true).substring(0, 100) : '…';
```
simison: Perhaps we could send a bit longer excerpt if we let CSS anyway do the truncation?

petr-hajek: We could, but it is anyway truncated to 50 chars so I just left it as it is.

PR #927, opened 2018-12-07
Docs: adding new icons

simison:
Adds docs about Fontello icons.


simison:
Just tested this and seems like these instructions are incorrect; `fontello` task is missing from the Gulpfile for some reason.

The function exists so it'll be easy enough to fix:

https://github.com/Trustroots/trustroots/blob/1c3430fdd1079f358d1c789126c51cc5d4d8acd1/gulpfile.js#L400-L411

PR #928, opened 2018-12-07
Improve email templates for References

### LINKED ISSUE #849: References notification templates
As part of #718 we've implemented references API. Notifications sent out via emails and push notifications need copy and templates. Read more: https://github.com/Trustroots/trustroots/issues/718#issuecomment-431699368

- [ ] email templates
- [ ] push notification templates

mrkvon:
#### Proposed Changes

* Improved email templates for references
* DRYing templates a bit
* Further suggestions for improvements are very welcome!

#### Currently looks like this:
Notification for user1 that user2 gave a reference:
reference_notification_1

Notification for user2 that user1 gave a reference back:
reference_notification_2

TODO show text email templates

#### Testing instructions
TODO (has to create references via API)

Work on #849

### DIFF
--- a/modules/core/server/views/email-templates/email-with-container.server.view.html
+++ b/modules/core/server/views/email-templates/email-with-container.server.view.html
@@ -1,0 +1,59 @@
+{% extends './email.server.view.html' %}
+
+{% block content %}
+
+  <!-- MODULE ROW // -->
+  <tr>
+      <td align="center" valign="top">
+          <!-- CENTERING TABLE // -->
+            <!--
+              The centering table keeps the content
+                tables centered in the emailBody table,
+                in case its width is set to 100%.
+            -->
+          <table border="0" cellpadding="0" cellspacing="0" width="100%">
+              <tr>
+                  <td align="center" valign="top">
+                      <!-- FLEXIBLE CONTAINER // -->
+                        <!--
+                          The flexible container has a set width
+                            that gets overridden by the media query.
+                            Most content tables within can then be
+                            given 100% widths.
+                        -->
+                      <table border="0" cellpadding="0" cellspacing="0" width="600" class="flexibleContainer">
+                          <tr>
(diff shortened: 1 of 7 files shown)


petr-hajek:
I'd rephrase the last paragraph to:
You have two weeks to do this...

guaka:
todo: add link to profile in "You are receiving this email because you have an active account on Trustroots." This should simply be a matter of adding `user` var to `params` now.

image

modules/core/server/views/email-templates/email-with-container.server.view.html
```diff
@@ -0,0 +1,59 @@
...
+  <!-- MODULE ROW // -->
+  <tr>
+      <td align="center" valign="top">
+          <!-- CENTERING TABLE // -->
+            <!--
+              The centering table keeps the content
+                tables centered in the emailBody table,
+                in case its width is set to 100%.
+            -->
+          <table border="0" cellpadding="0" cellspacing="0" width="100%">
```
guaka: Is there a way to make this DRYer?
I understand we need messy tables like this for emails, but would be good to keep it limited to one spot, if possible?

May be best to open an issue for this for now, as there are a dozen of these:
image

simison: Yep, that's a refactor long due.

Basically, those tables would need to be removed to a layout and reused each time with `{% extends './layout.server.view.html' %}`

simison: Templates are Nunjucks templates: https://mozilla.github.io/nunjucks/

mrkvon: There is an attempt for drying in this PR. See ` modules/core/server/views/email-templates/email-with-container.server.view.html`. It assumes a message and an action button at the end.

mrkvon:
Seems ready to me. Merging.

PR #931, opened 2018-12-08
Update signin.client.view.html

copyhold:
Login: make signup button at login page more prominent #891 
as a solution to [this](https://github.com/Trustroots/trustroots/pull/891)

### DIFF
--- a/modules/users/client/views/authentication/signin.client.view.html
+++ b/modules/users/client/views/authentication/signin.client.view.html
@@ -73,7 +73,7 @@
             </fieldset>
           </form>
           <!-- For bigger screens -->
-          <div class="text-white hidden-hs">
+          <div class="text-white hidden-xs">
             <br>
             <a ui-sref="signup" class="btn btn-inverse btn-lg btn-signup">Become a member</a>
             <span>&nbsp;&nbsp;or&nbsp;&nbsp;</span>
--- a/modules/users/client/views/authentication/signin.client.view.html
+++ b/modules/users/client/views/authentication/signin.client.view.html
@@ -72,7 +72,8 @@
               </div>
             </fieldset>
           </form>
-          <div class="text-white">
+          <!-- For bigger screens -->
+          <div class="text-white hidden-hs">
             <br>
             <a ui-sref="signup" class="btn btn-inverse btn-lg">Become a member</a>
             <span>&nbsp;&nbsp;or&nbsp;&nbsp;</span>
@@ -81,7 +82,6 @@
         </div>
       </div><!-- .row-->
 
-      <!-- For bigger screens -->
 
 
       <!-- For smaller screens -->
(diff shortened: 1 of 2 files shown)


modules/users/client/views/authentication/signin.client.view.html
```diff
@@ -49,42 +49,40 @@
...
+                   tabindex="4">
+                  Forgot?
+                  <span ng-if="auth.authError" class="hidden-xs">
+                    Recover your password
+                  </span>
+                </a>
               </div>
             </fieldset>
           </form>
+          <div class="text-white">
```
petr-hajek: it is missing hidden-xs class so it is displayed twice in the mobile view
selection_007

modules/users/client/views/authentication/signin.client.view.html
```diff
@@ -49,42 +49,40 @@
...
-                <span ng-if="auth.authError" class="hidden-xs">
-                  Recover your password
-                </span>
-              </a>
-              <div class="text-center form-group">
-                <br><br>
+              <div class="form-group">
                 <button type="submit"
-                        class="btn btn-default btn-action"
+                        class="btn btn-default btn-lg"
```
petr-hajek: I'd leave it as it was so Login button is bigger

modules/users/client/views/authentication/signin.client.view.html
petr-hajek: Also the comment `<!-- For bigger screens -->` should be before this div

modules/users/client/views/authentication/signin.client.view.html
copyhold: `btn-action` looks too big to my taste. It is also not a Bootstrap standard.

simison: >It is also not a Bootstrap standard.

Bootstrap is just a base framework and not a design system per-se, it's okay to have additional components. ;-)

`btn-action` is big indeed, it's for "call to action" type of interactions and not designed to be used in forms.

modules/users/client/views/authentication/signin.client.view.html
```diff
@@ -72,7 +72,8 @@
               </div>
             </fieldset>
           </form>
-          <div class="text-white">
+          <!-- For bigger screens -->
+          <div class="text-white hidden-hs">
```
petr-hajek: It is class hidden-xs
Now it is still displaying both

modules/users/client/views/authentication/signin.client.view.html
petr-hajek: OK, fair enough

modules/users/client/views/authentication/signin.client.view.html
copyhold: 7a1b22f6ae92ffc04df004b818f2341ce4ce5c2c

PR #932, opened 2018-12-08
Add Prettier - an opinionated code formatter

### LINKED ISSUE #858: prettier: Opinionated Code Formatter
https://github.com/prettier/prettier

simison:
- Adds a git pre-commit hook that automatically formats your code.
- Reformats the code in this same PR

Eslint rules still need work, since now "prettier" rules are applied everywhere.

https://prettier.io/docs/en/eslint.html

Resolves #858

### Testing
- When you make changes to files and `git commit`, Prettier should process those files.


simison:
Can be useful here: https://www.npmjs.com/package/pretty-quick

ahstro:
What still needs to be done for this to be merged? Prettier is so convenient :heart_eyes:

simison:
@ahstro I think the path for this starts to be clear; we've converted codebase to ES6, consolidated different Eslint configs and we're working on [removing stuff from Gulp setup, too](https://github.com/Trustroots/trustroots/issues/1149).

Wanna help out with this? :-)

simison:
We have some choices of different strategies here — any preferences which way to go?

1) Aggressive reformatting on commit
- Each staging file (in `git commit`) would get automatically run by Prettier
- Those files would also get  `/* @format */` pragma added
- Your editor will run Prettier only on files marked with `@format`
- Everything just goes on automatically as we touch files (and someday we can reformat everything remaining)
- Causes more noise in PRs since lots of unrelated changes will be in each file

2) Manually reformat first
- Any file that we want in Prettier has to be formatted and committed to `master` first.
- We'll have `/* @format */` pragma on those files
- Your editor will run Prettier only on files marked with `@format`
- Prettier will run for staged files on `git commit` only if a file has `@format` pragma, and otherwise keep them as-is
- More manual work up-front, less noise in PRs

3) Prettier everything
- Do a mega PR by Prettier-reformatting _everything_ at once
- No `@format` pragma, Prettier always runs on all files (in your editor and on `git commit`)
- Just works — HUUUGE PR — Potential rebase nightmare for open PRs?

simison:
I tested with some files and unrelated Prettier changes are so much noise that I think we're better off with option 2. One big mega PR we can do once we're doing PRs mainly to Prettified files so that we don't end up in rebase nightmare.

package.json
```diff
@@ -250,5 +254,16 @@
...
     "url": "https://github.com/Trustroots/trustroots/issues"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run -s install-dependencies && npx lint-staged"
+    }
+  },
+  "lint-staged": {
```
simison: An alternative for `lint-staged` could be [`pretty-quick](https://www.npmjs.com/package/pretty-quick) but its geared towards Prettier only. With lint-staged we can also run linter on files.

simison:
Once everything is in Prettier, we might be able to remove these Eslint rules from our config, since Prettier-eslint config [turns them `off` anyway](https://github.com/prettier/eslint-config-prettier/blob/5cf39be323b5bdd3208e5519cf539026edaf289d/index.js):

- `comma-dangle`
- `comma-spacing`
- `curly`
- `indent`
- `key-spacing`
- `keyword-spacing`
- `no-multi-spaces`
- `no-trailing-spaces`
- `object-curly-spacing`
- `one-var-declaration-per-line`
- `semi`
- `space-before-function-paren`
- `space-in-parens`
- `quote-props`
- `wrap-iife`
- `arrow-spacing`

For now, we need to keep them since not every file is Prettified.

nicksellen:
I actually don't mind mega PRs formatting change so much, I tend to find a way to run the prettier on my other PRs before trying to rebase/merge, so that it's more-or-less in the right shape anyway. Might be more fiddly getting the exact prettier config to use. But perhaps others are less happy doing PR fiddling, and so maybe your option 2 is best. I guess once we rebase our other PRs ontop, we could turn the `@format` thing on as we go...

package.json
nicksellen: Yes, seems good to be able to have it doing general linting too. Although we also have linting in `npm run lint:watch` _and_ in the webpack dev server... (for client files), linting everywhere... (not to mention my editor is doing it too).

Personally, I would not add the git hooks to my own local setup, I don't know why but they feel invasive, maybe I can learn to love them though. But not much unlinted makes it through the 3 lint stages above anyway.

simison: Hehe yeah. We should drop linting either here _or_ in webpack dev-server+lint:watch. I like the simplicity of this setup and since I have continuous linting in editor, I don't really need the dev-server linting. I appreciate it might be easier to approach for new devs/volunteers, tho.

What do you think?

In future I'd like to add Stylelint to lint *.less files so it would get added here, too.

simison:
I'm happy to do mega PR if you all are fine with it. I don't have that much to rebase as you all have. ;-)

mrkvon:
Rebased.

simison:
>And is it possible to format everything without adding the /** @format */ piece?

Yes, I indeed changed it to that in the latest commit but didn't update the PR description yet. 🤦‍♂  I initially added `format` pragma just because I didn't want to do one mega-PR and choose portions of the code to convert in smaller chunks.

package.json
```diff
@@ -34,6 +34,7 @@
     "migrate-down": "migrate down",
     "migrate": "migrate up",
     "postshrinkwrap": "node -e \"fs.utimesSync('./node_modules', new Date(), new Date());\"",
+    "reformat-files": "prettier --ignore-path .eslintignore --write",
```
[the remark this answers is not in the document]
simison: It asks for files or globs — so e.g. `npm run reformat-files modules/**/api/*`

There might be something else off since my last commit if it doesn't work standalone. 🤔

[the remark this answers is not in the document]
mrkvon: Well, I haven't tested with the glob parameter. Could that be the case?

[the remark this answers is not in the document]
simison: Should work with single file part too.

mrkvon:
@simison How do you execute the prettifying on all the files?

My experience is as follows:
- The `npm run reformat-files modules/**/*.js` works. :tada:
  However it prettifies only the files with `/** @format */`.
- When I make a commit, the hooks get executed (there is a list of tasks and _husky_ doing something) :tada: 
  During this process eslint complains `space-before-function-paren`; only on files with `/** @format */`. So it seems that during committing only files with `/** @format */` get prettified, then eslint is unhappy about those changes and then everything is reverted.

So if I'm not missing something, there are two issues to be addressed:
- everything should be prettified, even without the`/** @format */`
- the prettier and eslint shouldn't clash on `space-before-function-paren`

Then we'll be all set, we can make a big prettier commit and hopefully merge this. :)

mrkvon:
There is a [whole flamewar](https://github.com/prettier/prettier/issues/3847) on the topic of `space-before-function-paren`. It [got resolved in favor of `function ()`](https://github.com/prettier/prettier/pull/3903), but I haven't found which release will contain this. Possibly prettier 2.0, but I have hard time finding out.

package.json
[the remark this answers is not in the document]
mrkvon: > Should work with single file path too.

It does.

The remaining issue is: it still expects the `/** @format */`, otherwise nothing gets formatted.

mrkvon:
> everything should be prettified, even without the `/** @format */`

Is this going to be resolved as part of the big prettify commit?

.prettierrc
```diff
@@ -0,0 +1,8 @@
+singleQuote: true
+trailingComma: all
+jsxBracketSameLine: false
+insertPragma: true
+requirePragma: true
```
simison: This line needs to be removed.

simison:
Yup, we can squash everything here into one commit, then rebase and add another "reformat all the files" commit and quickly merge. and run. :-D

mrkvon:
Rebased.

@simison wrote:
> Yup, we can squash everything here into one commit, then rebase and add another "reformat all the files" commit and quickly merge. and run. :-D

@simison Do you wish to wait, or you don't have a capacity for this? If the latter is the case, I can do it.

simison:
Feel free to take over! I won't have free cycles before next weekend
probably.

package.json
```diff
@@ -248,5 +252,21 @@
   "bugs": {
     "url": "https://github.com/Trustroots/trustroots/issues"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "husky": {
```
mrkvon: @simison What do you think about moving `husky` and `lint-staged` config to `husky.config.js` and `lint-staged.config.js` respectively (or rc files) to keep package.json uncluttered?

If you think it's a good idea and if you don't have time for it, I will do it.

simison: Sure, I generally like that approach for configs as well.

Reason why I put them here initially was that they're so tied to actual `package.json` scripts. Configs themself were very "thin" — i.e. there wasn't much more to them than what commands to run — that having the `scripts` and these in the same file felt like easier to read.

Similar with `browserslist` config:

https://github.com/Trustroots/trustroots/blob/d07f7b66cebb4747d2a0a3ea7eac1749d7a98c59/package.json#L16-L18

But little by little we're having more of these in the `package.json` so it makes totally sense to move them to separate files. 👍  It also makes them easier to discover.

mrkvon: Makes sense. Then we can as well keep it in `package.json` and move it later. :slightly_smiling_face:

mrkvon:
I squashed the commits until now, removed some conflicting eslint rules and performed the prettier reformat-files on md, json, html and js files.

A few notes:
The original branch (before force-update) is saved at branch prettier-backup.
I made `Prettier <prettier@prettier.io>` to be the author of those formatting commits. I don't want to have my name everywhere to `git blame` and i don't want to mess up my coding statistics either.

Are we ready for such a large change?

Please, review the additional removed eslint rules efb6b8618ee91799dd8bdecbb12a32fe4bd1e70d, see that you're happy with the rest, too.

We'll still want to squash those additional eslint rules into the "configuration" commit and maybe squash together the reformatting commits as a single separate commit.

Please try to run the app and see that it still works as expected.

mrkvon:
We may want to undo reformatting of md files in `.github/` and `modules/core/server/views/email-templates-text/`, and possibly more.

simison:
Thanks, @mrkvon — on cursory looking Eslint config changes look good.

Should we just add  those .md files to [eslintignore](https://github.com/Trustroots/trustroots/blob/d07f7b66cebb4747d2a0a3ea7eac1749d7a98c59/.eslintignore) for now?

docs/Database.md
```diff
@@ -35,19 +36,23 @@ _TODO: Move to code using http://usejsdoc.org/ format._
 - **resetPasswordExpires** (date)
 
 ## References
-- **_id**
-- **[user](#users)** (_id)
+
+- **\_id**
```
simison: `\` before `_id` seems odd?

mrkvon: I suppose it has to do with `_` being a special character in markdown.

It seems that both
\_id (`\_id`)
and
_id (`_id`)
work fine with (github) markdown, but the first one removes some ambiguity.

I actually learned here. Didn't know we could escape underscores like that.

simison: Funky! At any rate, this doc file is old and outdated and should probably just be removed anyway. 😁

mrkvon: Shall we make a separate PR?

simison: Yes that's better for separate PR.

mrkvon:
I suppose the `git add` is necessary if we want the Prettier+commit work. (Haven't tested.)

We get warned rightfully. I find it scary indeed that there are changes which get committed and which a developer has no control over.

I wonder whether to squash the four formatting commits into one (md, html, js, json). I'm inclined to do it, but I'd like a second opinion. Probably doesn't matter anyways. :slightly_smiling_face:

mrkvon:
I've squashed the commits a bit, so there are 2 commits left: config and formatting. I think it would make sense to keep it that way and use the _Rebase and merge_ button.
I have it backed up locally, so if everything seems ok, @simison feel free to merge.
Otherwise I'll merge in 12 hours or so.

simison:
One huge reformat commit seems fine instead of splitting by format.. Might make rebasing easier?

I also thought precommit hook needs the git add but the warning made me think it might make it on our behalf?

simison:
The idea that commits Prettified files on your behalf is to provide convenience. :-) I have it set in editor myself, so it reformats as I type or safe files.

You'll still get chance to see the edits before pushing and then finally before merging.

mrkvon:
@simison Indeed, your understanding is correct. I've just tested that with the `git add` removed from `lint-staged` config, the prettier changes are still committed.

The change was included with the latest `push --force`

mrkvon:
With that done, let's merge.

Kudos @simison :tada: 

And let's see how rebasing after this will go...

PR #934, opened 2018-12-08
Implicit meet offer does not display through date

### LINKED ISSUE #923: Implicit meet offer does not display through date
**Describe the bug**
When you open new meet request it shows the Visible through field without the date. Only if you change the length to something else and back to month it works.

**To Reproduce**
1. Click on Meet button
2. Scroll to How long should this be visible?
3. No date is displayed (also if you click "A month")

**Expected behavior**
The date is displayed automatically.

**Screenshots**
image

FF 63.0.3 (Linux Mint)

topalavlad:
`setValidUntilDays()` is not called when first opening the "New meet" page because the default is already set

#### Proposed Changes

On the `activate()` function of the directive a call to `setValidUntilDays()` will be done if the date was not pre-populated

#### Testing Instructions

- Open `offer/meet/add` page. Check that the text under `How long should this be visible?` is not only `Visible through` but also contains a date until the meet will be visible.
- Try to modify an existing Meet and check that the date is still populated correctly.

Fixes #923

### DIFF
--- a/modules/offers/client/directives/tr-offer-valid-until.client.directive.js
+++ b/modules/offers/client/directives/tr-offer-valid-until.client.directive.js
@@ -66,6 +66,8 @@
         if ($scope.validUntil) {
           vm.isCalendarVisible = true;
           vm.validUntil = $scope.validUntil;
+        } else {
+          setValidUntilDays(vm.offerValidityInDays);
         }
 
         $scope.$watch('trOfferValidUntil.offerValidityInDays', function (newValue, oldValue) {


petr-hajek:
When testing I came across another existing issue: https://github.com/Trustroots/trustroots/issues/937
This fix is making it more visible. I'm trying to fix #937 first, before we merge this.

simison:
@petr-hajek @topalavlad  if the bug exists there with or without this PR, would we be okay merging this even before the fix?

petr-hajek:
@simison With this change it means the current time is used instead of end of the day when using the calendar to select dates because of setValidUntilDays does not set endOf('day')

simison:
>Changing Line 98: `   vm.validUntil = moment().endOf('day').add(add).toDate();`
Should fix all issues.

Seems reasonable 👍  I.e. here:

https://github.com/Trustroots/trustroots/blob/b29b6b9ebd45a969d531be16425e3fd55ead27cf/modules/offers/client/directives/tr-offer-valid-until.client.directive.js#L85-L99

petr-hajek:
I looked more deeply and it is improvement in all cases so I'll add the other fix as a separate PR

PR #935, opened 2018-12-08
[WIP] Manual for a migration from AngularJS to React

mrkvon:
#### Proposed Changes

* documentation

#### Testing Instructions

* [read it!](https://example.com)

Fixes #


mrkvon:
FYI @Akronix 

First draft. Work in progress.

docs/Development.md
```diff
@@ -23,6 +23,7 @@ We're trying waffle.io to get [an overview of issues](https://waffle.io/Trustroo
 # The application
 * **MEAN** stack, seeded originally with [MEAN.js](http://meanjs.org/) boilerplate: [MongoDB](www.mongodb.org), [ExpressJS](http://expressjs.com/), [AngularJS](https://angularjs.org/) v1, [NodeJS](http://nodejs.org/). Additionally stuff like [Bootstrap](http://getbootstrap.com/), [Leaflet](http://leafletjs.com/) etc. 
 * [[Database]] scheme (look for `*.server.model.js` project files to check most up to date info)
+* We're migrating the client to React. Read a [migration guide](React.md).
```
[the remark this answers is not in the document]
mrkvon: For me it works as it is. Both documents are within the docs folder.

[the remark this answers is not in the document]
simison: Oh right! Sorry, my bad. 🤦‍♂️ I looked at the file and thought I'm looking at `README.md` from the root folder.

docs/React.md
```diff
@@ -0,0 +1,30 @@
...
+## How to wire up React components to AngularJS?
+
+Create a react component with a path `modules/**/client/components/SomeName.component.js`. It will get imported to Angular automatically. Then you can use this component within Angular like `<some-name></some-name>`.
+See an [example component](https://github.com/Trustroots/trustroots/blob/master/modules/pages/client/components/Volunteering.component.js).
+
+## Issues and solutions
+
+### Routing
+
+We're still using AngularJS for routing.
```
simison: Noting that we're not using AngularJS's internal router, but a bit more capable UI-Router: https://ui-router.github.io/ng1/

docs/React.md
```diff
@@ -0,0 +1,30 @@
...
+
+We want to have separate codebases for server and client. The client will be written in React.
+
+We'll replace the current AngularJS components one by one. Eventually, we'll drop the AngularJS and create a React project. 
+
+### Rationale
+
+- React is more attractive to developers.
+- AngularJS is being replaced by newer frameworks.
+- [MEAN is not maintained](https://github.com/Trustroots/trustroots/issues/638).
```
simison: I'd add:

- React is simple and minimalistic; we won't risk getting stuck with a framework like  happened with AngularJS v1.

- React is not exotic; plenty of tooling and documentation available & lots of folks are already familiar with React.

- It has the potential to help with [React Native](https://github.com/Trustroots/trustroots-expo-mobile/) implementation of Trustroots. Not necessarily by re-using components but by allowing us to [_"learn once, write everywhere"_](https://www.youtube.com/watch?v=LIeqUPvh8qY).

mrkvon:
@simison What is the bigger picture? I imagine we may eventually (and relatively soon) have a separate codebase for React app and a separate codebase for API (i.e. two git repositories or two independent folders within a repository).
I.e. when we have enough Angular components migrated to React, we'd just create a new React project with `create-react-app` and use the components there; with React routing etc. and with no traces of AngularJS.
What other ways can you think of? And what would you prefer?

docs/React.md
mrkvon:
```

simison:
Good question!

> I imagine we may eventually (and relatively soon) have a separate codebase for React app and a separate codebase for API (i.e. two git repositories or two independent folders within a repository).

Yes — for the ease of setting things up and making changes to both simultaneously, it would be better to keep these in one mono-repo. How soon, it's hard to tell yet before we see how quickly we can migrate different parts. Some of the sections in Trustroots are pretty complicated; such as messages and especially the search.

> I.e. when we have enough Angular components migrated to React, we'd just create a new React project with `create-react-app` and use the components there; with React routing etc. and with no traces of AngularJS.

I could imagine these approaches:

1. We do one big rewrite-jump at some point when we have a bunch of React components, no React'y state management yet and AngularJS is still handling all the routing but providing mostly just wrapper views/templates for each page. Angular would probably still resolving data via routes for these views. Much like you described.

I don't have much experience from `create-react-app` to comment on the usefulness of that. What are your thoughts on that? I'd imagine it bootstraps some things such as build process and file structure (that both we already have).

I'd imagine the viability of that approach and amount of work required in the "big jump" would be clear only when we get closer to that point.

How do you imagine the point of the jump being?

2. Another route (_pun intended_) I'd imagine is to have a mixture of sections. For example, all and everything under "profile" section would be handled by React and some router for React but once you jump out to other routes we would again bootstrap the old Angular app. We would have mixtures of views like that and bootstrap the application again when navigating between sections. Perhaps we would find a way to keep both apps bootstrapped during the whole session.

-- 

Will be exciting to figure this all out!

PR #946, opened 2018-12-10
Add eod time to new meetup. Allow to make meetup for current day.

### LINKED ISSUE #937: Incorrect validUntil time in Meet
**Describe the bug**
When user adds new meetup it can have two incorrect times. In first case it is set at the end of the day before (or in case of using today, to one month later), in second case the date is correct, but the time is not the eod, but current time.

**To Reproduce**
1. Go to Meet
2. Add new Meetup
_Case 1 - day before_
3. Click at Other..
4. Select tomorrow
5. Save
6. Meetup is set to end today
In the db validUntil is set to the day before selected at 23:00:00.000Z 
In case you select today, it will create Meetup ending in one month (because the date would be in the past).

_Case 2 - incorrect time_
3. Click at one week/2w
4. Save
When you look it up in db, it has the correct date, but time of creation. The time stays even if you change the date.

**Expected behavior**
The date is saved in all cases with correct date and end of day.

FF 63.0.3 Mint

**Additional context**
setValidUntilDays does not use end of day

petr-hajek:
Fixes #937 
#### Proposed Changes

* Added end of day when creating new meetup
* Allow validUntil to be current day
* Fixed a typo

#### Testing Instructions

* Check if new meetups are created with correct time
* Check if modifying meetup is adding correct time

### DIFF
--- a/modules/offers/client/directives/tr-offer-valid-until.client.directive.js
+++ b/modules/offers/client/directives/tr-offer-valid-until.client.directive.js
@@ -38,7 +38,7 @@
       vm.offerValidityInDays = 30;
 
       // Predefined options visible without calendar
-      vm.choises = {
+      vm.choices = {
         'A week': 7,
         'Two weeks': 14,
         'Three weeks': 23,
@@ -95,7 +95,7 @@
         // @link https://momentjs.com/docs/#/manipulating/add/
         var add = days ? { days: days } : appSettings.limits.maxOfferValidFromNow;
 
-        vm.validUntil = moment().add(add).toDate();
+        vm.validUntil = moment().endOf('day').add(add).toDate();
       }
 
     }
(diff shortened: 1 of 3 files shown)


simison:
Looks like the same Moment.js `endOf` and `startOf` APIs could be used instead of `set(zeroTime)` in:

https://github.com/Trustroots/trustroots/blob/9670c9c1d6468d9008d5c2ffbce9cff97bf61596/modules/offers/server/controllers/offers.server.controller.js#L171

...but that's for another PR. :-)

petr-hajek:
> Looks like the same Moment.js `endOf` and `startOf` APIs could be used instead of `set(zeroTime)` trick in:
> 
> [trustroots/modules/offers/server/controllers/offers.server.controller.js](https://github.com/Trustroots/trustroots/blob/9670c9c1d6468d9008d5c2ffbce9cff97bf61596/modules/offers/server/controllers/offers.server.controller.js#L171)
> 
> Line 171 in [9670c9c](/Trustroots/trustroots/commit/9670c9c1d6468d9008d5c2ffbce9cff97bf61596)
>  .set(zeroTime); 
> 
> ...but that's for another PR. :-)

Yea, I was thinking about if to change it along this changes :)

simison:
> Yea, I was thinking about if to change it along this changes :)

I looove tiny, focused PRs. :-)

PR #949, opened 2018-12-10
Show Save&Exit only when editing existing Meetup

### LINKED ISSUE #924: Remove save&exit button and add cancel in Meet (desktop)
**Is your feature request related to a problem? Please describe.**
In the desktop view of Meet there is Save & exit button which can be pressed before the user finished the flow. At the same time it is missing a cancel button

**Describe the solution you'd like**
Replace the Save & Exit with Cancel button (so it will be same like in the mobile view).

petr-hajek:
Fixes #924 
#### Proposed Changes
All is for big-screen version
* Show Save&Exit only when editing existing Meetup
* added Cancel button to both add/edit

#### Testing Instructions

* Check if it feels like good UI

### DIFF
--- a/modules/offers/client/less/edit.less
+++ b/modules/offers/client/less/edit.less
@@ -166,3 +166,7 @@
     margin-top: 40px;
   }
 }
+
+.btn-inverse-primary {
+  margin-left: 5px;
+}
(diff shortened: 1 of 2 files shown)


petr-hajek:
@mrkvon That was not touched by this change. It is also in line with mobile views of other parts of TR

PR #954, opened 2018-12-11
UI: Make description required in the Host offer

### LINKED ISSUE #943: Make description required in hosting offer
**Is your feature request related to a problem? Please describe.**
Description in hosting offer (yes or maybe) should be required so we avoid the empty offers

petr-hajek:
Fixes #943
Similar to #942 changes

#### Proposed Changes

* Next, Save, Location buttons disabled until user fills at least 5 chars of description (if it is yes/maybe)
* Added tooltips for Next & Save when disabled

#### Testing Instructions

* Check if UI feels good and should work also for small screen

### DIFF
--- a/modules/offers/client/views/offer-host-edit.client.view.html
+++ b/modules/offers/client/views/offer-host-edit.client.view.html
@@ -23,7 +23,7 @@
           ng-disabled="offerHostEdit.isLoading ||
           (offerHostEdit.offer.status!=='no' && (offerHostEdit.offer.description|plainTextLength)<5)"
           uib-tooltip="Write longer description first"
-          tooltip-enable="(offerMeet.offer.description|plainTextLength)<5"
+          tooltip-enable="(offerMeet.offer.description|plainTextLength)<5 && offerHostEdit.offer.status !== 'no'"
           tooltip-placement="bottom"
           aria-hidden="true">
     Save and Exit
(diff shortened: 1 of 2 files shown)


modules/offers/client/views/offer-host-edit.client.view.html
```diff
@@ -20,7 +20,11 @@
 
   <button type="submit"
           class="btn btn-lg btn-inverse-primary pull-right hidden-xs"
-          ng-disabled="offerHostEdit.isLoading"
+          ng-disabled="offerHostEdit.isLoading ||
+          (offerHostEdit.offer.status!=='no' && (offerHostEdit.offer.description|plainTextLength)<5)"
```
simison: There's quite a bit of repetition here with `<5` — let's move it to a method in a controller to consolidate it more?

petr-hajek: I tried it, but I couldn't make it work with $scope.$watch

petr-hajek: People here in collective also don't know, so either you tell me or I leave it as it is :)

mrkvon: I stumbled upon a `$scope.$watch` that works. 83863861f265a6fa166e549c26865821d4826416
Still needs to get wired up with the other fields.
```js
$scope.$watch('offerHostEdit.offer.description', function (newValue) {
  console.log($filter('plainTextLength')(newValue));
});
```
I believe we were trying this together with @petr-hajek . I wonder why it didn't work.

Issue #826, opened 2018-11-11
MailDev: only open ports 1025 and 1080 on localhost


guaka:
**Describe the bug**
Currently MailDev opens up to the whole network. That's not good.

**To Reproduce**
1. install trustroots locally
2. `nmap localnetwork 1080` from another machine

**Expected behavior**
1080 and 1025 should not be open to the outside

**How it should behave**

MailDev should only be open on localhost, not on external network interfaces.

simison:
How should it behave instead? 🙃 

Remember that it's just a developer tool meant to run only while developing the site.

guaka:
** how it should behave **

MailDev should only be open on localhost, not on external network interfaces.

Not extremely hypothetical:
- work on trustroots on hacker conference such as CCC
- people are portscanning everything
- people find a security issue with MailDev and hack your machine

simison:
Right, how can this be done? Maildev listens locally just like Nodemon and the rest of the development setup.

I feel like limiting access should probably be handled locally on the computer by your firewall. 🤔

guaka:
Didn't see anything at https://github.com/djfarrelly/MailDev/issues - I may post something there

guaka:
## macOS 10.14

I added these lines to my `/etc/pf.conf`:
```
block in on en0 proto tcp to port 1080
block in on en0 proto tcp to port 1025
```

then did `sudo pfctl -f /etc/pf.conf` to reload the configuration.

This is probably enough, but I'd still prefer to have this handled in our codebase (and/or on MailDev's side) so that it will never be an issue for any of our developers.

zrthstr:
>> Right, how can this be done? 
Maildev, according to the doc can be told to bind to the local interface (lo) by telling it to bind to 127.0.0.1
This would solve the issue.. 

from: https://danfarrelly.nyc/MailDev/
--ip <ip address<               IP Address to bind SMTP service to

So adding --ip 127.0.0.1 should love this problem.

PR #955, opened 2018-12-11
Add to docs how to access server from another device

petr-hajek:
#### Proposed Changes
I had some fun time figuring this out so maybe could be useful to someone else. Good to test on a mobile phone.


guaka:
Kinda the opposite of #826 :)

PR #957, opened 2018-12-11
Interface for selecting an app language

mrkvon:
Follow up of #921 

- [x] visitor can select a language in a header
  - the selected language will be saved in a cookie
- [x] user can select a language in account page
  - [x] the selected language will be saved in the database
  - [ ] the saved language will be loaded to i18n on signin or page reload #1054 
- [x] design of buttons (see [a proposed design](https://github.com/Trustroots/trustroots/pull/921#issuecomment-446142245))
  - [x] the language switch button in header needs positioning and some other details. i can't do that.

- API was updated in #962

### DIFF
--- a/modules/core/client/components/LanguageSwitchPresentational.js
+++ b/modules/core/client/components/LanguageSwitchPresentational.js
@@ -10,8 +10,6 @@ import locales from '@/config/shared/locales';
  * @param {Function} onChangeLanguage - what to do when language is changed
  */
 export function LanguageSwitchDropdown({ currentLanguageCode, onChangeLanguage }) {
-  // languages which are not selected
-  const otherLanguages = locales.filter(language => language.code !== currentLanguageCode);
   // selected language
   const currentLanguage = locales.find(language => language.code === currentLanguageCode);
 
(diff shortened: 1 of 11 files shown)


mrkvon:
Since we don't have a centralized place in React where we could load the user's language, this is as far as I can go now.

Added positioning. Ready for review and/or merge.

I disabled the display of dropdown-menu on hover for the language switch. Before it was disappearing because of a space between the button and the menu.

modules/core/client/components/LanguageSwitchPresentational.js
```diff
@@ -0,0 +1,66 @@
...
+      title={currentLanguage.name}
+      id={`dropdown-basic-${currentLanguage.code}`}
+      pullRight
+      className="disable-hover-display"
+    >
+      <Dropdown.Toggle className="btn-inverse">
+        {currentLanguage.name}
+      </Dropdown.Toggle>
+      <Dropdown.Menu>
+        {otherLanguages.map(language => (
```
simison: I think we could simplify this and list all the languages regardless, and just mark list item that matches `currentLanguage.code` with `active` -prop and it'll get highlighted.

It'll be double in toggle and in the list but it's okay, user can orient themself easier with list where items don't dissapear. (and code is simpler, yay!)

simison:
Tried rebasing this but I don't think I did such a good job at it, so I'll leave it to ya. :-) 

This seems ready code-wise, but I'd be happy to give this visual inspection.

mrkvon:
rebased, still some pending comments

modules/core/client/components/LanguageSwitchPresentational.js
mrkvon: 247f92c88dce87906516c9a1521b2e624ab94d06

selector

mrkvon:
Resolved.

The visual inspection would be nice.

Look into (top header for non-logged user) and to (user account). Both contain different-looking language switches (and the account switch additionally saves the user's locale to database (even though we don't use it, yet)).

PR #962, opened 2018-12-13
Save user's locale to database (API)

mrkvon:
#### Proposed Changes

* add `locale` to user model via `PUT /api/users` with body `{ locale: 'eng', ... }`

#### Testing Instructions

* `npm run test:server:watch`

Continued work on #492

### DIFF
--- a/modules/users/server/controllers/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users.profile.server.controller.js
@@ -1,5 +1,8 @@
 'use strict';
 
+var path = require('path'),
+    languages = require(path.resolve('./config/languages/languages'));
+
 /**
  * Module dependencies.
  */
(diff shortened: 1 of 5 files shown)


modules/users/server/controllers/users.profile.server.controller.js
```diff
@@ -299,6 +302,12 @@ exports.update = function (req, res) {
     });
   }
 
+  // validate locale
+  // @TODO validation framework
+  if (req.body.locale && (typeof req.body.locale !== 'string' || !Object.keys(languages).includes(req.body.locale))) {
+    return res.status(400).end();
```
simison: Just to keep things consistent with the rest of the API, let's return:

```js
return res.status(400).send({
  message: errorService.getErrorMessageByKey('bad-request')
});
```

mrkvon: 2f809c6

Issue #944, opened 2018-12-09
Improve Search User API


mrkvon:
Follow up of #531.

Improve /api/users?search=string

- [x] pagination
- [ ] expand the user profile fields returned, or allow user to specify which fields to return
- [x] ~~improve search via usernames; figure out how dots, dashes etc. are treated by text index~~
_it is working fine for periods, hyphen and underscores. If you put period or hyphen into the search it works the same like a space_

mrkvon:
ping @Akronix 

Feel free to update the issue with further requirements.

Akronix:
How about fuzzy search?
Searching for "Abe" should return all instances of Abel, one of them myself: Abel Por Qué No.
And better, searching for "Avel" (similar names) should find "Abel" as well...

simison:
>Searching for "Abe" should return all instances of Abel

So this was possible with MongoDB regexp queries but the issue was that they weren't relying on indices and thus very slow — is that correct @mrkvon ?

I wonder if we could do 2nd expensive query with regexp if first one doesn't yeld results?

>And better, searching for "Avel" (similar names) should find "Abel" as well...

I agree that's how it should work ideally. Great if it's easy and performant enough with MongoDB Indices but I suspect it won't be either. :-( Investing time and maintenance energy in something like ElasticSearch at this point probably isn't worth the effort on the other hand?

guaka:
Sounds like premature scaling.
How slow is a regexp search really?
How many searches do we expect on a daily basis?

simison:
Yep, that's a valid point. If someone whips up a PR I'm happy to help to measure with production data (~34K users).

guaka:
The admin search I've been working on works pretty well with 34k users -
it's quite crude but does the job well.
https://github.com/Trustroots/trustroots/pull/960

mrkvon:
> potentially very slow

I believe the main issue is being a potential vector of DoS attack. Not the slowness on its own.
So, can we limit amount of requests per second for a specific url + method?

Also, unlike a search using indices, this will get worse as the amount of members grows.

simison:
>I believe the main issue is being a potential vector of DoS attack. Not the slowness on its own.
So, can we limit amount of requests from one user per second for a specific API endpoint?

Yes, we already do that for the whole `/api` but adding a separate more strict "bucket" for specific API endpoints is very straightforward.

It's configured in Nginx.

mrkvon:
[...]

PR #965, opened 2018-12-13
Search users

### LINKED ISSUE #530: Find contact by typing in username or name

Akronix:
#### Proposed Changes

* Add a new /search/users endpoint to look for users by username. 
* Coded UsersList, Avatar and SearchUsers components in React.
* Added a button to switch from map search to user search and vice versa.

First approach to resolve #530

#### Testing Instructions

1. Run the dev server and go to http://localhost:3000/search/users.
2. Write an existing username, like admin1 and press the search button.

#### Current status
Screenshot of current status:
image

#### Still needs to be done

- [x] Show profile pics
- [ ] Show other info like living location, language, gender, last seen,.. Depends on #944
- [x] Pressing Enter key, do submit search.
- [ ] Fuzzy search. Depends on #944
- [x] Check they are well translated and remove angular left-overs.
- [ ] add tooltips (?)
- [ ] Use `search?=` http parameter to init defaultProps.
- [ ] Add a link from the /profile/<username>/contacts page to /search/users
- [x] Final: remove console.logs and prepopulated `admin1`


modules/search/client/components/UsersList.js
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const Avatar = ({ avatarSize, avatarSource, userData }) => (
+const Avatar = ({ avatarSize, avatarSource, displayName, profileUrl }) => (
```
simison: Let's start this as an external component so that it becomes reusable (we'll literally need it everywhere from now on).

Our previous client-side Avatar directive is terrible; we should clearly move all that logic to the backend:

https://github.com/Trustroots/trustroots/blob/bbacc11d05923134a8fc140472473b43ded17716/modules/users/client/directives/tr-avatar.client.directive.js

guaka:
When I go to http://localhost:3000/search/users the page keeps reloading every second or so.

petr-hajek:
I think to be releasable we need to do just these:
- [ ] Show profile pics + remove non-implemented info
- [ ] Pressing Enter key, do submit search.
- [ ] Final: remove console.logs and prepulated `admin1`
Rest can be done in later iterations.

I am mostly thinking if the button to get to the search is intuitive. I think I wouldn't notice as regular user. Maybe we could add a text link somewhere else also?

Akronix:
> I think to be releasable we need to do just these:
> 
>     * [ ]  Show profile pics + remove non-implemented info
> 
>     * [ ]  Pressing Enter key, do submit search.
> 
>     * [ ]  Final: remove console.logs and prepulated `admin1`
>       Rest can be done in later iterations.
> 

Agree!

> 
> I am mostly thinking if the button to get to the search is intuitive. I think I wouldn't notice as regular user. Maybe we could add a text link somewhere else also?

Yes, I forgot to say! There could be a link from the contacts page.

simison:
>* [ ]  Show profile pics + remove non-implemented info

Profile images might require a bit more work to work so I'd be okay to hide it for the first iteration.

Akronix:
> > * [ ]   Show profile pics + remove non-implemented info
> 
> Profile images might require a bit more work to work so I'd be okay to hide it for the first iteration.

Really? But I think it is the most important info to display for a search users feature.

I might have some time today to do this:

>  [ ]  Pressing Enter key, do submit search.

simison:
> >   Profile images might require a bit more work to work so I'd be okay to hide it for the first iteration.

>Really? But I think it is the most important info to display for a search users feature.

I think so too and we definitely need to add it eventually. I meant that since it's quite some work either at client side or at the backend to make all the possible avatar sources work (none/uploaded/gravatar/facebook), it would make sense to drop it from the first version just to get something already functional published.

Creating an avatar component in React is pretty high on priorities though, since it's needed literally everywhere.

modules/search/client/components/UsersList.js
```diff
@@ -0,0 +1,103 @@
...
+};
+
+Avatar.propTypes = {
+  avatarSize: PropTypes.number,
+  avatarSource: PropTypes.string,
+  displayName: PropTypes.string,
+  profileUrl: PropTypes.string
+};
+
+const User = ({ userData }) => {
```
simison: Would be nice to extract this to an external reusable component (we could reuse it e.g. for the contacts list). That's okay for another PR too.

modules/search/client/components/UsersList.js
```diff
@@ -0,0 +1,103 @@
...
+};
+
+User.propTypes = {
+  userData: PropTypes.object
+};
+
+
+export default function UsersList({ users }) {
+  const usersEvenComponent = [];
+  const usersUnEvenComponent = [];
```
simison: We can pretty easily achieve this layout by using CSS flex or CSS grid instead of using Bootstrap columns — but that can be another PR as well.

modules/search/client/components/search.api.js
```diff
@@ -0,0 +1,10 @@
+export async function searchUsers(query) {
+  const response = await fetch(`/api/users?search=${query}`, {
```
simison: We can now rebase this PR and switch to use `axios` instead of `fetch`, see docs:

https://github.com/Trustroots/trustroots/blob/master/docs/React.md#api-calls-from-react

Akronix: Is this service already anywhere or is it going to be anywhere anytime soon?

Akronix:
> > > Profile images might require a bit more work to work so I'd be okay to hide it for the first iteration.
> 
> > Really? But I think it is the most important info to display for a search users feature.
> 
> I think so too and we definitely need to add it eventually. I meant that since it's quite some work either at client side or at the backend to make all the possible avatar sources work (none/uploaded/gravatar/facebook), it would make sense to drop it from the first version just to get something already functional published. Lack of avatar isn't blocking anyone from using this feature.
> 
> Creating an avatar component in React is pretty high on priorities though, since it's needed literally everywhere.

OK, then. I can remove all the avatar stuff, but somebody should be responsible to take that task and start coding it. 

Maybe that person wants to start from my code?

simison:
> Maybe that person wants to start from my code?

Yep I think this section is great to get it started but separate PR would be better; this one is quite big already.

simison:
Opened an issue for avatar: https://github.com/Trustroots/trustroots/issues/982

modules/search/client/components/search.api.js
simison: I don't think was added yet. Axios library is in already tho.

simison: Adding this in https://github.com/Trustroots/trustroots/pull/984

modules/search/client/components/UsersList.js
mrkvon: See #997

simison: https://github.com/Trustroots/trustroots/pull/997 is now merged and this can be changed to use new Avatar component instead. 💪

modules/search/client/components/SearchUsers.component.js
```diff
@@ -91,8 +92,10 @@ class SearchUsers extends React.Component {
     this.setState({ resolved: false });
     console.log('Searching for: ' + this.state.searchQuery);
 
+    // eslint-disable-next-line react/no-access-state-in-setstate
     const userResults = await searchUsers(this.state.searchQuery);
```
Akronix: I'm getting this eslint error in that line:
`95:43  error  Use callback in setState when referencing the previous state  react/no-access-state-in-setstate`
I have no idea on how to fix it. Any idea @simison @mrkvon ?
This is the eslint docs about it: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md

mrkvon: This is quite peculiar.
This rule takes care specifically of:
```jsx
// incorrect, can lead to state inconsistency
this.setState({ isFoo: !this.state.isFoo }); // flipping a random boolean

// correct
this.setState((previousState) => ({ isFoo: !previousState.isFoo }));
```
@Akronix Your link is describing the reasons for this.

Perhaps try
```
this.setState(() => ({ ... }))
```
but it shouldn't be necessary.

In a while, if not resolved, I can try locally.

modules/search/client/components/SearchUsers.component.js
```diff
@@ -0,0 +1,182 @@
...
+      'borderLeftStyle': 'solid',
+      'borderColor': '#eee' };
+
+    const switchToSearchPlaces = (
+      <span className="input-group-btn"
+        style={borderStyle}>
+        <a type="button"
+          href="/search"
+          className="btn btn-lg btn-default"
+          tooltip-enable="::search.screenWidth >= 768"
```
simison: A little bit of Angular tooltip directive here that can be removed.

modules/search/client/components/UsersList.js
simison: Done in https://github.com/Trustroots/trustroots/pull/1009

modules/search/client/components/SearchUsers.component.js
```diff
@@ -0,0 +1,182 @@
...
+    this.setState({ resolved: true, users: userResults.data });
+    console.log(this.state.users);
+  };
+
+  clickClear() {
+    this.setState({ searchQuery: '' });
+  }
+
+  render() {
+    console.log('SearchUsers: render');
```
simison: Let's remember to remove `console.log`s from the PR before merging.

modules/search/client/components/SearchUsers.component.js
```diff
@@ -0,0 +1,182 @@
...
+
+    const borderStyle = { 'borderLeftWidth': '1px',
+      'borderLeftStyle': 'solid',
+      'borderColor': '#eee' };
```
simison: Let's remember to move this style object to Less file with other our styles.

modules/search/client/components/SearchUsers.component.js
Akronix: sure :+1:

mrkvon: [eslint's `no-console`](https://eslint.org/docs/rules/no-console) can be interesting for us.

Just disable it explicitly whenever we really need it with `eslint-disable-line`.

modules/search/client/components/SearchUsers.component.js
Akronix: Well, I left all the angular stuff just to make sure they were being correctly translated to react (at least, the basic ones), but I can remove them very easily.

modules/search/client/components/SearchUsers.component.js
Akronix: hum, something to add to the todo list.

modules/search/client/components/SearchUsers.component.js
mrkvon: Still there? For me `npm run lint` passes.

mrkvon: Two lines below, use `this.setState(() => ({ resolved: true, users: userResults.data }))` instead. `const userResults = ...` references `this.state` and that's what this rule is worried about.

I haven't wrapped my head around this particular example, but at least the lint error disappears! :slightly_smiling_face:

simison: I think this section should be structured like this for it to work as expected and make the linter happy (untested):

```js
this.setState({ resolved: false }, async () => {
  searchUsers(this.state.searchQuery)
    .then((data) => {
      // Do something in success
    })
    .catch((error) => {
      // Do something in errors
    })
    .finally(() => {
      this.setState({
        resolved: true,
      });
    });
});
```

simison: That said, perhaps kinda `resolved` (I can regognice Angularism) could be renamed to `isQuerying`, `isSearching` or similar?

modules/search/client/components/SearchUsers.component.js
simison: Valid point @mrkvon ! 👍 :shipit:

simison: @mrkvon added the rule in https://github.com/Trustroots/trustroots/pull/1015

modules/search/client/components/SearchUsers.component.js
Akronix: @simison I'm thinking to better leave the tooltip directives just in case we got something similar working in the future for React? At the end, they don't really annoy so much, do they?

Akronix:
We got the basic points done:

- [x] Show profile pics
- [x] Pressing Enter key, do submit search.
- [x] Check they are well translated and remove angular left-overs.
- [x] Final: remove console.logs and prepopulated `admin1`

This is ready to review from my side :tada:

modules/search/client/components/SearchUsers.component.js
simison: It's really better to remove anything that's not used. 🙂 It can be confusing to have dead code around. If we create tooltips in future (probably not as they're not great user interface) it'll look different to Angular directives so they won't work "out of the box" with this syntax anyway.

modules/search/client/components/SearchUsers.component.js
```diff
@@ -0,0 +1,164 @@
...
+    <div>
+      {!isSearching?
+        <div className="content-wait"
+          role="alertdialog"
+          aria-busy="true"
+          aria-live="assertive">
+          <small>Wait a moment...</small>
+        </div>
+        :
+        (users && users.length > 0? userList : noUsers)
```
simison: I really like this variable -way of rendering components. 👍

modules/search/client/components/SearchUsers.component.js
```diff
@@ -0,0 +1,164 @@
...
+          </h4>
+        </div>
+      </div>
+    </div>
+  );
+
+  const noUsers = (
+    <div className="row content-empty">
+      <i className="icon-3x icon-users"></i>
+      <h4>No users found with this name.</h4>
```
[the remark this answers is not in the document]
Akronix: Done!

modules/search/client/views/search-input.client.view.html
```diff
@@ -1,6 +1,18 @@
...
+      <a type="button"
+        href="/search/users"
+        class="btn btn-lg btn-default"
+        tooltip-enable="::search.screenWidth >= 768"
+        tooltip-placement="bottom"
+        uib-tooltip="Go to search users"
+        aria-label="Go to search users by username">
+        <i class="icon-users"></i>
+      </a>
+    </span>
```
simison: I feel like we could take a little different route with discoverability/linking map search and user search to each other.

While these function well I think we can improve them to work around these drawbacks:

- It's not clear what happens from the button or that it even _is_ a button:

screenshot 2018-12-30 at 20 41 14
screenshot 2018-12-30 at 20 41 24

- It looks like it's part of the search input and there to provide visual symbol for thes earch rather than it's a way to access completely different major interface.

The tooltip helps little but we cannot rely on that; especially on mobile where we don't have hover-state. Seems like it has issues also on the desktop:

screenshot 2018-12-30 at 20 41 32

We can assume most folks not seeing these tooltips so we need to build UI that's clear without them and hide only additional, optional information in tooltips.

The intent at the user search side is little bit more clearer but I'd say it has similar issues and it makes the search input rather busy with all the icons:

image

I would suggest to hide these UIs for now and e.g. add link to this feature only behind user-dropdown. That's not very discoverable of course so we'll need to work more on it, but we can have different PR for that so that we don't slow down merging this.

I know you're AFK right now; hope you don't mind if I go ahead and do those PRs and we can still add these inputs in some format if we so conclude! :-)

simison:
Aight, so I:
- removed icon-cross links from map search to here; I'll add them back in a separate PR in slightly different format
- Restructured just a little bit `isLoading` state
- Removed "admin1" demo query

modules/search/client/views/search-input.client.view.html
simison: Work in progress here: https://github.com/Trustroots/trustroots/pull/1020

modules/search/client/config/search.client.routes.js
```diff
@@ -80,6 +80,15 @@
         data: {
           pageTitle: 'Search'
         }
+      }).
+      state('search-users', {
```
Akronix: @simison  I find this search-users /search/members inconsistent. Shall we move everything to members naming convention?

PR #968, opened 2018-12-13
Add Search users pagination

### LINKED ISSUE #944: Improve Search User API

petr-hajek:
Partially deals with #944 
#### Proposed Changes

* Added pagination to search users API
* Updated tests and added checks for this
* Limit for one page is configurable

#### Testing Instructions

* call `GET /api/users?page=2&search=xxx`
* calling without (or with invalid) page parameter should return the first page
* also request `GET /api/users?page=2&limit=10&search=xxx` should be supported

### DIFF
--- a/modules/users/server/controllers/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users.profile.server.controller.js
@@ -1496,6 +1496,8 @@ exports.search = function (req, res, next) {
     });
   }
 
+
+
   // perform the search
   User
     .find({ $and: [
@@ -1510,7 +1512,9 @@ exports.search = function (req, res, next) {
     .select(exports.userSearchProfileFields)
     .sort({ score: { $meta: 'textScore' } })
     // limit the amount of found users (config)
-    .limit(config.limits.userSearchLimit)
+    .limit(config.limits.userSearchPageLimit)
+    // skip to the page, automatically handles invalid page number
+    .skip((req.query.page-1)*config.limits.userSearchPageLimit)
     .exec(function (err, users) {
       if (err) return next(err);
       return res.send(users);
--- a/modules/users/server/controllers/users.profile.server.controller.js
+++ b/modules/users/server/controllers/users.profile.server.controller.js
@@ -1496,8 +1496,6 @@ exports.search = function (req, res, next) {
     });
   }
 
-
-
   // perform the search
   User
     .find({ $and: [
@@ -1511,10 +1509,10 @@ exports.search = function (req, res, next) {
     // select only the right profile properties
     .select(exports.userSearchProfileFields)
     .sort({ score: { $meta: 'textScore' } })
-    // limit the amount of found users (config)
-    .limit(config.limits.userSearchPageLimit)
+    // limit the amount of found users
+    .limit(req.query.limit)
     // skip to the page, automatically handles invalid page number
-    .skip((req.query.page-1)*config.limits.userSearchPageLimit)
+    .skip(req.skip)
     .exec(function (err, users) {
       if (err) return next(err);
       return res.send(users);
--- a/modules/users/tests/server/search-users.server.routes.tests.js
+++ b/modules/users/tests/server/search-users.server.routes.tests.js
@@ -38,7 +38,7 @@ describe('Search users: GET /users?search=string', function () {
   });
 
   beforeEach(function () {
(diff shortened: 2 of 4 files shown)


modules/users/server/controllers/users.profile.server.controller.js
```diff
@@ -1503,7 +1505,9 @@ exports.search = function (req, res, next) {
     .select(exports.userMiniProfileFields + ' -_id')
     .sort({ score: { $meta: 'textScore' } })
     // limit the amount of found users (config)
-    .limit(config.limits.userSearchLimit)
+    .limit(config.limits.userSearchPageLimit)
+    // skip to the page, automatically handles invalid page number
+    .skip((req.query.page-1)*config.limits.userSearchPageLimit)
```
mrkvon: I'm still quite curious to understand the behavior of this thing.
- `req.query.page` is a string. can be parsed like `+req.query.page`.
- `req.query.page` can benefit from a validation (only positive integer allowed).

modules/users/tests/server/search-users.server.routes.tests.js
```diff
@@ -315,44 +314,57 @@ describe('Search users: GET /users?search=string', function () {
...
+          });
+        });
+        // TCs with different or missing page parameter
+        var pageTests = [
+          { params: '', expected: limit },
+          { params: '&page=1', expected: limit },
+          { params: '&page=2', expected: testUsers.length - limit }
+        ];
+
+        pageTests.forEach(function (test) {
```
mrkvon: I enjoyed learning the new technique for testing. Also :+1: refactor.

modules/users/server/controllers/users.profile.server.controller.js
petr-hajek: `+req.query.page` evaluates to 1 when it is xxx, 0 or negative. Same with `parseInt(req.query.page)`
I don't really know what to do with that

simison: We actually have a pagination middleware helper; take a look at threads and messages apis.

simison: Here:

https://github.com/Trustroots/trustroots/blob/f6893632a696d56ecfe64696313ac025112e0169/modules/messages/server/models/thread.server.model.js#L39

https://github.com/Trustroots/trustroots/blob/master/modules/messages/server/controllers/messages.server.controller.js#L145

simison: Actually thinking further:

- [Mongoose Paginate](https://github.com/edwardhotchkiss/mongoose-paginate) module that we use isn't maintained anymore

- We still don't want to replicate Pagination headers, query validation etc etc in every API endpoint, so we'll keep using well maintained [Express.js paginate](https://github.com/expressjs/express-paginate)

- We can just do the actual limit/pagination just like you do in this PR (see also [this example](https://github.com/expressjs/express-paginate#example-with-mongoose-odm-see-example-2-for-sequelize-orm)) — we should just ensure we rely on Express pagination so that we don't have one API endpoint outside it while the rest are using it.

mrkvon: So the express-paginate can take care of a validation for us, right? Or is there more in it?

mrkvon: @petr-hajek It was probably the `express-paginate` which was doing weird things with our `req.query.page`, setting it to 1. (Remember the confusion when we talked about it in person?)

It is set up in `/config/lib/express.js` and was doing funny validation stuff without our knowledge.

mrkvon: This is resolved with bba47c14d6b48b008d6b712.
We use `mongoose-paginate` with the knowledge that we use it.

mrkvon:
I believe this is ready for merge. Since I made the last commit, an external review might still be useful.

There is still some ambiguity with sorting search results (what happens when users have the same score?), but I can live with that now.

simison:
Thanks @mrkvon  & @petr-hajek !

PR #980, opened 2018-12-14
Translate components of ReferencesNew

mrkvon:
#### Proposed Changes

* translating create-references-react #827 
* fixing and refactoring tests (figuring out how to [test components with translation](https://react.i18next.com/misc/testing))
* additional refactoring

#### Testing Instructions

* in your `config/env/local.js` enable references and i18n
  ```js
  {
    ...,
    featureFlags: {
      i18n: true,
      reference: true
    },
    ...
  }
  ```
* `npm start`
* go to `localhost:3000/profile/{username}/references/new`
* switch the language button in header from `eng` to other language
* observe the changes
* _Note that most of the translations are not provided. Even after switching the language, most texts will remain in English._

To get the translations working, we'll need...
* a tool which extracts the keys to `/public/locales/{lng}/reference.json`
* to translate

Follow up of #827

### DIFF
--- a/config/lib/i18n.js
+++ b/config/lib/i18n.js
@@ -41,6 +41,7 @@ i18n
       escapeValue: false, // react already safes from xss
       format
     }
+    // debug: true // show missing translation keys in console.log
   });
 
 export default i18n;
--- a/modules/references/client/components/create-reference/Info.js
+++ b/modules/references/client/components/create-reference/Info.js
@@ -1,5 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import '@/config/lib/i18n';
+import { withNamespaces } from 'react-i18next';
 
 // @TODO provide the value from API config endpoint
 const timeToReplyReference = { days: 14 };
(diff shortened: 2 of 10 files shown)


modules/references/client/components/create-reference/Info.js
```diff
@@ -19,24 +21,24 @@ UserLink.propTypes = {
 /**
  * Error message when trying to give a reference to oneself.
  */
-export function ReferenceToSelfInfo() {
-  return (<div className="alert alert-warning">Sorry, you can&apos;t give a reference to yourself.</div>);
-}
+export const ReferenceToSelfInfo = withNamespaces('reference')(function ({ t }) {
+  return (<div className="alert alert-warning">{t('Sorry, you can\'t give a reference to yourself.')}</div>);
```
simison: I wonder if `\'` will end up to strings and if we could use `"` here:
```js
t("Sorry, you can't give a reference to yourself.")
```

modules/references/client/components/create-reference/Info.js
```diff
@@ -19,24 +21,24 @@ UserLink.propTypes = {
...
+});
 
 /**
  * Info that data are loading.
  */
-export function LoadingInfo() {
-  return (<div>Wait a moment...</div>);
-}
+export const LoadingInfo = withNamespaces('reference')(function ({ t }) {
+  return (<div className="alert alert-warning">{t('Wait a moment...')}</div>);
```
simison: https://github.com/Trustroots/trustroots/pull/1009#discussion_r243779493 ;-)

config/lib/i18n.js
```diff
@@ -41,6 +41,7 @@ i18n
       escapeValue: false, // react already safes from xss
       format
     }
+    // debug: true // show missing translation keys in console.log
```
simison: In another separate PR we can add this to config files so that it's easy to turn on/off locally.

simison: ...actually, take that back; we can't pull backend config files to frontend. :-)

modules/references/client/components/create-reference/Info.js
mrkvon: This will produce a linter warning:

```
warning  Strings must use singlequote  quotes
```

It seems that the `\` doesn't become a part of the translation keys. (At least they're logged without `\` when translation debugger is on.)

I'd prefer to leave this as it is for now (more consistency in quotes, personal bias) and if it ever causes issues, it can be replaced.

config/lib/i18n.js
mrkvon: If we find a way to do this, it can be helpful. I guess a separate client-only config can do it. API doesn't care about this.

PR #981, opened 2018-12-15
Configs: remove process.env

simison:
#### Proposed Changes

Some configs cleanup:
- Remove `process.env.*` variables since they're not used anywhere and just add noise (apart from `process.env.DB_1_PORT_27017_TCP_ADDR` which I think is used in Travis)
- Remove `domain` from production config as it's provided via `local.js` in production and should otherwise default to localhost.
- Add an example of how to override `featureFlags` in `local.sample.js`
- Remove redundant comments around featureFlags and put them on top of the file for clarity
- Add i18n feature flag to the development config
- ES6

#### Testing Instructions

The app should run as usual.

### DIFF
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -1,5 +1,3 @@
-'use strict';
-
 /*
  * Please don't make your own config changes to this file!
  * Copy local.sample.js to local.js and make your changes there. Thanks.
@@ -11,10 +9,9 @@
  */
 
 module.exports = {
-  // Feature flags
   featureFlags: {
-    // enable references?
-    reference: true
+    reference: true,
+    i18n: false
   },
   db: {
     uri: 'mongodb://' + (process.env.DB_1_PORT_27017_TCP_ADDR || 'localhost') + '/trustroots-dev',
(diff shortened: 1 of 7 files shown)


config/env/development.js
```diff
@@ -11,10 +9,9 @@
  */
 
 module.exports = {
-  // Feature flags
   featureFlags: {
-    // enable references?
-    reference: true
+    reference: true,
+    i18n: false
```
simison: @mrkvon there wasn't any specific reason `i18n` flag was missing from development config, right?

mrkvon: Probably not. It was probably supposed to use the default (whatever that was).

simison: It was also falsy in at default config but the thing is that we don't deep merge these configs, just shallow merge, so whole `featureFlags`  object here in development config will replace the one in default config.

mrkvon: Ah. Ok. Probably good to remember that.

simison: Yeah, there were reasons for it (not to have surprise values cropping up from default configs) but it might be surprising.

Also good for clarity to keep the same flags visible on each config so that folks don't need to go opening files or "think abstract" to understand the flag-system. We probably would need them just in production+development and not at all in default.

PR #983, opened 2018-12-16
Eslint: dynamicly load React version

simison:
#### Proposed Changes

- Load React version in `eslint` config dynamically. Fewer numbers to keep in sync.
- Move `react` and related modules from `devDependencies` to`dependencies`
- While at it, `npm` alphabetized dependencies

(Split out from bigger linting refactor PR https://github.com/Trustroots/trustroots/pull/932)

#### Testing Instructions

Does eslinting React files still work as expected?

In `package.json` React version is range semver `~16.6.0` instead of static `16.6` and if that causes trouble, I'll re-format it with [`semver` module](https://www.npmjs.com/package/semver)


simison:
Looking at the source for parsing this:

https://github.com/yannickcr/eslint-plugin-react/blob/8ef86c53003d78c88416a35e1e627914fcfabdf4/lib/util/version.js#L27-L47

...I don't think `~16` would work:
```js
const onfVer = "~16.6.0";
confVer = /^[0-9]+\.[0-9]+$/.test(confVer) ? `${confVer}.0` : confVer;
confVer.split('.').map(part => Number(part));

// => Array(3) [ NaN, 6, 0 ]
```

.eslintrc.js
```diff
@@ -1,3 +1,8 @@
+const semver = require('semver');
+
+// Converts semver range `~16.6.0` to strict version `16.6.0`
+const reactVersion = semver.coerce(require('./package.json').dependencies.react).raw;
```
mrkvon: I wonder what is the difference between `semver.coerce().raw` and `semver.coerce().version`.

package.json
```diff
@@ -80,9 +79,10 @@
     "angular-ui-bootstrap": "~2.5.0",
     "angular-ui-router": "~0.3.2",
     "angular-waypoints": "~2.0.0",
+    "angular": "1.5.11",
```
mrkvon: it seems that the `package1-name` and `package1-name-subname` ... changed the sorting order (shortest became last). Where do these changes come from? I see plenty of these around.

package.json
```diff
@@ -145,33 +146,36 @@
...
     "passport-github": "~1.1.0",
     "passport-local": "~1.0.0",
     "passport-twitter": "~1.0.4",
+    "passport": "~0.4.0",
+    "prop-types": "~15.6.2",
     "prunecluster": "~2.1.0",
     "randomcolor": "~0.5.1",
+    "react-dom": "~16.6.0",
     "react-i18next": "~8.3.8",
+    "react": "~16.6.0",
```
mrkvon: :+1: from devDependencies to here! (and a few more)

package.json
simison: No idea 🤷‍♂️ and it seems to have sorted itself with one more `npm install react --save` weird are the ways of `npm` sometimes!

PR #985, opened 2018-12-16
Small linting fixes

simison:
Some tiny linting fixes; these files are actually ignored by linters currently but at least karma-config should be linted in future (I'll PR separately).

PRing to see CI go green.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -3,9 +3,7 @@
 /**
  * Module dependencies.
  */
-var _ = require('lodash'),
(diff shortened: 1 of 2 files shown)


karma.conf.js
```diff
@@ -3,9 +3,7 @@
 /**
  * Module dependencies.
  */
-var _ = require('lodash'),
-    defaultAssets = require('./config/assets/default'),
```
simison: These were un-used.

PR #988, opened 2018-12-16
Update/simplify editorconfig

simison:
Simplified Editorconfig file:

- Keep most commonly implemented editorconfig features and leave the rest of eslint
- Remove types of files we don't even have (such as makefile)
- Reorganised for properly cascading rules (no need to repeat if `**` already captures the rule

http://editorconfig.org

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,7 @@
 # http://editorconfig.org
 root = true
 
-[*]
+[**]
 indent_style = space
 indent_size = 2
 end_of_line = lf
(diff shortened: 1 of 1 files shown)


simison:
Simplified Editorconfig file:

- Keep most commonly implemented editorconfig features and leave the rest of eslint
- Remove types of files we don't even have (such as makefile)
- Reorganised for properly cascading rules (no need to repeat if `**` already captures the rule

http://editorconfig.org

PR #990, opened 2018-12-16
Refactor eslint out from Gulpfile

simison:
#### Proposed Changes

* Run eslint directly from npm
* Consolidate Angular and other eslint configs to one; should make it easier to phase out Angular files during rewrite.
* Add `.eslintignore` file with some problematic directories (we'll clean up some of these in later PRs) https://eslint.org/docs/user-guide/configuring#eslintignore

#### Testing Instructions

1. Run `npm start`, `npm run test:server:watch` or `npm run test:client:watch`
2. Modify a file, you should see linting error:
    ```
    trustroots/karma.conf.js (1/0)
      ✖  21:57  Unexpected trailing comma  comma-dangle
    
    ✖ 1 error (9:13:59 PM)
    ```
3. Fix the file, you should see linting success:
    ```
    ✓ Clean (9:14:01 PM)
    ```

Confirm that `npm run lint` works, too.

Confirm that changing anything in Angular files (e.g. `const test = window.test;`) brings up warning about using Angular `$window` service, while that change would work in React files.


simison:
Next up after this we can add Prettier in https://github.com/Trustroots/trustroots/pull/932

simison:
>create a linting error and save file. Error. Open another file and save. Clean. So it lints only the last changed file. Therefore not reliable and npm run lint will need to be run separately. Is that a feature or unexpected behaviour?

Valid point; I disabled re-running all files because it felt slow but I've just added cache and now all the errors are shown continuously. Thanks for pointing that out.


> In watch scripts it is not clear where the Clean comes from. If I'm a random developer, I don't know that it's a linter talking.

I agree — I'd like to hide the whole message but doesn't seem to be possible. Opened an issue: https://github.com/rizowski/eslint-watch/issues/147

>A good opportunity to add "lint:fix": "eslint --fix ." to scripts.

I've been cautious of adding this to _save_ (and have it at `git commit` instead) but let's do it! :muscle: We can change it to commit only if it gets on our way somehow.

.travis.yml
```diff
@@ -4,7 +4,7 @@ node_js:
   - 11
 env:
   - $RUN="npm run test:client"
-  - $RUN="npm run test:server:no-lint"
+  - $RUN="npm run test:server"
   - $RUN="npm run travis-ci"
```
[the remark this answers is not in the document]
simison: Correct; just to run them in parallel.

I guess alternatively I could run this directly here as a Travis command:
```bash
concurrently --raw --kill-others-on-fail 'npm:lint' 'npm:build:prod'
```

...but it's a bit much to show in here:
image

.eslintrc.js
```diff
@@ -1,14 +1,103 @@
...
+};
+
 const es2018rules = {
   'no-var': 2,
   'prefer-const': 2,
   'arrow-spacing': [2, { before: true, after: true }]
 };
 
+const angularRules = {
+  ...defaultRules,
```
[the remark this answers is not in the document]
simison: Good catch, removed.
https://github.com/Trustroots/trustroots/pull/990/commits/4071ba05d38f6c283378dd0d6022f437b159ea8a

simison:
@mrkvon I addressed your feedback — good for another look!

package.json
```diff
@@ -34,7 +34,8 @@
     "ensure-uploads-dir-exists": "node ./bin/ensure-uploads-dir-exists.js",
     "fontello": "gulp fontello",
     "install-dependencies": "node ./bin/install-deps.js",
-    "lint": "gulp lint",
+    "lint:watch": "esw --cache --fix --watch .",
```
mrkvon: > `--fix`

I was rather thinking about adding a new separate script - `lint:fix` which would fix the linting errors on demand only.
Fixable rules coming with newer EcmaScript like `prefer-const` and `no-var` can bring havoc to our migrated files (with lots of `var`) when uncontrolled.
I'd prefer to keep `--fix` out of watch mode, or at least keep it optional.
What do you think?

simison: Valid point. Reverted and added documentation for how to use `--fix` manually:
```
npm run lint -- --fix
npm run lint:watch -- --fix
```

simison: > Cool! I didn't know one could pass additional parameters to npm like that.

I don't have many complains about NPM but with [`yarn`](https://yarnpkg.com/en/) one wouldn't need that extra clunky looking `--`  in between. 🤷‍♂️ 

>And thanks for keeping up with my whims. 🙂

Your attention to detail pleases me. ❤️

PR #918, opened 2018-12-05
List user references

mrkvon:
#### Proposed Changes

* Read references of user.
React component.

`/profile/{username}/references`
- [x] show list of references which user received
  - [x] name, username, avatar, interactions, recommendation, when the reference was given
  - [ ] design
- [ ] when looking at my references, show which ones I haven't given back (public or private) and how much time is left till they become public. Provide a button.
  - [x] show unpublished references to me
    - [x] avatar, link, time left to give
    - [ ] design
- [ ] show list of references which user gave
- [ ] when looking at my references, show which ones I gave and they're still private (and time left)

#### Testing Instructions

* manual testing

Continues fixing #718 
Depends on #827 
It will need a UX and looks design.

#### To do
- [ ] show a number of references in user tab (separated by recommend (yes,no,unknown))
- [ ] show a list of references when the tab is clicked

### Screenshots

05_list_references_self

PR #997, opened 2018-12-19
Rewrite avatar for React

mrkvon:
This component mimics the behaviour of [angular `tr-avatar` directive](https://github.com/Trustroots/trustroots/blob/master/modules/users/client/directives/tr-avatar.client.directive.js).

We hardcoded the base url for gravatar image. Before it was using protocol, host and port from configuration.
Also we hardcoded `https` where protocol had been variable.

prerequisite for #918 and #965


modules/users/client/components/Avatar.component.js
```diff
@@ -0,0 +1,129 @@
...
+};
+
+Avatar.propTypes = {
+  ...AvatarHOC.propTypes,
+  t: PropTypes.func.isRequired
+};
+
+Object.defineProperty(AvatarHOC, 'name', { value: 'Avatar' });
+
+export default AvatarHOC;
```
simison: To me, this structure with component+translation HOC reads a bit funky and is confusing because Avatar is now named as the HOC, not the translation HOC.

I'm guessing `defineProperty` trick is to give it a name for React devtools to recognise it?

If we start copying this pattern elsewhere, it'll have an impact on code readability and potentially make enchancing components by composing multiple HOCs more complicated:
```js
export default compose(
	withNamespaces,
	someOtherHoc
)( Avatar );
```

Would something from [these examples work?](https://react.i18next.com/components/withnamespaces#sample)

My favourite, simply:
```js
export default withNamespaces('user')(Avatar);
```

I think it's okay if `t` isn't covered by propTypes?

PR #998, opened 2018-12-19
Only update tribes in db after all users have been added during seeding

### LINKED ISSUE #994: Refactored seeding script hangs when run in docker
I'm still trying to run TR in docker. The refactored seeding of #902 now results in the error below, and then hangs. Would you have a look @Jagjeet ?

I'm bringing up with `docker-compose up` and try to run the script using `docker-compose  exec trustroots npm run seed`. Reverting #902 and it works again.

```
> trustroots@0.3.3 seed /trustroots                                                                                
> npm run seed:tribes 100 && npm run seed:users 1000 -- --userNames admin1 admin2 admin3 && npm run seed:messages 2
000 10                                                                                                             
                                                                                                                   
                                                                                                                   
> trustroots@0.3.3 seed:tribes /trustroots                                                                         
> node ./bin/fillTestData/Tribes.js "100"                                                                          
                                                                                                                   
                                                                                                                  
Loaded "development" environment configuration                                                                    
                                                                                                                  
YError: Too many arguments provided. Expected max 2 but received 3.                                               
    at module.exports (/trustroots/node_modules/yargs/lib/argsert.js:31:13)                                       
    at Object.Yargs.self.usage (/trustroots/node_modules/yargs/yargs.js:425:5)                                    
    at Object.<anonymous> (/trustroots/bin/fillTestData/Tribes.js:12:18)                                          
    at Module._compile (internal/modules/cjs/loader.js:688:30)                                                    
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)                                      
    at Module.load (internal/modules/cjs/loader.js:598:32)                                                        
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)                                                      
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)                                               
    at Function.Module.runMain (internal/modules/cjs/loader.js:741:12)                                            
    at startup (internal/bootstrap/node.js:285:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:739:3)
Generating undefined tribes...
```

Jagjeet:
#### Proposed Changes

* Previously the new tribe count was updated in the DB on every user addition to the tribe. This change waits until all the users have been added and then updates the tribe counts for all the tribes in the DB. (eg. saving a ton of database calls)

#### Testing Instructions

* Verify the script works as previously. 
* Test with Docker 

Fixes #994

### DIFF
--- a/bin/fillTestData/Users.js
+++ b/bin/fillTestData/Users.js
@@ -69,7 +69,7 @@ function random(max) {
 /**
  * Generates a random float value for locations
  *
- * @returns {float} 
+ * @returns {float}
  */
 function randomizeLocation() {
   let random = Math.random();
--- a/bin/fillTestData/Users.js
+++ b/bin/fillTestData/Users.js
@@ -141,8 +141,7 @@ const addUsers = function () {
           const getTribes = Tribe.find();
 
           Promise.all([getUsers, getTribes]).then((results) => {
-            users = results[0];
-            tribes = results[1];
+            [users, tribes] = results;
             done(null);
           }).catch(function (err) {
             console.log(err);
--- a/bin/fillTestData/Users.js
+++ b/bin/fillTestData/Users.js
@@ -136,27 +136,12 @@ var addUsers = function () {
       async.waterfall([
 
         function getUsersAndTribes(done) {
-          var getUsers = new Promise(function (resolve, reject) {
-            User.find(function (err, foundUsers) {
-              if (err) {
-                reject(err);
-              }
-              users = foundUsers;
-              resolve();
-            });
-          });
+          const getUsers = User.find();
+          const getTribes = Tribe.find();
 
-          var getTribes = new Promise(function (resolve, reject) {
-            Tribe.find(function (err, foundTribes) {
-              if (err) {
-                reject(err);
-              }
-              tribes = foundTribes;
-              resolve();
-            });
-          });
-
-          Promise.all([getUsers, getTribes]).then(function () {
+          Promise.all([getUsers, getTribes]).then(function (results) {
+            users = results[0];
+            tribes = results[1];
             done(null);
           }).catch(function (err) {
             console.log(err);
--- a/bin/fillTestData/Users.js
+++ b/bin/fillTestData/Users.js
@@ -86,7 +86,7 @@ const addOffer = function (id, index, max, usersLength, limit, callback) {
   offer.location = location;
   offer.locationFuzzy = location;
 
-  offer.save(function (err) {
+  offer.save((err) => {
     if (err != null) console.log(err);
     else {
       savedOffers++;
@@ -129,8 +129,8 @@ const addUsers = function () {
   config.db.debug = debug;
 
   // Bootstrap db connection
-  mongooseService.connect(function () {
-    mongooseService.loadModels(function () {
+  mongooseService.connect(() => {
+    mongooseService.loadModels(() => {
       const Tribe = mongoose.model('Tribe');
       const User = mongoose.model('User');
 
@@ -140,7 +140,7 @@ const addUsers = function () {
           const getUsers = User.find();
           const getTribes = Tribe.find();
 
-          Promise.all([getUsers, getTribes]).then(function (results) {
+          Promise.all([getUsers, getTribes]).then((results) => {
             users = results[0];
             tribes = results[1];
             done(null);
@@ -264,7 +264,7 @@ const addUsers = function () {
           } else {
             // Update tribes
             for (let j = 0; j < tribes.length; j++) {
-              Tribe.findByIdAndUpdate(tribes[j]._id, tribes[j], function (err) {
+              Tribe.findByIdAndUpdate(tribes[j]._id, tribes[j], (err) => {
                 if (err) {
                   console.error(err);
                 }
@@ -280,7 +280,7 @@ const addUsers = function () {
 
         // disconnect from mongo
         function disconnect(done) {
-          mongooseService.disconnect(function () {
+          mongooseService.disconnect(() => {
             done(null);
           });
         }
--- a/bin/fillTestData/Users.js
+++ b/bin/fillTestData/Users.js
@@ -322,7 +322,7 @@ function addUsers() {
           let numTribesUpdated = 0;
 
           // If we didn't add any users, tribes do not need to be updated
-          if (savedUsers === 0) {
+          if (savedUsers === 0 || tribes.length === 0) {
             resolve();
           } else {
             // Update tribes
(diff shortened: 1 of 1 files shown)


Jagjeet:
@quite & @simison - This should make the user seeding much more efficient which was the main bottleneck in seeding. In my own testing, it seems significantly faster (eg. 5000 users in a few minutes). 

@quite - Can you verify if this fixes your docker issues?  I want to make sure if you were experiencing a hang versus super slow code. If you see issues, I can try and get up and running with docker as well.

Jagjeet:
@quite  - Thanks for your feedback! With a little restructuring/promises, it should be possible to save all the tribes after adding all the users. I had also intended to separate out the `addNextUser()` method during previous restructuring, but didn't get to it. The main function has also got a bit unwieldy in terms of length/readability so definitely could stand to break things out more. 

Regarding the anonymous function, you might have also been looking for the term `IIFE`. I always saw them, but only recently did I discover there was term for it. This [article](https://medium.com/@vvkchandra/essential-javascript-mastering-immediately-invoked-function-expressions-67791338ddc6) has a pretty good summary. 

Anyways, I will hopefully get to dive back into this tonight California time. Thanks again!

Jagjeet:
I went ahead and made the requested changes and also cleaned up some of the exiting of the User script. Other changes related to breaking things out probably make sense for a dedicated PR or can be implemented as the script evolves. 

The script does start to run out of heap space once there are a lot of users and tribes in the database. (eg. In testing, I had around 14500 users and 400 tribes). Right now, they are both just being stored in memory so if we start needing the script for larger use cases, we'll have to address that. Note that is related to _already existing_ database elements and not the number of elements being added. That said, I have added up to 5000 users without issue. Well, minus poor status for the user as it is hard to tell that the script is still active.  

Let me know if you see anything further. Thanks!

quite:
I'm not very familiar with js. But I have now understood this way of using the waterfall function to chain things to be done, and continue doing them as long as it goes well. The saving of each tribe will only be done once and only if actually needed! :+1:

Jagjeet:
@simison - Did you want to check this out as well? Or should I just go ahead and merge? 

@quite - Thanks! I came from C/C++ and have been slowly getting more experience with it myself. The asynchronous nature of everything takes a while to wrap your brain around and then learning all the control structures to synchronize things where necessary.  Now that I am writing more code day-to-day, it is all clicking much more. :)

bin/fillTestData/Users.js
```diff
@@ -133,138 +134,172 @@ var addUsers = function () {
...
+
+        // disconnect from mongo
+        function disconnect(done) {
+          mongooseService.disconnect(function () {
+            done(null);
           });
         }
-      ]);
 
+      ]); // asyc.waterfall
```
simison: We can now start porting code to ES6 :-) :tada:

@mrkvon has been working on some docs here: https://github.com/Trustroots/trustroots/pull/1001

mrkvon: Yes! Any readability feedback and what is missing is much appreciated. Still heavily a work in progress.

bin/fillTestData/Users.js
```diff
@@ -133,138 +134,172 @@ var addUsers = function () {
...
+            User.find(function (err, foundUsers) {
+              if (err) {
+                reject(err);
+              }
+              users = foundUsers;
+              resolve();
+            });
+          });
+
+          var getTribes = new Promise(function (resolve, reject) {
```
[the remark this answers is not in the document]
Jagjeet: Thanks! Jumping back on this. With the holidays and turning 40 got a bit delayed in getting back to it.

bin/fillTestData/Users.js
Jagjeet: @simison - I went ahead and made a good amount of ES6 changes to the User script, but was seeing a failure due to async/await syntax. Apparently, the `bin` directory had its own `.eslintrc.js` file which had not been updated to use `ecmaVersion2018`. 

I went ahead and made the change, but just want to be sure we don't need to do any overrides similar to what you did for the main eslint file. It didn't to appear to cause any issues in the other bin directory files, but just want to make sure I am not missing anything. 

Thanks! I can update the Tribes and Messages in another PR.

bin/fillTestData/Users.js
```diff
@@ -1,23 +1,23 @@
...
+      path = require('path'),
+      mongooseService = require(path.resolve('./config/lib/mongoose')),
+      chalk = require('chalk'),
+      yargs = require('yargs'),
+      faker = require('faker'),
+      fs = require('fs'),
+      moment = require('moment'),
+      mongoose = require('mongoose'),
+      config = require(path.resolve('./config/config')),
+      cities = JSON.parse(fs.readFileSync(path.resolve('./bin/fillTestData/data/Cities.json'), 'utf8'));
```
mrkvon: Unless the cities get edited + used elsewhere, `cities = path.resolve('./bin/fillTestData/data/Cities')` [may be sufficient](https://nodejs.org/dist/latest-v11.x/docs/api/modules.html#modules_file_modules) here.

bin/.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   parserOptions: {
-    ecmaVersion: 6,
```
mrkvon: I wonder what is the reason for having a separate `.eslintrc` here. I guess there would be too many linting errors from the past, based on the default `.eslintrc`.

We should aim for replacing this eslintrc with the default one. Not in this PR though.

bin/fillTestData/Users.js
Jagjeet: @mrkvon - Thanks! I am assuming you meant `cities = require(path.resolve('./bin/fillTestData/data/Cities.json'));`. Just added the change.

Jagjeet:
I finished up a first stab at ES6 conversion and also added some comments as well. Let me know if there is any other feedback or if we are good to merge. Thanks!

bin/fillTestData/Users.js
simison: >Apparently, the `bin` directory had its own `.eslintrc.js` file which had not been updated to use `ecmaVersion2018`.

Yep, that was by mistake — fixed in with https://github.com/Trustroots/trustroots/pull/1042 and ES6 is ok again in `bin/` :-)

bin/.eslintrc.js
simison: Resolved in https://github.com/Trustroots/trustroots/pull/1042

simison:
Rebased:
- No more `bin/.eslintrc.js`
- Tiny change to `replace` function in https://github.com/Trustroots/trustroots/commit/39037f3f58912af8ddbdf9e77cd31342d0a10353

PR #999, opened 2018-12-19
remove tagline from meet search results

### LINKED ISSUE #979: update Kasper's welcome message
Gotta update this:
> 
> Hi guaka guaka!
> 
> I'm Kasper from the Trustroots Foundation. How are your first impressions from Trustroots — anything missing? If you have any feedback for us, you can simply reply to this email.
> 
> Trustroots is not only about hosting others but also about meeting community members. To make it easier for others to find you, you can create a meetup.
> 
> Like our Facebook page, join our Facebook group and follow us on Twitter or Instagram — it would really help us out!
> 
> Have a good one! Kasper 

- [ ] remove social media stuff (we have 118 followers on instagram, so doesn't really do much)
- [ ] instead, tell people to tell their friends about trustroots

guaka:
#### Proposed Changes

* remove social media links from "my" welcome message

#### Testing Instructions

* check maildev

Fixes #979

### DIFF
--- a/modules/search/client/views/search-sidebar-results.client.view.html
+++ b/modules/search/client/views/search-sidebar-results.client.view.html
@@ -64,6 +64,7 @@
         </span>
       </div>
       <div class="search-result-tagline"
+           ng-if="search.offer.type != 'meet'"
            ng-if="::search.offer.user.tagline"
            ng-bind="::search.offer.user.tagline"></div>
       <div class="search-result-hosting">


guaka:
I think something weird happened with the PR description :)

modules/search/client/views/search-sidebar-results.client.view.html
```diff
@@ -64,6 +64,7 @@ <h4>
         </span>
       </div>
       <div class="search-result-tagline"
+           ng-if="search.offer.type != 'meet'"
```
simison: You could use [one-time `::` binding](https://docs.angularjs.org/guide/expression#one-time-binding) here to make Angular faster (see other examples from this file)

guaka:
> The element has ng-if twice; they should be combined into one.

It works, and it looks clear enough to me. Plus we're migrating to React anyway :)

simison:
>>The element has ng-if twice; they should be combined into one.
>
>It works, and it looks clear enough to me. Plus we're migrating to React anyway :)

Lazy! :-)

I'm just trying to keep the codebase readable, clear and predictable for volunteers who might not always be very experienced. They're already complaining it's hard to understand at times.

Even I couldn't tell what happens in Angular with double `ng-if` without checking — it's not a clear pattern.

I combined the if into one, good to merge now 👍

PR #1000, opened 2018-12-19
User profile basic react

agatatalita:
#### Proposed Changes

* rewrite profile sidebar to React
* add internationalization with i18next to profile sidebar

#### Testing Instructions

* manual testing


mrkvon:
#1000 :tada:

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -13,9 +34,87 @@ export default function ProfileViewBasics({ profile }) {
...
+          {profile.locationFrom}</a>
+      </div>}
+    {profile.languages.length > 0 &&
+      <div className="profile-sidebar-section">
+        <h4 id="profile-languages">{t('Languages')}</h4>
+        <ul className="list-unstyled" aria-describedby="profile-languages">
+          {profile.languages.map(code => <li key={code}>{t(languages[code], { ns: 'languages' }) || code}</li>)}
+        </ul>
+      </div>}
+      { (hasConnectedAdditionalSocialAccounts() || profile.extSitesBW || profile.extSitesCS || profile.extSitesWS) &&
```
mrkvon: `hasConnectedAdditionalSocialAccounts()` magic.

public/locales/cze/languages.json
```diff
@@ -0,0 +1,3 @@
+{
+    "Polish": "Polski"
```
mrkvon: `Polsky` or `Polština`

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -2,9 +2,30 @@ import React from 'react';
...
+        <br/>
+        {t('Replies within {{replyTime, fromNow}}.', { replyTime: profile.replyTime })}
+      </span>}
+    </div>}
+    {/* show div only if birthdate or gender is present */}
+    {(profile.birthdate || profile.gender) &&
+    <div className="profile-sidebar-section" >
+      {profile.birthdate && t('{{birthdate, age}} years', { birthdate: new Date(profile.birthdate) })}
+      {(profile.birthdate && profile.gender) && <span>, </span>}
+      <span className={(!profile.birthdate) ? 'text-capitalize' : null }>{t(profile.gender)}.</span>
```
simison: We should probably introduce [`classnames`](https://www.npmjs.com/package/classnames) module here for easier conditional management of class selectors.

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -13,9 +34,87 @@ export default function ProfileViewBasics({ profile }) {
         {(profile.seen) ? t('Online {{date, fromNow}}', { date: new Date(profile.seen) }) : t('Online long ago')}
       </span>
     </div>
+    {profile.locationLiving &&
```
simison: Another way to structure this component might be:


```js
const locationLiving = locationLiving => (
  <div className="profile-sidebar-section">
         <i className="icon-fw icon-building text-muted"></i>
         {t('Lives in')}
         <a ui-sref="search.map({location: locationLiving})">
           {locationLiving}
         </a>
  </div>
);

{ profile.locationLiving && this.renderLocationLiving( profile.locationLiving ) }}
```

It just reads different and might make the main code flow (with lots of conditional rendering) little easier to read, but up to you how you see best. :-)

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -13,9 +34,87 @@ export default function ProfileViewBasics({ profile }) {
...
+          </h4>
+          <ul className="social-profiles list-unstyled"
+            aria-describedby="profile-networks">
+            {/*
+              Facebook profile link is hidden here until issue with their API gets resolved
+              See https://github.com/Trustroots/trustroots/issues/237
+            */}
+            {profile.additionalProvidersData && Object.keys(profile.additionalProvidersData).map(
+              (key, indeks) =>
+              {return (key !== 'facebook') && <li className="social-profile" key={indeks}><i className={`social-profile-icon icon-fw icon-lg icon-${key}`}></i>
```
simison: Since this is a list with some unique values, we can use those unique values as list's key.

React supports using an index as a key, but it's recommended to avoid it if possible:

> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.

via https://reactjs.org/docs/lists-and-keys.html

In this specific case it probably wouldn't cause re-renderings in any situation but what do you think about doing something like this:

```jsx
Object.keys(profile.additionalProvidersData).map( (service) => {
  return service !== 'facebook' && (
    <li className="social-profile" key={network}>
      <i className={`social-profile-icon icon-fw icon-lg icon-${network}`}></i>
      ...
```

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -13,9 +34,87 @@ export default function ProfileViewBasics({ profile }) {
...
+        </div>}
   </>) }</NamespacesConsumer>);
 };
 
 ProfileViewBasics.propTypes = {
-  profile: PropTypes.object
+  profile: PropTypes.object,
+  hasConnectedAdditionalSocialAccounts: PropTypes.func,
+  isWarmshowersId: PropTypes.func,
+  socialAccountLink: PropTypes.func
```
simison: I wonder if we could put these utility functions elsewhere and import them to both here in React component and also in Angular controller, so that we wouldn't need to pass them down to React component from the Angular controller.

@mrkvon suppose we cannot `import` in Angular controllers just yet?

These specific ones are so small that I wouldn't really mind even just to duplicate them for the time being.

I'd put them in `modules/users/client/utils/networks.js`

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -2,9 +2,30 @@ import React from 'react';
 import { NamespacesConsumer } from 'react-i18next';
 import '@/config/lib/i18n';
 import PropTypes from 'prop-types';
+import * as languages from '@/config/languages/languages';
 
-export default function ProfileViewBasics({ profile }) {
-  return (<NamespacesConsumer ns="user-profile">{ t => (<>
+export default function ProfileViewBasics({ profile, hasConnectedAdditionalSocialAccounts, isWarmshowersId, socialAccountLink }) {
```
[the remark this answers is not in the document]
simison: Whoops, missed your comment and wrote essentially the same here:

https://github.com/Trustroots/trustroots/pull/1000#discussion_r243778511

modules/users/client/components/ProfileViewBasics.component.js
simison: For translations, we shouldn't probably pass variables as they're not picked by scripts pulling out strings for translating.

I know this structure is ported from the Angular-directive but we could improve it. Maybe something along these lines?

```jsx
const getBirthdate = (birthdate) => t('{{birthdate, age}} years', { birthdate: new Date(birthdate);

const getGender = (genderCode, sentence) => {
  switch(genderCode) {
    case 'female':
      return sentence ? t('Female') : t('female');
    // ...
  }
}

if(profile.birthdate && profile.gender) {
  sprintf( // just demonstrating, what ever equavelent of this would be in i18n-next
    t(%birthdate, %gender),
    getBirthdate(profile.birthdate),
    getGender(profile.gender, false)
  );
} else if(profile.birthdate) {
  getBirthdate(profile.birthdate);
} else if(profile.gender) {
  getGender(profile.gender, true);
}
```

modules/users/client/components/ProfileViewBasics.component.js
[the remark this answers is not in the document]
agatatalita: I moved them to the modules/users/client/utils/networks.js.
Should we keep them in two places (here and in Angular) until everything will be in React?

public/locales/cze/languages.json
agatatalita: Still polishing my Czech.

modules/users/client/components/ProfileViewBasics.component.js
agatatalita: I added that module.

modules/users/client/components/ProfileViewBasics.component.js
agatatalita: I like this way of structuring it.
Now it would cause repeating some code for translating and I suppose @mrkvon discovered a simpler way of including t() function which I would use for that.
So work in process on this, probably till tomorrow.

modules/users/client/components/ProfileViewBasics.component.js
agatatalita: Thanks for the article, it is interesting to know.
I changed it.

modules/users/client/components/utils/networks.js
```diff
@@ -0,0 +1,35 @@
+/**
+  * Determine if given user handle for Warmshowers is an id or username
+  * @link https://github.com/Trustroots/trustroots/issues/308
+  */
+function isWarmshowersId(profile) {
+  let x;
+  return isNaN(profile.extSitesWS) ? !1 : (x = parseFloat(profile.extSitesWS), (0 | x) === x);
```
mrkvon: Can you explain what's happening here? I have hard time understanding it.
1. does writing `!1` instead of `false` bring some improvement?
2. `(x = parseFloat(profile.extSitesWS), (0 | x) === x)` what does this code do? I see bitwise operators, assignment of something to x etc. and I don't understand.

agatatalita: I copied that function from the previous place without editing it.
As I understand !1 is just for making it shorter (or maybe faster too?).
(0 | x ) removes decimal part so "(0 | x) === x" makes that code return false if what we gave to it(the WS id) was a decimal number.
Although it took me some time to understand it.
Maybe we could change it to something like:

```
function isWarmshowersId(profile) {
  // if it is a number
  if (!isNaN(profile.extSitesWS)){
    const x = parseFloat(profile.extSitesWS);
    // false for decimal number
    return Math.floor(x) === x;
  }
  return false;
}
```

simison: Sorry, this is probably my bad and I have no memory of it anymore! 😆 

WS has Drupal user id _and_ usernames — for some folks links to their profile work with username, but not for everyone (they switched it on or off at some point). ID works for everyone.

So I think the goal is to just recognise if user entered numerical ID (that could start with 0 and that cannot be negative or have decimals). Unless I'm forgetting something important. 😅  

Simple regepx to check only for `0-9` would prolly work?

mrkvon: @agatatalita Thank you for explaining! I learned something new.

public/locales/cze/user-profile.json
```diff
@@ -1,5 +1,16 @@
 {
   "Member since {{date, MMM Do, YYYY}}": "Členem od {{date, Do MMM YYYY}}",
   "Online {{date, fromNow}}": "Online {{date, fromNow}}",
-  "Online long ago": "Online dávno"
+  "Online long ago": "Online dávno",
```
mrkvon: It's nice to have these translations which we can use for testing.

Anyways this is something that we don't have to do. We can leave it for people, who will be translating (when we figure out nice tools for that).

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,21 +1,174 @@
...
+          <a className="social-profile-handle"
+            href={`https://www.warmshowers.org/${isWarmshowersId(profile) ? 'user' : 'users' }/${profile.extSitesWS}`}>
+            Warmshowers
+          </a>
+        </li>}
+      </ul>
+    </div>
+  );
+
+  return (<>
```
mrkvon: I love the way this is refactored and split to smaller pieces! :+1:

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,21 +1,174 @@
...
       <span>
         {(profile.seen) ? t('Online {{date, fromNow}}', { date: new Date(profile.seen) }) : t('Online long ago')}
       </span>
+    </div>;
+  };
+
+  const renderLocationLiving = locationLiving => (
+    <div className="profile-sidebar-section">
+      <i className="icon-fw icon-building text-muted" />
+      {t('Lives in')}{' '}
```
mrkvon: It would be nice to have `t('Lives in {{location}}', { location: locationLiving })`. Some languages don't keep this word order.

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,21 +1,174 @@
...
+        {locationLiving}
+      </a>
     </div>
-  </>) }</NamespacesConsumer>);
+  );
+
+  const renderLocationFrom = locationFrom => (
+    <div className="profile-sidebar-section">
+      <i className="icon-fw icon-home text-muted"></i>
+      {t('From')}{' '}
```
mrkvon: :point_up:

simison: As a general rule, it's important to avoid partial sentences with translations as they're difficult for translators.

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,21 +1,174 @@
...
         {(profile.seen) ? t('Online {{date, fromNow}}', { date: new Date(profile.seen) }) : t('Online long ago')}
       </span>
+    </div>;
+  };
+
+  const renderLocationLiving = locationLiving => (
+    <div className="profile-sidebar-section">
+      <i className="icon-fw icon-building text-muted" />
+      {t('Lives in')}{' '}
+      <a ui-sref="search.map({location: locationLiving})">
```
mrkvon: the `ui-sref` won't work here. `href`.

modules/users/client/components/ProfileViewBasics.component.js
mrkvon: :point_down: The same thing with `ui-sref` as :point_up:

modules/users/client/components/utils/networks.js
```diff
@@ -0,0 +1,35 @@
...
+  * Determine if given user handle for Warmshowers is an id or username
+  * @link https://github.com/Trustroots/trustroots/issues/308
+  */
+function isWarmshowersId(profile) {
+  // WarmShowers id should contain only digits
+  return /^[0-9]*$/.test(profile.extSitesWS);
+}
+
+/**
+     * Check if there are additional accounts
```
mrkvon: I wonder if we can have eslint rule for this. I wouldn't expect this indentation (here and line 18).

This is truly a [nitpicking](https://en.wikipedia.org/wiki/Nitpicking). I don't feel a need to have this changed. :)

modules/users/client/components/utils/i18n-angular-load.js
```diff
@@ -0,0 +1,28 @@
+import '@/config/lib/i18n';
```
mrkvon: This can be deduplicated (rebase and use the file which is already in master).

Already discussed with @agatatalita .

mrkvon:
Added a few additional comments. When the significant ones get resolved, I'd say this is ready for merge.

modules/users/client/components/ProfileViewBasics.component.js
agatatalita: Added render functions.

modules/users/client/components/utils/networks.js
agatatalita: Should we add: 'multiline-comment-style': ["error", "starred-block"] ? :slightly_smiling_face:

mrkvon: There is a rule for that? Great. I'd say let's add it, but not necessarily now.

modules/users/client/components/ProfileViewBasics.component.js
mrkvon: Well, the `locationLiving={locationLiving}` as attribute of `Trans` seems to not be necessary, right? Works without.

mrkvon:
I noticed I don't see the `Online ... ago`.

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,21 +1,173 @@
...
+    </div>
+  );
+
+  const renderMemberSince = () => (
     <div className="profile-sidebar-section">
       {t('Member since {{date, MMM Do, YYYY}}', { date: new Date(profile.created) })}
     </div>
+  );
+
+  const renderSeenOnline = () => {
```
mrkvon: Curly brackets should be replaced with parentheses, otherwise nothing is returned and Seen Ago is not rendered.

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,21 +1,213 @@
...
-export default function ProfileViewBasics({ profile }) {
-  return (<NamespacesConsumer ns="user-profile">{ t => (<>
+
+export function ProfileViewBasics({ t, profile }) {
+
+  /*
+  * Functions passing strings to translation fuction for translation scripts
+  */
+  const getGender = (genderCode) => (
+    t(genderCode)
```
simison: Translation string extractor won't pick this up, because it's a variable and it doesn't know what's in that variable. In Angular version, we weren't too concerned beacause we didn't have i18n set up.

Previous version (https://github.com/Trustroots/trustroots/commit/7a5053050e96cf5fc2f63c3fb49b1b62a60ece55#diff-5be83c916d70e68382e672abc55aa996L15) of this function wasn't that great either because it had partial strings meant to be placed in between sentences, while in some languages the whole sentence should be translated differently depending on gender. More over it didn't seem like we were using sentence version anywhere, so it was too early optimization.

I'd suggest having this:

```js
  const getGender = (genderCode) => (
    switch (genderCode) {
      case 'female':
        return t('Female');
      case 'male':
        return t('Male');
      case 'non-binary':
        return t('Non-binary');
      case 'other':
        return t('Other');
      default:
        return null; // React expects `null` for non-rendering
    }
  };
```

Does that make sense?

simison: As a side note, is there a way to define "context" for strings here so that it's easier to translate?

something like `t_with_context('Other', 'gender');`

simison: Looks like I suggested the previous structure myself. :-) https://github.com/Trustroots/trustroots/pull/1000#discussion_r243780448

And now I see that we had "sentence" when there is birthdate: _"33 years, male."_ — since it was such a simple sentence, I didn't even think it in the traditional sense of "sentence". 🤔

agatatalita: I was not sure how 'text-capitalize' will work with languages other than English. Today I checked it briefly and it seemed fine.

mrkvon: @simison 
> As a side note, is there a way to define "context" for strings here so that it's easier to translate?
> something like `t_with_context('Other', 'gender');`

Do you have an example of what you mean?

modules/users/client/components/ProfileViewBasics.component.js
```diff
@@ -1,21 +1,213 @@
...
+  const renderReplyData = (replyRate, replyTime) => (
+    <div className="profile-sidebar-section text-muted">
+      {replyRate &&
+      <span>
+        {getReplyRate(replyRate)}
+      </span>}
+      {replyTime &&
+      <span>
+        <br/>
+        {getReplyTime}
```
mrkvon: We need to call this function.

PR #1001, opened 2018-12-19
Manual for ES2018 migration

mrkvon:
#### Proposed Changes

* documentation

#### Testing Instructions

* [read it!](https://github.com/Trustroots/trustroots/blob/es2018-docs/docs/ES2018.md)

### DIFF
--- a/docs/Async-waterfall-to-async-await.md
+++ b/docs/Async-waterfall-to-async-await.md
@@ -1,0 +1,163 @@
+# Example: Migrate async.waterfall to ES2018
+
+How to rewrite [async.waterfall](https://caolan.github.io/async/docs.html#waterfall) with [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)?
+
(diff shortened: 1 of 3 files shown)


Jagjeet:
Nice job on this! I used this for making some changes to the User seeding script though still working on some things. It was nice to have a quick checklist to start from. 

I liked that you linked out to MDN for some of the sections. It would be great to link out to the sections for Promises, Async and Await as well. 

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

Regarding examples, I think it is great the way you have short examples and link out for more info. That said, anything we can reference more specific to Trustroots usage is helpful. I found it a little confusing going from `async.waterfall` to using `async/await` so an example related to that might be more helpful. 
 
I might also consider an example of destructuring from an array though MDN has that as well. Something like: 

```
//old 
var x = [0, 1];
var a = x[0];
var b = x[1];

//new
let x = [0, 1];
let [a, b] = x;
```
Lastly, I would add a brief mention of `let/const` having block scope versus `var` having function scope.

simison:
Excellent points by @Jagjeet 

I think we can merge even as is, and then iterate.

mrkvon:
I made some changes following @Jagjeet's recommendation.

The code example for async.waterfall to async/await is probably clumsy and hard to read. But I'm fine with it for now. This is not a critical stuff.

PR #1002, opened 2018-12-19
Developer's manual to i18n

mrkvon:
#### Proposed Changes

* Documentation. How to make an internationalized component in React.

#### Testing Instructions

* [read it!](https://github.com/Trustroots/trustroots/blob/i18n-docs/docs/i18n.md)

It doesn't aim to be perfect, but wants to improve. You can help if you use it and see: what is useful, what is not, what is missing. Please provide the feedback!

### DIFF
--- a/docs/React.md
+++ b/docs/React.md
@@ -94,6 +94,10 @@ export default class ProfileViewBasics {
 }
 ```
 
+## Internacionalization
+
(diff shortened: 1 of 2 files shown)


simison:
Thanks for adding these, super helpful! 🙌

PR #1004, opened 2018-12-20
Set defaults for axios

mrkvon:
#### Proposed Changes

* setting defaults for axios library so we can return a response body directly

### DIFF
--- a/config/lib/axios.js
+++ b/config/lib/axios.js
@@ -1,0 +1,6 @@
+import axios from 'axios';
+
+// create an axios instance with defaults
+export default axios.create({
+  transformResponse: [({ data }) => data] // return only request body
+});
--- a/modules/users/client/api/search-users.api.js
+++ b/modules/users/client/api/search-users.api.js
@@ -1,4 +1,4 @@
-import axios from 'axios';
+import axios from '@/config/lib/axios';
 
 export async function searchUsers(query) {
   return await axios.get(`/api/users?search=${query}`);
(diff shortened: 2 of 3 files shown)


mrkvon:
Unfortunately `transformResponse` in config transforms only `data`. Not the whole response. This is not useful.

simison:
Do you mean it still returns the complete request response?

As a side note, we could rename export to `api` or similar?

config/lib/axios.js
```diff
@@ -0,0 +1,6 @@
+import axios from 'axios';
+
+// create an axios instance with defaults
+export default axios.create({
+  transformResponse: [({ data }) => data] // return only request body
+});
```
simison: Untested but we might need something like this:

```js
// Add a response interceptor
const axiosInstance = axios.create();

axiosInstance.interceptors.response.use((response) => {
    return Promise.resolve(response.data); // or just `return response.data;`?
  }, (error) => {
    // In Angular, we would emit a global event at this point and the application knows what to do with it (e.g. it would show "app is offline" or throw you to the login page.
    return Promise.reject(error);
});

export default axiosInstance;
```

modules/users/client/api/search-users.api.js
```diff
@@ -1,4 +1,4 @@
-import axios from 'axios';
+import axios from '@/config/lib/axios';
```
simison: We might need to have a separate high-level folder like this for the client since config is currently serverside only and if we start mixing these, separating concerns might get harder in future.

simison: Perhaps [`modules/core/client/config/`](https://github.com/Trustroots/trustroots/tree/327a47bcb4420afdd95b98ae787594939e301400/modules/core/client/config) or would that be mixing with angular files too much?

We could very well rename some Angular files with `.angular.` just to make it clearer.

PR #1005, opened 2018-12-20
rewrite profile about me in react with i18next

agatatalita:
#### Proposed Changes

* rewrite profile about me in React
* add i18next translation to profile about 

#### Testing Instructions

* manual testing: go to `/profile/:username` and observe that everything works the same as before.
*  `i18n` translations should work if data for translations are added to `public/locales/:languagecode` files

### DIFF
--- a/modules/users/client/components/AboutMe.component.js
+++ b/modules/users/client/components/AboutMe.component.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import { NamespacesConsumer } from 'react-i18next';
+import '@/config/lib/i18n';
+import PropTypes from 'prop-types';
+import { limitTo, plainTextLength } from './utils/filters';
+
+export default function AboutMe({ profile, profileDescriptionToggle, user, appSettings }) {
+  return (<NamespacesConsumer ns={['user-profile']}>{ (t) => (<>
+    <section className="panel panel-default">
+      <header className="panel-heading">
+        {t('About me')}
+      </header>
+      <div className="panel-body">
+        {/* Short descriptions */}
+        {(profile.description.length < 2400 || profileDescriptionToggle) &&
+        <div dangerouslySetInnerHTML={{ __html: profile.description }} ></div>}{/* TODO Can we trust this HTML?*/}
+
+        {/* Long descriptions */}
+        {profile.description.length >= 2400 && !profileDescriptionToggle &&
+        <div
+          className="panel-more-wrap" ng-click="profileCtrl.profileDescriptionToggle=true">
+          <div className="panel-more-excerpt" setInnerHTML={limitTo(profile.description, 2400)}> </div>
+          <div className="panel-more-fade">{t('Show more...')}</div>
+        </div>}
+
+        {/* If no description, show deep thoughts... */}
+        {(!profile.description || profile.description === '') &&
+        <blockquote className="profile-quote"
+          aria-label={t('Member has not written description about themself.')}>
+        “Everyone is necessarily the hero of their own life story.”
+        </blockquote>}
+      </div>
+
+      {/* User watching their own profile and it's too short */}
+      {(user._id === profile._id && (!profile.description || plainTextLength(profile.description) < appSettings.profileMinimumLength)) &&
+      <footer className="panel-footer">
+        <p className="lead">
+          {t('Your profile description should be longer so that you can send messages.')}
+          <br/><br />
+          <a href="/profile/edit" className="btn btn-primary">{t('Fill your profile')}</a>
+        </p>
+      </footer>}
+    </section>
+  </>) }</NamespacesConsumer>);
+};
+
+AboutMe.propTypes = {
+  profile: PropTypes.object,
+  profileDescriptionToggle: PropTypes.bool,
+  user: PropTypes.object,
+  appSettings: PropTypes.object
+};
--- a/modules/users/client/components/AboutMe.component.js
+++ b/modules/users/client/components/AboutMe.component.js
@@ -1,7 +1,5 @@
 import React, { Component } from 'react';
-import sanitizeHtml from 'sanitize-html';
 import { withNamespaces } from '@/modules/core/client/utils/i18n-angular-load';
-{/* import { Trans } from 'react-i18next'; */}
 import '@/config/lib/i18n';
 import PropTypes from 'prop-types';
 import { limitTo, plainTextLength } from './utils/filters';
@@ -31,13 +29,13 @@ export class AboutMe extends Component {
         <div className="panel-body">
           {/* Short descriptions */}
           {(profile.description.length < 2400 || this.state.profileDescriptionToggle) &&
-          <div dangerouslySetInnerHTML={{__html: sanitizeHtml(profile.description)}}></div>}{/* TODO Can we trust this HTML?*/}
+          <div dangerouslySetInnerHTML={{ __html: profile.description }}></div>}
 
           {/* Long descriptions */}
           {profile.description.length >= 2400 && !this.state.profileDescriptionToggle &&
           <div
             className="panel-more-wrap">
-            <div className="panel-more-excerpt" dangerouslySetInnerHTML={{__html: sanitizeHtml(limitTo(profile.description, 2400))}}></div>
+            <div className="panel-more-excerpt" dangerouslySetInnerHTML={{ __html: limitTo(profile.description, 2400) }}></div>
             <div className="panel-more-fade" onClick={this.changeProfileDescriptionToggle}>{t('Show more...')}</div>
           </div>}
 
--- a/modules/users/client/components/AboutMe.component.js
+++ b/modules/users/client/components/AboutMe.component.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import sanitizeHtml from 'sanitize-html';
 import { withNamespaces } from '@/modules/core/client/utils/i18n-angular-load';
 {/* import { Trans } from 'react-i18next'; */}
 import '@/config/lib/i18n';
@@ -30,13 +31,13 @@ export class AboutMe extends Component {
         <div className="panel-body">
           {/* Short descriptions */}
           {(profile.description.length < 2400 || this.state.profileDescriptionToggle) &&
-          <div>{profile.description }</div>}{/* TODO Can we trust this HTML?*/}
+          <div dangerouslySetInnerHTML={{__html: sanitizeHtml(profile.description)}}></div>}{/* TODO Can we trust this HTML?*/}
 
           {/* Long descriptions */}
           {profile.description.length >= 2400 && !this.state.profileDescriptionToggle &&
           <div
             className="panel-more-wrap">
-            <div className="panel-more-excerpt">{limitTo(profile.description, 2400)}> </div>
+            <div className="panel-more-excerpt" dangerouslySetInnerHTML={{__html: sanitizeHtml(limitTo(profile.description, 2400))}}></div>
             <div className="panel-more-fade" onClick={this.changeProfileDescriptionToggle}>{t('Show more...')}</div>
           </div>}
 
--- a/modules/users/client/components/AboutMe.component.js
+++ b/modules/users/client/components/AboutMe.component.js
@@ -1,52 +1,72 @@
-import React from 'react';
-import { NamespacesConsumer } from 'react-i18next';
+import React, { Component } from 'react';
+import { withNamespaces } from '@/modules/core/client/utils/i18n-angular-load';
+{/* import { Trans } from 'react-i18next'; */}
 import '@/config/lib/i18n';
 import PropTypes from 'prop-types';
 import { limitTo, plainTextLength } from './utils/filters';
 
-export default function AboutMe({ profile, profileDescriptionToggle, user, appSettings }) {
-  return (<NamespacesConsumer ns={['user-profile']}>{ (t) => (<>
-    <section className="panel panel-default">
-      <header className="panel-heading">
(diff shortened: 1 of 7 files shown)


modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,52 @@
...
+          <a href="/profile/edit" className="btn btn-primary">{t('Fill your profile')}</a>
+        </p>
+      </footer>}
+    </section>
+  </>) }</NamespacesConsumer>);
+};
+
+AboutMe.propTypes = {
+  profile: PropTypes.object,
+  profileDescriptionToggle: PropTypes.bool,
```
mrkvon: What is the meaning of `profileDescriptionToggle`?

Also the distinction between `profile` and `user` is not clear on the first sight. As far as I understand, it's _displayed user_ and _logged user_. Renaming of the variables; or at least some comments could help here.

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,52 @@
...
+export default function AboutMe({ profile, profileDescriptionToggle, user, appSettings }) {
+  return (<NamespacesConsumer ns={['user-profile']}>{ (t) => (<>
+    <section className="panel panel-default">
+      <header className="panel-heading">
+        {t('About me')}
+      </header>
+      <div className="panel-body">
+        {/* Short descriptions */}
+        {(profile.description.length < 2400 || profileDescriptionToggle) &&
+        <div dangerouslySetInnerHTML={{ __html: profile.description }} ></div>}{/* TODO Can we trust this HTML?*/}
```
mrkvon: @simison Can we trust `profile.description`? Is it well sanitized when it's saved to database? Or should we worry about [XSS](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)) here?

modules/users/client/components/AboutMe.component.js
mrkvon: The `profileDescriptionToggle` seems to keep the info whether the description is shown or not. Would it be clearer if it lived directly in the React component?

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,52 @@
...
+          className="panel-more-wrap" ng-click="profileCtrl.profileDescriptionToggle=true">
+          <div className="panel-more-excerpt" setInnerHTML={limitTo(profile.description, 2400)}> </div>
+          <div className="panel-more-fade">{t('Show more...')}</div>
+        </div>}
+
+        {/* If no description, show deep thoughts... */}
+        {(!profile.description || profile.description === '') &&
+        <blockquote className="profile-quote"
+          aria-label={t('Member has not written description about themself.')}>
+        “Everyone is necessarily the hero of their own life story.”
```
mrkvon: Should we translate this, too? Or keep it in English?

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,52 @@
...
+      <div className="panel-body">
+        {/* Short descriptions */}
+        {(profile.description.length < 2400 || profileDescriptionToggle) &&
+        <div dangerouslySetInnerHTML={{ __html: profile.description }} ></div>}{/* TODO Can we trust this HTML?*/}
+
+        {/* Long descriptions */}
+        {profile.description.length >= 2400 && !profileDescriptionToggle &&
+        <div
+          className="panel-more-wrap" ng-click="profileCtrl.profileDescriptionToggle=true">
+          <div className="panel-more-excerpt" setInnerHTML={limitTo(profile.description, 2400)}> </div>
```
mrkvon: We'll also need to change this to `dangerouslySetInnerHTML`. Currently longer description isn't shown for me at all (Firefox) and I get a warning instead:

> Warning: React does not recognize the `setInnerHTML` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `setinnerhtml` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,52 @@
...
+          <div className="panel-more-fade">{t('Show more...')}</div>
```
mrkvon: Clicking `Show more...` doesn't work for me. It might be related to the `setInnerHTML` issue above; or to failing logic related to `profileDescriptionToggle`.

modules/users/client/components/AboutMe.component.js
agatatalita: I changed it to the isSelf property.

modules/users/client/components/AboutMe.component.js
agatatalita: I am not sure how to deal with that, I am getting an error while changing it to dangerouslySetInnerHTML.
Can only set one of `children` or `props.dangerouslySetInnerHTML`.
I also don't know whether we should sanitize that HTML or we can trust it.

mrkvon: There are a few sanitization libraries around. I've stumbled upon [DOMPurify](https://github.com/cure53/DOMPurify) and [sanitize-html](https://github.com/punkave/sanitize-html). Even if we trust the user data, it might still be a good security measure to sanitize it in the frontend.

@agatatalita What is your error?

agatatalita: I was getting error: "Can only set one of `children` or `props.dangerouslySetInnerHTML`."
It was because I left a space in the div with property dangerouslySetInnerHTML.
`<div dangerouslySetInnerHTML={{__html: sanitizeHtml(profile.description)}}> </div>`

Now it works and I used sanitize-html.

modules/users/client/components/AboutMe.component.js
simison: We sanitize it on save _and_ output (most importantly):

https://github.com/Trustroots/trustroots/blob/65d2ad2bf7a2aafdc439c6b11994ee1278098501/modules/users/server/controllers/users.profile.server.controller.js#L906-L907

I'd say we can trust that for now. Other fields are not to be trusted.

Note also that _if_ we do any frontend sanitization here, we should pass `textService.sanitizeOptions` to `sanitize-html` so that we sanitize with same settings at the back- end frontend.

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,73 @@
...
+        <div className="panel-body">
+          {/* Short descriptions */}
+          {(profile.description.length < 2400 || this.state.profileDescriptionToggle) &&
+          <div dangerouslySetInnerHTML={{__html: sanitizeHtml(profile.description)}}></div>}{/* TODO Can we trust this HTML?*/}
+
+          {/* Long descriptions */}
+          {profile.description.length >= 2400 && !this.state.profileDescriptionToggle &&
+          <div
+            className="panel-more-wrap">
+            <div className="panel-more-excerpt" dangerouslySetInnerHTML={{__html: sanitizeHtml(limitTo(profile.description, 2400))}}></div>
```
simison: This splitting the string was always a bit risky because we might split in the middle of an HTML tag or link. :-(

It's okay to merge as such since that's how it was functioning previously.

Just some brainstorming for an alternative approach:

- We're really concerned only about the visual length of this area, so perhaps the toggle `profileDescriptionToggle` could only change a className?
- Length of the string doesn't necessarily correlate with the height of the area (depends on so many things such as width and text styling) so that's probably was the reason why I why I went with string splitting originally. :-(
- This (showing longer text vs and "show more") pattern is used in quite a few places in profile, I wonder if we could wrap it in a component and re-use everywhere?

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,73 @@
...
+            <div className="panel-more-fade" onClick={this.changeProfileDescriptionToggle}>{t('Show more...')}</div>
+          </div>}
+
+          {/* If no description, show deep thoughts... */}
+          {(!profile.description || profile.description === '') &&
+          <blockquote className="profile-quote"
+            aria-label={t('Member has not written description about themself.')}>
+          “Everyone is necessarily the hero of their own life story.”
```
simison: Let's simplify this and just not show the whole box if they didn't write anything.

We'll still need to show the area + edit link if `isSelf` is true.

modules/users/client/components/utils/filters.js
```diff
@@ -0,0 +1,7 @@
+export const limitTo = (text, length) => {
+  return text.substring(0, length);
+};
+
+export const plainTextLength = (text) => {
+  return text && typeof(text) === 'string' ? String(text).replace(/&nbsp;/g, ' ').replace(/<[^>]+>/gm, '').trim().length : 0; // eslint-disable-line angular/typecheck-string
```
simison: > `eslint-disable-line angular/typecheck-string`

Hmm, it's surprising that Angular linter is trying to lint here. 

No need to change anything, I'm just noting to look into this when working again on eslint stuff: https://github.com/Trustroots/trustroots/pull/932

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,73 @@
+import React, { Component } from 'react';
+import sanitizeHtml from 'sanitize-html';
+import { withNamespaces } from '@/modules/core/client/utils/i18n-angular-load';
+{/* import { Trans } from 'react-i18next'; */}
```
mrkvon: This can be removed.
(Also, the curly brackets around the comment are only needed inside a jsx/html block.)

modules/users/client/views/profile/profile-view-about.client.view.html
```diff
@@ -1,54 +1,19 @@
...
-        </p>
-      </footer>
-    </section>
-    <!-- /About me -->
-
+    <about-me profile="profileCtrl.profile" isSelf="app.user.id === profileCtrl.profile.id" appSettings="app.appSettings"></about-me>
   </div>
 
   <div class="col-md-6">
+    <!-- <offers profile="profileCtrl.profile" user="app.user"></offers> -->
```
mrkvon: can be removed

modules/users/client/views/profile/profile-view-about.client.view.html
```diff
@@ -1,54 +1,19 @@
...
 
     <!-- Offers -->
     <div class="hidden-xs">
       <div tr-offer-host-view="profileCtrl.profile"
            tr-offer-host-view-auth-user="::app.user"></div>
     </div>
 
+    <!-- <contacts-in-common profile="profileCtrl.profile" user="app.user" contacts="profileCtrl.contacts"></contacts-in-common> -->
```
mrkvon: can be removed

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,73 @@
...
+import '@/config/lib/i18n';
```
mrkvon: This is called in the `i18n-angular-load`, but does no harm here. So it can stay. (Just saying.)

package.json
```diff
@@ -168,7 +168,7 @@
     "react-bootstrap": "~0.32.4",
     "react-dom": "~16.6.3",
     "react-i18next": "~8.4.0",
-    "sanitize-html": "~1.19.3",
+    "sanitize-html": "^1.19.3",
```
mrkvon: can stay as it was (interesting! we were already using sanitize-html before! :slightly_smiling_face:)

modules/users/client/components/AboutMe.component.js
```diff
@@ -0,0 +1,73 @@
...
+            aria-label={t('Member has not written description about themself.')}>
+          “Everyone is necessarily the hero of their own life story.”
+          </blockquote>}
+        </div>
+
+        {/* User watching their own profile and it's too short */}
+        {(isSelf && (!profile.description || plainTextLength(profile.description) < appSettings.profileMinimumLength)) &&
+        <footer className="panel-footer">
+          <p className="lead">
+            {t('Your profile description should be longer so that you can send messages.')}
```
mrkvon: For some reason I see this piece at (everybody with no description)'s profile. As if isSelf is always true.

modules/users/client/views/profile/profile-view-about.client.view.html
```diff
@@ -1,54 +1,19 @@
...
-        <p class="lead">
-          Your profile description should be longer so that you can send messages.
-          <br><br>
-          <a ui-sref="profile-edit.about" class="btn btn-primary">Fill your profile</a>
-        </p>
-      </footer>
-    </section>
-    <!-- /About me -->
-
+    <about-me profile="profileCtrl.profile" isSelf="app.user.id === profileCtrl.profile.id" appSettings="app.appSettings"></about-me>
```
mrkvon: The `id` should be `_id` here. (Something from mongo, I got confused too a couple times.)

See a comment above.

modules/users/client/components/AboutMe.component.js
agatatalita: I removed sanitizing from the frontend.

modules/users/client/components/AboutMe.component.js
simison: Oh and again good for a follow up PR, not a blocker here.

mrkvon:
I believe this is ready for merge.

I just rebased with master and changed the `^` to `~` for sanitize-html in package.json

package.json
simison: I added a config file to keep `~` by default: https://github.com/Trustroots/trustroots/commit/7052711a13a664865132e47b5fe5dc4792c3a44c

PR #1009, opened 2018-12-23
Rewrite contact in React

mrkvon:
#### Proposed Changes

* Rewrite a contact from contact list `/profile/:username/contacts` in React.
  - _remove contact_ modal included

#### Testing Instructions

* Go to `/profile/:username/contacts` and observe that everything works the same as before. :slightly_smiling_face: 
* Additionally `i18n` should work, but we don't have the translations, yet.

I'll probably still write some comments to the code.

### DIFF
--- a/modules/contacts/client/components/Contact.component.js
+++ b/modules/contacts/client/components/Contact.component.js
@@ -1,0 +1,66 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import ContactPresentational from './ContactPresentational';
+import RemoveContact from './RemoveContact';
+
+export default class Contact extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      showRemoveModal: false,
+      removeInProgress: false
+    };
+
+    this.handleRemoveContactModal = this.handleRemoveContactModal.bind(this);
+    this.handleRemoveContact = this.handleRemoveContact.bind(this);
+  }
+
+  handleRemoveContactModal(show) {
+    this.setState(() => ({ showRemoveModal: show }));
+  }
+
+  async handleRemoveContact() {
+    this.setState(() => ({ removeInProgress: true }));
+    await new Promise(resolve => setTimeout(() => resolve(), 2000)); // eslint-disable-line angular/timeout-service
+    this.setState(() => ({ removeInProgress: false }));
+    this.handleRemoveContactModal(false);
+  }
+
+  render() {
+
+    const situation = getSituation(this.props.contact, this.props.selfId);
+
+    return (<>
+      <RemoveContact
+        contact={this.props.contact}
+        show={this.state.showRemoveModal}
+        inProgress={this.state.removeInProgress}
+        onHide={() => this.handleRemoveContactModal(false)}
+        onRemoveContact={this.handleRemoveContact}
+        situation={situation}
+      />
+      <ContactPresentational
+        contact={this.props.contact}
+        avatarSize={this.props.avatarSize}
+        hideMeta={this.props.hideMeta}
+        situation={situation}
+        onClickRemove={() => this.handleRemoveContactModal(true)}
+      />
+    </>);
+  }
+}
+
+Contact.propTypes = {
+  contact: PropTypes.object.isRequired,
+  avatarSize: PropTypes.number,
+  selfId: PropTypes.string.isRequired,
+  hideMeta: PropTypes.bool
+};
+
+function getSituation(contact, selfId) {
+  return (contact.confirmed === false && contact.userFrom === selfId && 'unconfirmedFromMe')
+    || (contact.confirmed === false && contact.userTo === selfId && 'unconfirmedToMe')
+    || 'confirmed';
+}
--- a/modules/contacts/client/components/Contact.component.js
+++ b/modules/contacts/client/components/Contact.component.js
@@ -20,8 +20,8 @@ export default class Contact extends Component {
     this.handleRemoveContact = this.handleRemoveContact.bind(this);
   }
 
-  handleRemoveContactModal(show) {
-    this.setState(() => ({ showRemoveModal: show }));
+  handleRemoveContactModal(showRemoveModal) {
+    this.setState(() => ({ showRemoveModal }));
   }
 
   async handleRemoveContact() {
--- a/modules/contacts/client/components/Contact.component.js
+++ b/modules/contacts/client/components/Contact.component.js
@@ -2,6 +2,9 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ContactPresentational from './ContactPresentational';
 import RemoveContact from './RemoveContact';
+import * as contacts from '../api/contacts.api';
+
+const api = { contacts };
 
 export default class Contact extends Component {
 
@@ -23,9 +26,11 @@ export default class Contact extends Component {
 
   async handleRemoveContact() {
     this.setState(() => ({ removeInProgress: true }));
-    await new Promise(resolve => setTimeout(() => resolve(), 2000)); // eslint-disable-line angular/timeout-service
+    await api.contacts.remove(this.props.contact._id);
     this.setState(() => ({ removeInProgress: false }));
     this.handleRemoveContactModal(false);
+    // broadcast the change to angular
+    this.props.onContactRemoved(this.props.contact);
   }
 
   render() {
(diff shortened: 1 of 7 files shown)


modules/contacts/client/components/Contact.component.js
```diff
@@ -0,0 +1,74 @@
...
+      showRemoveModal: false,
+      removeInProgress: false
+    };
+
+    this.handleRemoveContactModal = this.handleRemoveContactModal.bind(this);
+    this.handleRemoveContact = this.handleRemoveContact.bind(this);
+  }
+
+  handleRemoveContactModal(show) {
+    this.setState(() => ({ showRemoveModal: show }));
```
simison: Shouldn't this be:
```js
this.setState({showRemoveModal: show});
```

... unless I'm missing something?

It could even be shortened to:
```js
handleRemoveContactModal(showRemoveModal) {
  this.setState({showRemoveModal});
}
```

modules/contacts/client/components/RemoveContact.js
```diff
@@ -0,0 +1,75 @@
...
+    unconfirmedToMe: {
+      confirm: 'Yes, decline request',
+      title: 'Decline contact request?',
+      time: 'Requested'
+    },
+    // Removing confirmed contact
+    confirmed: {
+      confirm: 'Yes, remove contact',
+      title: 'Remove contact?',
+      time: 'Connected since'
```
simison: Shouldn't these be wrapped with `t()` already here instead of `t(labels.title)`?

Otherwise parsers won't be able to pick these strings.

modules/contacts/client/components/RemoveContact.js
```diff
@@ -0,0 +1,75 @@
...
+
+        <Modal.Footer>
+          <button className="btn btn-link" onClick={onHide} ng-disabled="removeContactModal.isLoading">Cancel</button>
+          <button className="btn btn-primary" onClick={onRemoveContact} ng-disabled="removeContactModal.isLoading">
+            {!inProgress && <span>{t(labels.confirm)}</span>}
+            {inProgress && <span
+              role="alertdialog"
+              aria-busy="true"
+              aria-live="assertive">
+              {t('Wait a moment...')}
```
simison: Excellent material for its own component since we seem to be reusing the same pattern elsewhere. :-)

Not a biggie to leave it here, tho.

modules/contacts/client/components/RemoveContact.js
```diff
@@ -0,0 +1,75 @@
...
+        <Modal.Body>
+
+          <p>
+            {t(`${labels.time} {{created, MMM D, YYYY}}`, { created: new Date(contact.created) })}
+          </p>
+
+        </Modal.Body>
+
+        <Modal.Footer>
+          <button className="btn btn-link" onClick={onHide} ng-disabled="removeContactModal.isLoading">Cancel</button>
```
simison: Cancel is missing `t()`

modules/contacts/client/components/ContactPresentational.js
```diff
@@ -0,0 +1,57 @@
...
+      <i className="icon-fw icon-building text-muted"></i>
+      <small>
+        {t('Lives in')}
+        <a href={`/search?location=${contact.user.locationLiving}`}>{contact.user.locationLiving}</a>
+      </small>
+    </div>}
+    {contact.user.locationFrom && <div>
+      <i className="icon-fw icon-home text-muted"></i>
+      <small>
+        {t('From')}
```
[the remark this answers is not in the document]
mrkvon: :+1: 

`t` supports parameters via
```js
t('some string with {{parameter, optionalFormat}}', { parameter: 'city1' })
```
And there is also `react-i18next`'s [`<Trans>`](https://react.i18next.com/components/trans-component) component which simplifies dealing with text that contains html.

This will be supported out of the box. I was just lazy to re-research. :snail: :bear: :cat2: Will fix.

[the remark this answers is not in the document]
simison: We'll have plenty of these cases so would be great to get an example for this in the codebase so that everyone can learn and copy from it. :-)

modules/contacts/client/components/Contact.component.js
mrkvon: #### Variable naming
:+1: 

#### Object vs functional updater

[Both are valid options.](https://reactjs.org/docs/react-component.html#setstate)

The issue with passing object to `this.setState` starts, when we pass object like `{ foo: this.state.foo + 1 }`. We should rather pass `(state) => ({ foo: state.foo + 1 })`.

When we don't reference `this.state` inside `this.setState`, both are fine. Functional updater is my personal preference based on a word of mouth.

simison: >Functional updater is my personal preference based on a word of mouth.

Generally speaking, if you don't pass it a function, you'll give React a chance to optimize setting the state (by letting it batch them) and potentially also improve performance if React doesn't need to re-render between each state-update. 

In situations where you need the previous state, you'll of course need the function to ensure the order of operation.

React is funky! 🙃

agatatalita: As I am always using functional updaters I got curious about the React performance. :)  Do you have any source?
I was taught React try to batch them while passing functions too and official documentation didn't point any difference while talking about it https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous.

modules/contacts/client/components/RemoveContact.js
mrkvon: :+1: 
I'll change that.

FYI: There is also a [runtime extraction](https://react.i18next.com/guides/extracting-translations#3-runtime-extraction), which, when properly set, could save to some api whatever keys were missing. In that context this would work, too. It's on my wishlist todo.

simison: Neat! Probably kinda early to look into stuff like that. :-) For now we should spread good patterns everyone can copy across the codebase.

mrkvon:
Travis-CI tests for Node v11 fail with an unrelated error, npm version mismatch (which is a match, actually).
```
NPM dependencies up to date.
npm: 6.5.0-next.0
Error: Wanted npm version >=6.0.0 (>=6.0.0)
To install npm, run `npm install -g npm@>=6.0.0`
```
:slightly_frowning_face:

modules/contacts/client/components/Contact.component.js
simison: @agatatalita I believe this bit from the docs had me believe that render happens in between synchronous state updates:

>In React, both this.props and this.state represent the rendered values, i.e. what’s currently on the screen.

via https://reactjs.org/docs/faq-state.html

but I'll do more reading and/or testing. :-)

modules/contacts/client/components/RemoveContact.js
mrkvon: 15f8cba5a99783deab9b254792a76a43bb56f99b

modules/contacts/client/components/ContactPresentational.js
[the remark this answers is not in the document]
mrkvon: 15f8cba5a99783deab9b254792a76a43bb56f99b

modules/contacts/client/components/RemoveContact.js
mrkvon: 15f8cba5a99783deab9b254792a76a43bb56f99b

modules/contacts/client/components/Contact.component.js
mrkvon: Changed the variable naming in 15f8cba5a99783deab9b254792a76a43bb56f99b.

The `setState` behaviour still seems to be unresolved.

mrkvon: The [`setState` in React's source](https://github.com/facebook/react/blob/master/packages/react/src/ReactBaseClasses.js#L58-L67). It is passing both the updater object and function to `this.updater.enqueueSetState()`. What is the `this.updater` there?

From the [source code](https://github.com/facebook/react/blob/master/packages/react/src/ReactBaseClasses.js#L26-L27):
> We initialize the default updater but the real one gets injected by the renderer.

Here I lost the thread. We'd have to find the real updater to see how it treats objects and functions passed to `setState`.

simison:
For some reason in this branch, nothing happens when I try to press these icons at my own profile page's navigation on a mobile screen:

image

mrkvon:
> nothing happens when I try to press these icons

@simison is it still the case for you? it seems to work for me (in browser with the tool _toggle device toolbar_ on and simulating touch events; don't have a smartphone to test).

simison:
Rebased

PR #1010, opened 2018-12-23
Chore: upgrade dependencies

simison:
Only major is "i18next": "~12.1.0" to  "~13.0.0"

@mrkvon Anything here you'd see affecting us:

https://github.com/i18next/i18next/blob/master/CHANGELOG.md#1300


simison:
Interesting errors indeed. Rebased, let's see if that does anything.

simison:
And now it passed CI ... 🤷‍♂️ 🤷‍♂️ 🤷‍♂️

PR #1014, opened 2018-12-25
Include user._id in user search API results

simison:
User ID was listed out from User search API results.

Since it's needed to show avatars, we'll need to include it.

Rest of the fields required by avatars are included in `userMiniProfileFields`.


simison:
😁 @mrkvon do you have a patch with updated tests locally as well? Feel free to just push your changes to this branch.

simison:
Suppose we're ready to merge once CI passes.

mrkvon:
Still will fail. The fix is on the way.

simison:
>AssertionError: expected Array [ '_id', 'score' ] to equal Array [ 'score', '_id' ] (at '0', A has '_id' and B has 'score')

😀  Almost!

Issue #1019, opened 2018-12-28
Rewrite user profile in React


mrkvon:
This is a roadmap for rewriting user's profile in React.
All the functionality should survive this.
Design should survive this.

### Routes

```
/profile/:username
/....../........./overview
/....../........./accomodation
/....../........./contacts
/....../........./references
```
#### Routing

Let's implement routing within the profile with e.g. [`react-router`](https://github.com/ReactTraining/react-router).

### Components

Components in Angular to rewrite in React. Break down to subcomponents.

`/modules/users/client/views/profile/` -> `modules/users/client/components/`
- [ ] **[TODO]** `./profile-view.client.view.html`
  - eventually only this one should be imported to AngularJS
- [ ] **[TODO]** `./profile-view-tribes.client.view.html`
- [ ] **[TODO]** wrap #1058 #1059 and #1060 together (completely replace `profile/profile-view-basics.client.view.js`)
- [ ] **[WIP]** `./profile-view-accommodation.client.view.html`
  - [ ] **[WIP]** Offers `/modules/offers/client/views/directives/tr-offer-host-view.client.view.html` (#1045 @agatatalita) 
  - [x] Location (#1023 @mrkvon) 
- [x] **[WIP]** References (@mrkvon) 
- [x] **[DONE]** Avatar and Name for small screen (#1058 @mrkvon)
- [x] Panel for Avatar, Modal with Avatar and ProfileViewBasics (#1060 @mrkvon)
- [x] **[WIP/DONE]** `./profile-view-references.client.view.html` (@mrkvon #918)
- [x] Report Member link (#1059 @mrkvon)

- [x] Contacts in common (#1026 @RGHuynh)
- [x] Contact List (#1017 @mrkvon)
- [x] About `./AboutMe.component.js` (#1005 @agatatalita) 
- [x] Profile View Basics `./ProfileViewBasics.component.js` (#1000 @agatatalita)
- [x] Avatar (#997 @mrkvon)

We can also consider a file structure independent from MEAN.

#### How to get there

First, we need to replace the components one by one and wire them to Angular. (top-down: starting with leaf components and then get deeper to Angular)
Then, we can wire them together to one React (sort of) module with a router and a state.

Alternatively, we can create a feature branch for this and work bottom up. Set up a router and then build leaves.
#### Bootstrap
[...]

PR #1023, opened 2019-01-03
Offer Location in React

mrkvon:
#### Proposed Changes

* Rewrite tr-offer-location in React. (Map in user profile offer.)

#### Testing Instructions

* Open user profile and see that the offer location element works the same
* Open React Developer Tools and see that the element is written in React

Continued work on #1019


simison:
rebased

modules/offers/client/components/OfferLocationPresentational.js
```diff
@@ -0,0 +1,108 @@
...
+    </Map>
+  );
+}
+
+/**
+ * Get url of marker icon
+ */
+function getMarkerUrl(marker) {
+  const markerName = ['marker', 'icon', ...((marker) ? [marker] : [])].join('-');
+  return `/img/map/${markerName}.svg`;
```
simison: This feels unnecessarily complex, especially since `marker` travels through all these functions and components so that at the end it's really hard to tell what it contains. :-/

Would this help?
```js
/**
 * Get url of marker icon
 *
 * @param {String} marker - maybe, meet, yes or an empty string 
 */
function getMarkerUrl(marker) {
  const markerFile = marker ? `-${marker}` : '';
  return `/img/map/marker-icon${markerFile}.svg`;
}
```

modules/offers/client/components/OfferLocationPresentational.js
```diff
@@ -0,0 +1,108 @@
...
+};
+
+/**
+ * Size of the map icon in pixels (bigger for smaller screens)
+ * @param {Number} windowWidth - current width of window
+ * @returns {Number} - icon size
+ */
+function getIconSize(windowWidth) {
+  return (windowWidth < 768) ? 30 : 20;
```
simison: This is okay now in this PR, but just noting that I'm looking forward to using [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) or even something like this helper:

https://github.com/WordPress/gutenberg/tree/0f18024e477bafec475ec3785bb2209562a961c4/packages/viewport#readme

(There are probably more generic ones than the WordPress one)

modules/offers/client/components/OfferLocationPresentational.js
```diff
@@ -0,0 +1,108 @@
...
+function getAriaLabel(marker) {
+  switch (marker) {
+    case 'yes':
+      return 'Yes host';
+    case 'maybe':
+      return 'Maybe host';
+    case 'meet':
+      return 'Meet host';
+    default:
+      return 'Other';
```
simison: This is okay now, but we can loop back later to add translation functions in these.

modules/offers/client/components/OfferLocation.component.js
```diff
@@ -0,0 +1,63 @@
...
+    this.updateWindowDimensions();
+    window.addEventListener('resize', this.updateWindowDimensions);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.updateWindowDimensions);
+  }
+
+  updateWindowDimensions() {
+    this.setState(() => ({ windowWidth: window.innerWidth }));
```
simison: What do you think about adding [`_.throttle`](https://lodash.com/docs/4.17.11#throttle) to this so that we don't end up resizing more than maybe once every 200-500ms?

modules/offers/client/components/OfferLocation.component.js
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    this.handleChangeZoom = this.handleChangeZoom.bind(this);
+    this.updateWindowDimensions = this.updateWindowDimensions.bind(this);
+  }
+
+  /* updating windows size with https://stackoverflow.com/a/42141641 */
+  componentDidMount() {
+    this.updateWindowDimensions();
+    window.addEventListener('resize', this.updateWindowDimensions);
```
simison: This is good here now, just noting for future situations where ResizeObserver might come handy:

https://developers.google.com/web/updates/2016/10/resizeobserver
https://www.npmjs.com/package/resize-observer-polyfill

simison: Oh right, actually not here because we're specifically interested in `windowWidth` only. :-)

simison:
I'm sure we'll generalize this further on once we start using the same component elsewhere in the UI.

modules/offers/client/components/OfferLocation.component.js
mrkvon: added in 0ec4e507a9e258e0cad0b481a421c7607ef6a07f

modules/offers/client/components/OfferLocationPresentational.js
mrkvon: added TODO in 0ec4e507a9e258e0cad0b481a421c7607ef6a07f

modules/offers/client/components/OfferLocationPresentational.js
mrkvon: added TODO in 0ec4e507a9e258e0cad0b481a421c7607ef6a07f

mrkvon:
If the tests pass, I'll merge. :crossed_fingers: 
manual testing works.
Thank you @simison for the detailed review!

PR #1028, opened 2019-01-07
Optimize tribe and message seeding

Jagjeet:
#### Proposed Changes
* Convert tribe and message seeding to ES6
* Add comment blocks to improve file structure
* Improve performance by loading countDocuments() instead of loading entire db structure 
* Add additional comments and modularized code
* Disconnect from the database instead of just exiting

#### Testing Instructions

* Verify tribe seeding and message seeding work as previously 

Fixes #
This is somewhat related to #994 as well, but performance of these scripts was not as bad so the changes ended up being more structural.

### DIFF
--- a/bin/fillTestData/Messages.js
+++ b/bin/fillTestData/Messages.js
@@ -12,7 +12,7 @@ const _ = require('lodash'),
 
 const argv = yargs.usage('$0 <numberOfThreads> <maxMessages>',
   'Seed database with number of threads with up to max messages per thread',
-  function (yargs) {
+  (yargs) => {
     return yargs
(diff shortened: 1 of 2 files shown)


Jagjeet:
The errors here should be resolved once #994 merges. These are the same issues regarding async arrow functions with the `/bin` level eslintrc file. Once we get it resolved there, I will revisit this.

simison:
rebased, eslint issues are gone ☁️  :-)

simison:
Whoooosh! Generating seed data is soooo fast now! 👏

Jagjeet:
Thanks for wrapping this up @simison!

PR #1020, opened 2018-12-30
Add discoverability points for search users UI

simison:
WIP

See https://github.com/Trustroots/trustroots/pull/965#discussion_r244543200

PR #1031, opened 2019-02-02
Update/search improvements

simison:
- Small fix to user search (Got `.finally is not a function` when searching)
- Link to search for various places in the UI

Kinda supersedes #1020

### DIFF
--- a/modules/search/client/components/SearchUsers.component.js
+++ b/modules/search/client/components/SearchUsers.component.js
@@ -67,18 +67,16 @@ class SearchUsers extends React.Component {
     this.setState({ searchQuery: event.target.value });
   }
 
-  async actionSearch(event) {
+  actionSearch(event) {
     event.preventDefault();
     event.stopPropagation();
 
     this.setState({ isSearching: true }, () => {
-      searchUsers(this.state.searchQuery)
-        .then(({ data }) => {
-          this.setState({ users: data });
-        })
-        .finally(() => {
-          this.setState({ isSearching: false });
-        });
+      searchUsers(this.state.searchQuery).then(({ data }) => {
+        this.setState({ users: data, isSearching: false });
+      }).catch(() => {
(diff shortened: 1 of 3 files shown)


modules/search/client/components/SearchUsers.component.js
```diff
@@ -67,18 +67,16 @@ class SearchUsers extends React.Component {
...
+  actionSearch(event) {
     event.preventDefault();
     event.stopPropagation();
 
     this.setState({ isSearching: true }, () => {
-      searchUsers(this.state.searchQuery)
-        .then(({ data }) => {
-          this.setState({ users: data });
-        })
-        .finally(() => {
```
simison: @mrkvon do you know why `.finally` didn't work here?

@Akronix and I were expecting `searchUsers()` to return a standard promise and thus have `finally`. 🤔

mrkvon: One possibility - you have a browser which [supports `Promise`](https://caniuse.com/#feat=promises), but [not `Promise.prototype.finally`](https://caniuse.com/#feat=promise-finally). That would also mean that we don't [transpile Promises](https://stackoverflow.com/questions/38960490/how-can-i-polyfill-promise-with-webpack) and we should.

simison:
Latest Firefox, should support ".finally" I suppose.

modules/search/client/components/SearchUsers.component.js
mrkvon: > Latest Firefox, should support ".finally" I suppose.

I can confirm the error before the fix.

Akronix: I think this is not critical. We can leave it as it is for now. My only concern is that the search members feature works and this isn't there yet. So I'd prioritize to have the indexing working.

Yep, the async was wrongly added to actionSearch, but we already removed it in this PR.

PR #1035, opened 2019-02-10
Mongoose: add logging for autoindexing and indexing errors

simison:
#### Proposed Changes

* Add logging for indexing events in Mongoose models
* Add config to turn autoindexing on or off easily
* Move a couple `mongoose.set` settings directly to the connection config
* Add `ensure-indexes` NPM script to manually ensure indexes in MongoDB
* Add more logging to Mongoose process in general

### Testing

- Run `npm start` in dev mode; you should see lines logged about indexing
- Switch `autoIndex` to ``false` in `config/env/development.js` and you should see _less_ autoindexing events. Some still pop up somehow before the app boots. I'm not sure what causes these — I suspect either Worker or Agenda on their own somehow. Good for another PR another time I'd say.
- Run `npm run ensure-indexes` to index all the schemas, run `npm run ensure-indexes User` to index only User Schema

### DIFF
--- a/bin/db-maintenance/ensure-indexes.js
+++ b/bin/db-maintenance/ensure-indexes.js
@@ -36,9 +36,6 @@ mongooseService.connect(async (connection) => {
 
   const modelNamesToIndex = predefinedModel ? [predefinedModel] : modelNames;
 
-  await mongooseService.ensureIndexes(modelNamesToIndex);
+  await mongooseService.ensureIndexes(modelNamesToIndex).catch(console.error);
   await mongooseService.disconnect();
 });
-
-
-console.log('Done!');
--- a/bin/db-maintenance/ensure-indexes.js
+++ b/bin/db-maintenance/ensure-indexes.js
@@ -28,13 +28,17 @@ if (process.argv[2]) {
 
 mongooseService.connect(async (connection) => {
   await mongooseService.loadModels();
-  const allModels = connection.modelNames();
-  if (!allModels.includes(predefinedModel)) {
-    console.error(`"${predefinedModel}" is not a valid model name. Models: ${allModels.join(', ')}`);
+  const modelNames = connection.modelNames();
+
+  // Validate manually defined model
+  if (predefinedModel && !modelNames.includes(predefinedModel)) {
+    console.error(`"${predefinedModel}" is not a valid model name. Models: ${modelNames.join(', ')}`);
     process.exit(1);
   }
-  const models = predefinedModel ? [predefinedModel] : allModels;
-  await mongooseService.ensureIndexes(connection, models);
+
+  const modelNamesToIndex = predefinedModel ? [predefinedModel] : modelNames;
+
+  await mongooseService.ensureIndexes(connection, modelNamesToIndex);
   await mongooseService.disconnect();
 });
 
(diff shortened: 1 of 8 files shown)


bin/db-maintenance/ensure-indexes.js
```diff
@@ -0,0 +1,46 @@
...
+
+let predefinedModel;
+if (process.argv[2]) {
+  console.log(`Ensuring indexes only for Mongo collection "${process.argv[2]}"`);
+  predefinedModel = process.argv[2];
+} else {
+  console.log('Ensuring indexes for all Mongo collections');
+};
+
+mongooseService.connect(async (connection) => {
```
simison: Hmm, @mrkvon do you know why eslint reports failure for this:

> [lint] /home/travis/build/Trustroots/trustroots/bin/db-maintenance/ensure-indexes.js
> [lint]   29:44  error  Parsing error: Unexpected token =>

I've added the file to migrated -list. Wondering if the separate override for `bin/*` messes something up?

simison:
So with the logging and debugging, I've ensured that on development mode all the indexes get created.

Still not sure why that doesn't happen in production — could be that `NODE_ENV=production` changes something.

Anyway, this is a good change to explicitly turn autoIndexing off for production and just create indexes on demand (with NPM script). It's a recommended way by Mongoose docs: https://mongoosejs.com/docs/connections.html#options (see `autoIndex`)

The separate script to generate indexes is `npm run ensure-indexes` or for specific model: `npm run ensure-indexes User`

bin/db-maintenance/ensure-indexes.js
mrkvon: I found a `/bin/.eslintrc.js` which sets `ecmaVersion: 6`.
After deleting, a few other errors appear, in unrelated files.

In the process I learned `eslint --print-config path/to/file`. It's very useful because it prints the configuration used for the file. (to be used with `npx`).

simison: Good catch!

simison: Removed in https://github.com/Trustroots/trustroots/pull/1042

simison:
I think this is good for a look @mrkvon you wanna see what goes on with our indices. :-)

config/lib/mongoose.js
```diff
@@ -100,19 +138,54 @@ module.exports.disconnect = function (callback) {
...
+          });
+          callback(error);
+        } else {
+          log('info', 'Indexed Mongoose Schema ' + modelName);
+          callback();
+        }
+      });
+    }, function (error) {
+      // if any of the file processing produced an error
+      if (true || error) {
```
mrkvon: `true` leftover from debugging? this will always complain about error.

simison: Yep! Totally. Dealt with in https://github.com/Trustroots/trustroots/pull/1035/commits/14c73511da35cb2b9c91f5c1a18f176bc1b4a782

mrkvon: :) ok. Seems I don't have the most recent version locally.

simison: I mean, I dealt it with after you pointed it out. ;-) You spot it well, I cleaned it out fast. ;-)

mrkvon:
Looking forward to see this run in production.

(What is happening with that text index? :thinking: )

PR #1040, opened 2019-02-17
Chore: update dependencies

simison:


package.json
```diff
@@ -85,7 +85,7 @@
     "angulartics": "~1.6.0",
     "angulartics-google-analytics": "~0.5.0",
     "async": "~2.6.1",
-    "autolinker": "~1.8.3",
+    "autolinker": "~3.0.0",
```
simison: Breaking changes all good 👍 

https://github.com/gregjacobs/Autolinker.js/blob/f7464a5c32aff803880057fb20ab95c1ed29cfd0/README.md#upgrading-from-v2x---v3x-breaking-changes

package.json
```diff
@@ -217,12 +217,12 @@
...
-    "webpack": "~4.28.2",
-    "webpack-cli": "~3.1.2",
-    "webpack-merge": "~4.1.5",
+    "supertest": "~3.4.2",
+    "webpack": "~4.29.4",
+    "webpack-cli": "~3.2.3",
+    "webpack-merge": "~4.2.1",
     "webpack-stream": "~5.2.1",
-    "yargs": "^12.0.5"
+    "yargs": "~13.2.0"
```
simison: All good for yargs https://github.com/yargs/yargs/blob/master/CHANGELOG.md#breaking-changes

package.json
```diff
@@ -154,42 +154,42 @@
     "morgan": "~1.9.1",
     "multer": "~1.4.1",
     "ng-file-upload": "~12.2.13",
-    "nodemailer": "~4.7.0",
+    "nodemailer": "~5.1.1",
```
simison: DNS resolver has some changes, should be good. Will need to double check in production.

package.json
```diff
@@ -154,42 +154,42 @@
...
     "nunjucks": "~3.1.6",
     "passport": "~0.4.0",
-    "passport-facebook": "~2.1.1",
+    "passport-facebook": "~3.0.0",
```
simison: Defaults for FB API version changed, shouldn't affect us.

PR #1041, opened 2019-02-17
Update i18n dependencies

simison:
Updates i18next related deps.

- [x] update modules
- [x] fix breaking changes
- [x] useTranslation hook (partially)
- [x] update i18n documentation

The preferred way of using `react-i18next` should now be via [`useTranslation`](https://react.i18next.com/latest/usetranslation-hook) hook. The higher order components [`withTranslation`](https://react.i18next.com/latest/withtranslation-hoc) (former `withNamespaces`) can be replaced with [`useTranslation`](https://react.i18next.com/latest/usetranslation-hook) in a different PR.


mrkvon:
gee, are they running fast!

2 major versions since a few months.
Seems to fail, will look at.

mrkvon:
Seems that it is a challenge:
https://react.i18next.com/latest/migrating-v9-to-v10

api has changed substantially.

simison:
Seems mostly renamed things?

Good we're not that far in yet, less to upgrade.

simison:
I won't have time to work on this soon so if anyone feels like they wanna take over, feel free to! :-)

mrkvon:
I'll be looking into this. If anybody wants to take over, let me know.

mrkvon:
It would be great if you can review @simison and/or @agatatalita. In particular that the documentation makes sense.

@agatatalita it will be necessary to use the new syntax after this is merged. Can be relevant for you to read. Also the hooks + building components without class is interesting stuff and seems like the proposed future of React.

docs/i18n-legacy.md
```diff
@@ -0,0 +1,52 @@
+# How to provide `t` function to a class component
```
simison: I think we can remove the whole file?

agatatalita:
@mrkvon  I am glad hooks are already in a stable version, just noticed it now. Thanks for mentioning it! Last time I wanted to use them it was still in alpha.

docs/i18n-legacy.md
mrkvon: This is useful for class components. Unless/until move to hooks completely (including for lifecycle methods and state management).

Perhaps instead of _legacy_ it should be called _deprecated_ or _use only sometimes_.

simison: Oh gotcha, just having a notice like what you just wrote would be useful to make it clear 👍

mrkvon: 69c7f6978d07ea2a06750b5e43913de7d686dda1

mrkvon:
Please note @agatatalita @RGHuynh @Akronix @simison and whoever is dealing with Trustroots internationalization that the [i18n API has changed](https://github.com/Trustroots/trustroots/blob/master/docs/i18n.md). All the current works in progress need to be rebased and updated. All the future work should use the new API.

simison:
Thank you!

PR #1043, opened 2019-02-17
i18n: move to config/client folder

simison:
What do you think if we establish a new `config/client` folder which contains config/initialization files meant for client side?

This is so that it would be more clear that `config/lib` is for the server side only and `config/client` for the client side only.

I'd be happy to rename `config/lib` to `config/server` but that's better in another PR so leaving it for now.

#### Proposed Changes

* Move `i18n` config to `config/client`  folder

#### Testing Instructions

* Everything still works and builds fine?


.eslintrc.js
```diff
@@ -172,6 +172,8 @@ module.exports = {
     // ES 2018 - specify migrated files and folders here
     files: [
       'bin/db-maintenance/ensure-indexes.js',
+      'config/env/**',
+      'config/webpack/**',
```
simison: These are really backend code, but they were in the clientside eslint list.

simison:
Oh and I assume at some point we use the same library for backend code, but we'll re-organise at that some point somehow for isomorphic code (shared between client and server side) or have "shared" folder for common parts. No need to optimize for it yet. :-)

.eslintrc.js
simison: On second look (when they failed on linters), they're actually kinda mixture. 🤔 

Reverting this change (better anyway for more focused PR).

simison:
Noting that super important to rebase some branches after this as well to rebase this before merging to ensure there were no new introductions of the old path.

simison:
rebased

PR #1045, opened 2019-02-19
User profile offers react

agatatalita:
### Proposed Changes

  * rewrite user profile offers in React

  * add i18next translation to user profile offers

#### Testing Instructions

  * manual testing: go to `/profile/:username` and observe that everything works the same as before.

  * `i18n` translations should work if data for translations are added to `public/locales/:languagecode` files


modules/utils/filters.js
```diff
@@ -0,0 +1,13 @@
...
+export const limitTo = (text, length) => {
+  return text.substring(0, length);
+};
+
+export const plainTextLength = (text) => {
+  return text && typeof(text) === 'string' ? String(text).replace(/&nbsp;/g, ' ').replace(/<[^>]+>/gm, '').trim().length : 0; // eslint-disable-line angular/typecheck-string
+};
+
+export const sanitizeHtml = (text) => {
+  return sanitizeHtmlFunction(text);
```
simison: This would need to be identical to this:

https://github.com/Trustroots/trustroots/blob/master/modules/offers/server/controllers/offers.server.controller.js#L78

(i.e. `sanitizeOptions`)

modules/offers/client/components/Offers.component.js
```diff
@@ -0,0 +1,374 @@
...
+   * Helper for hosting label
+   */
+  hostingStatusLabel(status) {
+    switch (status) {
+      case 'yes':
+        return 'Can host';
+      case 'maybe':
+        return 'Might be able to host';
+      default:
+        return 'Cannot host currently';
```
simison: will need `t()` wrappers

modules/offers/client/components/Offers.component.js
```diff
@@ -0,0 +1,374 @@
...
+          <a ui-sref="offer.host.edit"
+            className="btn btn-inverse-primary btn-round btn-raised pull-right"
+            aria-label="Modify hosting offer">
+            <span className="icon-edit"></span>
+          </a>
+        }
+
+        {/*  Short descriptions  */}
+        {(offer.description && offer.description.length < 2000) &&
+          <div dangerouslySetInnerHTML={{ __html: sanitizeHtml(offer.description) }}>
```
simison: This will break description texts since default sanitization opens are different from APIs.

public/locales/cs/offers.json
```diff
@@ -0,0 +1,16 @@
+{
+  "Accommodation": "Můj barák",
+  "No guests.": "No guests.",
+  "At most one guest.": "Ne víc jak jeden člověk",
+  "At most {{guestNumber}} guests.": "Ne víc jak {{guestNumber}} lidi.",
+  "Can host": "Može mět lidi",
+  "Might be able to host": "Třeba može mět lidi",
+  "Cannot host currently": "Nemože mět lidi",
+  "Hosting status: {{statusLabel}}": "Hosting status: {{statusLabel}}",
+  "I can host": "Možu mět lidi",
```
mrkvon: :smile:

modules/utils/filters.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const plainTextLength = (text) => {
+  return text && typeof(text) === 'string' ? String(text).replace(/&nbsp;/g, ' ').replace(/<[^>]+>/gm, '').trim().length : 0; // eslint-disable-line angular/typecheck-string
+};
+
+export const sanitizeHtml = (text) => {
+  // TODO solve a problem with sanitize html
+  return text;
+  // return sanitizeHtmlFunction(text, textService.sanitizeOptions);
+};
```
agatatalita: `sanitize-html` was causing errors and `text.server.service` shouldn't be imported from the backend files as @mrkvon noticed so it will be changed to some architecture with a general frontend backend config in this PR.

public/locales/cs/offers.json
mrkvon: I find these translations really funny, but shouldn't we have there the actual official czech language? It's likely to land into production eventually.

As a native czech speaker I offer to update this. :)

.eslintrc.js
```diff
@@ -191,9 +191,11 @@ module.exports = {
       'modules/**/client/components/**',
       'modules/**/client/api/**',
       'modules/**/client/utils/**',
+      'modules/**/offers/components/**',
       'modules/core/client/directives/tr-boards.client.directive.js',
       'modules/core/client/services/photos.service.js',
-      'modules/references/tests/client/**'
+      'modules/references/tests/client/**',
+      'modules/utils/**',
```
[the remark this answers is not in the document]
mrkvon: What is it a module, really?

modules/offers/client/components/OffersPresentational.js
```diff
@@ -0,0 +1,277 @@
...
+      this.setState(() => ({ offerDescriptionToggle: toggleState }));
+    };
+  }
+
+  /* Content Formatting Functions */
+  guestNumberDescription(guestNumber) {
+    const { t } = this.props;
+    switch (guestNumber){
+      case 0: return t('No guests.');
+      case 1: return t('At most one guest.');
```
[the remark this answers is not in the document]
mrkvon: See also [public/locales/cs/contact.json](https://github.com/Trustroots/trustroots/blob/master/public/locales/cs/contact.json)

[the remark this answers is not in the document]
agatatalita: Thanks for the example and the advice!

public/locales/cs/offers.json
agatatalita: ; )

simison:
Rebased

modules/offers/client/components/OffersPresentational.js
```diff
@@ -0,0 +1,276 @@
...
+        </div>
+
+        {/*  The map (React component)  */}
+        {this.renderMap()}
+      </div>
+    );
+  };
+
+  render(){
+    return (this.props.isOwnOffer || this.props.isUserPublic) &&
```
simison: We'll need to ensure return `null` when there's nothing to render, so something like:
```js
    return this.props.isOwnOffer || this.props.isUserPublic
      ? this.renderOffer()
      : null;
```

modules/offers/client/components/OffersPresentational.js
```diff
@@ -0,0 +1,276 @@
...
+        {isOwnOffer && (!offer.status || offer.status === 'no') &&
+        <div className="text-center">
+          <br />
+          <hr className="hr-gray hr-tight hr-xs" />
+          <a href={'/offer/host?status=yes'}className="btn btn-inverse-primary">
+            Start hosting travellers
+          </a>
+          &nbsp;
+          <a href={'/offer/meet'} className="btn btn-inverse-primary">
+            Meet people
```
simison: Needs `t()`

modules/offers/client/components/OffersPresentational.js
```diff
@@ -0,0 +1,276 @@
...
+  hostingStatusLabel(status) {
+    const { t } = this.props;
+    switch (status) {
+      case 'yes': return t('Can host');
+      case 'maybe': return t('Might be able to host');
+      default: return t('Cannot host currently');
+    }
+  }
+
+  getAriaLabel(status) {
```
simison: Since this is used just once and is basically one-liner, I'd just include this in the rendering component instead.

modules/offers/client/components/OffersPresentational.js
```diff
@@ -0,0 +1,276 @@
...
+
+  /* Action Functions */
+  setOfferDescriptionToggle(toggleState) {
+    return () => {
+      this.setState(() => ({ offerDescriptionToggle: toggleState }));
+    };
+  }
+
+  /* Content Formatting Functions */
+  guestNumberDescription(guestNumber) {
```
simison: Since this is used just once and is a simple ternary, I'd inline this in the rendering component too:
```js
{ offer.maxGuests > 0
  ? t('At most {{count}} guests.', { count: offer.maxGuests })
  : t('No guests.')
}
```

modules/offers/client/components/OffersPresentational.js
```diff
@@ -0,0 +1,276 @@
...
+    this.setOfferDescriptionToggle = this.setOfferDescriptionToggle.bind(this);
+    this.state = {
+      offerDescriptionToggle: false,
+      isLoading: true,
+      isMobile: window.navigator.userAgent.toLowerCase().indexOf('mobile') >= 0 || window.isNativeMobileApp
+    };
+  }
+
+  /* Action Functions */
+  setOfferDescriptionToggle(toggleState) {
```
simison: Since we only ever toggle this to `true`, we can maybe simplify to:
```js
toggleLongerDescription() {
  this.setState({ showLongerDescription: true });
}
```

modules/offers/client/less/view.less
```diff
@@ -21,20 +21,21 @@
   margin-top: 10px;
 }
 
-.dropdown-menu {
-  a.offer-hosting-yes:hover {
+// TODO - check places where it is used
```
simison: Just highlighting this todo :-)

modules/utils/filters.js
```diff
@@ -1,7 +1,14 @@
+import sanitizeHtmlFunction from 'sanitize-html-react';
```
simison: I'll move this to `core` module — modules are feature-based so `utils` as its own module feels little wonky.

We've been using the `core` module as "catch-all", tho I agree we could have some better place for common things.

modules/offers/client/components/OffersPresentational.js
nicksellen: I didn't know about returning `null`, and just to add a bit more detail:
- if you are doing conditional rendering inside some jsx, false(y) is still fine (see https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator)
- returning null from a render function causes the whole component not to render at all as if the parent didn't use it (see https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering) - it says this is for _rare cases_. I imagine in normal cases the parent component is in charge of deciding whether to render at all, and the child one would normally render _something_ (which might just be an error/warning message if it has invalid props/state/whatever).

simison: Right on! Not entirely sure how/with what I mixed this up. :-)

.eslintrc.js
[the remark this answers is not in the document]
simison: These could go somewhere under `modules/core/*`

simison:
I extracted separate "read more" component here which can be used here too once merged https://github.com/Trustroots/trustroots/pull/1161

mrkvon:
rebased, fixed comma-dangle and translations

nicksellen:
I'm looking into this, aiming to get it merged. I just:
- merged master into it
- fixed the way of using the location map (passing in specific props instead of the whole offer)
- reformatting
- fixed use of `offer.hosting` -> `offer.status` (not sure what `offer.hosting` used to be)

I didn't switch it to use function components, would prefer to get it merged, and work on that in the future....

modules/offers/client/components/OffersPresentational.js
```diff
@@ -0,0 +1,295 @@
...
+            className="cursor-pointer offer-hosting-no"
+          >
+            {t(`I can't host currently`)} {/* eslint-disable-line quotes */}
+          </MenuItem>
+        </DropdownButton>
+      </ButtonGroup>
+    );
+  }
+
+  renderButtonOther() {
```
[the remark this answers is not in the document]
nicksellen: Yup not sure, I have a vague feeling about reading something about react automatically binding methods, however I think we will just rewrite it into a function component, and that topic will go away.

PR #1046, opened 2019-02-20
Docs: Add F-Droid repository instructions

platschi:
#### Proposed Changes

* Add Info for F-Droid repository with Trustroots mobile app

#### Testing Instructions

* Add repository to your F-Droid installation, search for Trustroots, install app. Verify fingerprints.

Fixes https://github.com/Trustroots/trustroots-expo-mobile/issues/3

### DIFF
--- a/docs/Fdroid.md
+++ b/docs/Fdroid.md
@@ -1,0 +1,23 @@
+# Trustroots App on F-Droid
+
+F-Droid users can manually add additional repositories in order to receive applications. Below repository provides the Trustroots app to F-Droid users who wish to avoid the Play Store version.
+
+## Issue
+
+We don't have resources yet to build a separate "native app" for Android so we're using Expo.io framework instead.
+
+Using Expo seems to block inclusion to the official F-Droid repository. As an alternative, a private repository has been set up to serve F-Droid users with the latest Trustroots Android mobile app.
+
+## Repository details
+
+```html
+Repository address: https://platschi.net/fdroid/repo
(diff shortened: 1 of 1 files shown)


docs/Fdroid.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Issue
+
+We don't have resources yet to build a separate "native app" for Android so we're using Expo.io framework instead.
+
+Using Expo seems to block inclusion to the official F-Droid repository. As an alternative, a private repository has been set up to serve F-Droid users with the latest Trustroots Android mobile app.
+
+## Repository details
+
+```html
```
simison: Maybe just ``` without HTML since there's no HTML to syntax highlight?

simison: Did in https://github.com/Trustroots/trustroots/pull/1046/commits/ea68a5eb728ceb55d99f466c821244f33d3b7ad5

simison:
@platschi Do you think it would be good to explain that notifications won't work in Fdroid version?

platschi:
> Do you think it would be good to explain that notifications won't work in Fdroid version?

Yes, probably a small list of "Limitations" would be helpful (if any). Notifications do not work because the Play Store version uses GCM to push notifications?

simison:
>Notifications do not work because the Play Store version uses GCM to push notifications?

Notifications _alwys_ require you to push them via operating system manufacturer's servers, like Google/Apple for phones and Google, Mozilla, Apple, Microsoft etc for browsers. Any implementation (GCM/Firebase or others) work the same way. Federated notifications don't exist. :-)

PR #1026, opened 2019-01-05
Migrate Contacts in common to React component

RGHuynh:
#### Proposed Changes

* Migrating contacts in common to react component

continue work on #1019

PR #1047, opened 2019-02-21
IETF language codes

mrkvon:
#### Proposed Changes

Use the [IETF language codes](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) for i18next.

#### Motivation
Both [momentjs](http://momentjs.com/docs/#/i18n/) and [i18next](https://jsfiddle.net/jamuhl/3sL01fn0/#tabs=result) support these codes, so the pluralization and momentjs formatting works correctly.

Without this the [pluralization](https://www.i18next.com/translation-function/plurals) doesn't work.

#### Testing Instructions

* Manual: See that language switch still works. Go to a page for which a translation exists (e.g. /volunteering with Czech (česky)) and switch.

#1026 depends on this. We have pluralization there and it doesn't work currently.

* Note: The language switch labels don't matter now. Nor the picked languages. It's a design issue. We can use language codes or whatever. We'll probably list the languages which have a substantial part of the translations done.
This is how it currently looks:
language-switch

#### Challenges

API code uses [different language codes](https://github.com/Trustroots/trustroots/blob/master/config/languages/languages.json). Comparing the [original Trustroots languages](https://github.com/Trustroots/trustroots/blob/master/config/languages/languages_orig.json) and [IANA registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry), it vaguely seems that the IANA uses the shortest available code (e.g. standard `iso_639_1` rather than `iso_639_3`), whereas Trustroots prefer 3-letter codes.
_Solution:_ Since we'll have a limited number of translations for a long time or always, we can just keep the (client language list) and (its map to server language list) as needed.
_Solution:_ Alternatively we can change the server languages to fit the client needs (more, and boring, work).


config/lib/i18n.js
```diff
@@ -33,7 +37,7 @@ i18n
   .use(backend)
   .use(reactI18nextModule) // passes i18n down to react-i18next
   .init({
-    lng: 'eng',
+    lng: languages[0].code,
     // allow keys to be phrases having `:`, `.`
     nsSeparator: false,
     keySeparator: false,
```
simison: Note duplicate `keySeparator`, fixed in https://github.com/Trustroots/trustroots/commit/39037f3f58912af8ddbdf9e77cd31342d0a10353

config/lib/i18n.js
```diff
@@ -33,7 +37,7 @@ i18n
   .use(backend)
   .use(reactI18nextModule) // passes i18n down to react-i18next
   .init({
-    lng: 'eng',
+    lng: languages[0].code,
```
simison: Maybe just export this code as a constant from `i18n-languages.js` and we could then let languages be sorted alphabetically?

config/client/i18n-languages.js
```diff
@@ -0,0 +1,29 @@
...
+export default [
+  { code: 'en', label: 'English' }, // the first language is the default language of the app
+  { code: 'es', label: 'español' },
+  { code: 'fr', label: 'français' },
+  { code: 'ar', label: 'العربية' },
+  { code: 'ru', label: 'русский язык' },
+  { code: 'zh', label: '中文' },
+  { code: 'cs', label: 'česky' },
+  { code: 'fi', label: 'suomi' },
+  { code: 'de', label: 'Deutsch' }
```
simison: Here's an alphabetical list from Atom editor's `sort()`:

česky
Deutsc
English
español
français
suomi
русский язык
العربية
中文

mrkvon: @simison Do you wish to have it sorted alphabetically now? Currently it's vaguely sorted by importance + personal quirks. It's likely going to be reduced to the existent translations only.

But it's probably easier to fix than to talk about. :slightly_smiling_face:

simison: Oh right, since these are just for development anyway,for now, I don't think it matters. Can be as they are now.

config/lib/i18n.js
mrkvon: rebased

config/lib/i18n.js
mrkvon: d7e958a8c2b75fdd0ae7b775eb5df7c28790ae03

mrkvon:
Perhaps merge #1043 first? There might be conflicts which will need to be resolved.

simison:
@mrkvon Merged #1043 👍 

This is good to merge after rebasing methinks.

simison:
rebased

PR #1048, opened 2019-02-21
Fix/code smells

simison:
Cosmetic changes. Good to merge after CI 💚 

Tested also manually by running `npm run seed:users 10`


modules/references-thread/server/controllers/reference-thread.server.controller.js
```diff
@@ -45,24 +45,23 @@ exports.createReferenceThread = function (req, res) {
         },
         'userTo userFrom',
         function (err, thread) {
-
           if (err || !thread) {
             return res.status(400).send({
               message: 'Thread does not exist.'
             });
-          } else if (thread) {
```
simison: `thread` was always truthy

modules/users/server/controllers/users.profile.server.controller.js
```diff
@@ -823,10 +823,10 @@ exports.userByUsername = function (req, res, next, username) {
             return res.status(404).send({
               message: errorService.getErrorMessageByKey('not-found')
             });
-          } else if ((profile && req.user) && req.user._id.equals(profile._id)) {
```
simison: `profile` was always truthy

modules/users/server/models/user.server.model.js
```diff
@@ -8,7 +8,6 @@ var _ = require('lodash'),
     textService = require(path.resolve('./modules/core/server/services/text.server.service')),
     languages = require(path.resolve('./config/languages/languages.json')),
     authenticationService = require(path.resolve('./modules/users/server/services/authentication.server.service')),
-    textService = require(path.resolve('./modules/core/server/services/text.server.service')),
```
simison: duplicate

bin/fillTestData/Users.js
```diff
@@ -202,7 +202,7 @@ var addUsers = function () {
                 // non admin user
                   user.email = index + faker.internet.email();
                   user.password = faker.internet.password();
-                  user.username = index + user.displayName.toLowerCase().replace('\'', '').replace(' ', '');
```
simison: replace without regexp replaces only the first occurrence of the searched string, not all of them.

config/lib/i18n.js
```diff
@@ -36,7 +36,6 @@ i18n
     lng: 'eng',
     // allow keys to be phrases having `:`, `.`
     nsSeparator: false,
-    keySeparator: false,
```
simison: duplicate

PR #1049, opened 2019-02-21
Specify supported locales

mrkvon:
Depends on #1047 

#### Proposed Changes

Specify a list of allowed locales in `/config/shared/locales.json`. These will be the locales supported both by server (user will be able to save only these) and by client (only these will appear in LanguageSwitch).

The locales selection is arbitrary and should be edited as needed.

#### Testing Instructions

Check that

* CI-tests for the server pass
* LanguageSwitch shows the specified languages only


config/client/i18n-languages.js
```diff
@@ -28,6 +29,11 @@ export default [
   { code: 'de', label: 'Deutsch' }
```
simison: What's the purpose of this section if we anyway pull from ` config/shared/locales.json` ?

I.e. why two separate lists? :-)

mrkvon: I think `json` is a nice format for shared config, since it is fully compatible with both `require` (Node) and `import` (React); and we're kind of sure it won't import anything from other files.

1. server doesn't care about `label`. It only needs the code for validation.
2. I didn't want to delete all the unused labels, and json doesn't support comments (not a good enough reason)
3. We need to specify the default language for client (e.g. `config/client/i18n-languages`). (not a good reason to keep the list, just to keep the file)


So the reasons are not very strong. Maybe it's not nice to maintain and we should keep a single shared list in `json` with codes and labels. WDYT @simison?

mrkvon: It's a good idea to keep it in a single place. I'm gonna make a commit with that and we can revert if there is a reason to.

simison: Yeah, I think we could have shared JSON with labels; I agree it's not ideal to keep labels for the backend but it's so much nicer to have just once canonical place and one source of truth that it's okay methinks.

JSON is perfect for this use indeed 👍

simison: Re: JSON not supporting comments:

You could add `README.md` or `LOCALES.md` in the same folder with the json file if you'd like to document it.

config/shared/locales.json
```diff
@@ -0,0 +1,5 @@
+[
+  { "code": "en", "label": "English" },
+  { "code": "cs", "label": "česky" },
+  { "code": "fi", "label": "suomi" }
+]
```
simison: The nice thing about this file is that we could someday in future add languages here that are first visible only in the development environment only before we release them. :-)

config/client/i18n-languages.js
mrkvon: ed43d582b4c6f4f2c622558a2ab3232fc67aeb51

PR #1050, opened 2019-02-23
Travis CI: run only server tests in Node.js 11

simison:
- Since we're running on Node.js 11 containers only to ensure future versions compatibility, we don't need to run linters and client tests in it.
- Rename "travis-ci" NPM script to "lint-and-build-prod" for clarity

This should speed up CI tests a bit since we'll run two containers less.

https://github.com/nodejs/Release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,19 @@
 language: node_js
 node_js:
   - 10
-  - 11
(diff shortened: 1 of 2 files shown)


simison:
image

niiice

PR #1056, opened 2019-02-24
Create endpoint for saving missing translations

mrkvon:
#### Proposed Changes
Save missing translations in translation files.

[Runtime extraction](https://react.i18next.com/guides/extracting-translations#3-runtime-extraction) of translations.

Alternative is [using an extraction tool](https://react.i18next.com/guides/extracting-translations#2-using-an-extraction-tool), e.g. [babel-plugin-i18next-extract](https://github.com/gilbsgilbs/babel-plugin-i18next-extract).

_Note: I have an impression that @nicksellen was researching into the babel extraction plugin, but I haven't found the relevant discussion. Anyways, let's have this PR as a temporary solution which will help us up and running, but will be likely replaced by another (e.g. more convenient) solution later._

When the running app in development mode encounters an untranslated text, it sends a request
```
POST /api/locales/:language/:namespace` with body `{ "Translated Sentence": "Translated Sentence", _t: Time }
```
The backend then adds the missing text to the translation file in `/public/locales/{language}/{namespace}.json`.

The text still needs to be translated. That is not a concern of this PR.

This endpoint is not safe in production! It's intended for development use only. It is enabled in development environment. `config.i18nBackend: Boolean`

#### Testing Instructions

* `git switch save-translations`
* See that the translation strings get saved
  * Run the app in development mode (`npm start`)
  * Switch the language (e.g. to `suomi`)
  * Open some non-localized page e.g. http://localhost:3000/volunteering
  * Observe that the files in `/public/locales/{language}/{namespace}` have changed.
* Run the server tests with `npm run test:server`. The tests should pass.

#### Issues

How to deal with the outdated translation strings? They keep hanging around

### DIFF
--- a/modules/i18n/server/controllers/backend.server.controller.js
+++ b/modules/i18n/server/controllers/backend.server.controller.js
@@ -12,6 +12,12 @@ function enqueue(func, ...args) {
 }
 
 /**
+ * Object property order matters since ES2015
+ * So we can sort it alphabetically by keys
+ */
+const sortObject = object => Object.fromEntries(Object.entries(object).sort(([a], [b]) => a > b ? 1 : -1));
+
+/**
  * Here we execute the standard request, not concerned with waiting
  */
 async function processRequest(req, res) {
@@ -27,11 +33,12 @@ async function processRequest(req, res) {
     await fs.ensureFile(file);
 
     // read current translations or set a default
-    const rawContent = await fs.readFile(file, 'utf8') || '{}';
-    const content = JSON.parse(rawContent);
+    const content = await fs.readJson(file, 'utf8') || '{}';
+
+    const newContent = sortObject({ [key]: value, ...content });
 
     // save the new translation
-    await fs.writeJSON(file, { ...{ [key]: value }, ...content }, { spaces: 2 });
+    await fs.writeJSON(file, newContent, { spaces: 2 });
   } catch (e) {
     // clean up in case of an error
     await fs.writeJSON(file, {}, { spaces: 2 });
--- a/modules/i18n/server/controllers/backend.server.controller.js
+++ b/modules/i18n/server/controllers/backend.server.controller.js
@@ -22,11 +22,11 @@ const sortObject = object => Object.fromEntries(Object.entries(object).sort(([a]
  */
 async function processRequest(req, res) {
   // get the needed variables
-  const { lng, ns } = req.params;
+  const { language, namespace } = req.params;
   const [key] = Object.keys(req.body).filter(key => key !== '_t');
-  const value = (lng === 'en') ? req.body[key] : '';
+  const value = (language === 'en') ? req.body[key] : '';
 
-  const file = path.resolve(`./public/locales/${lng}/${ns}.json`);
+  const file = path.resolve(`./public/locales/${language}/${namespace}.json`);
 
   try {
     // create file if it doesn't exist
--- a/modules/i18n/server/controllers/backend.server.controller.js
+++ b/modules/i18n/server/controllers/backend.server.controller.js
@@ -33,7 +33,8 @@ async function processRequest(req, res) {
     await fs.ensureFile(file);
 
     // read current translations or set a default
-    const content = await fs.readJson(file, 'utf8') || '{}';
+    const rawContent = await fs.readFile(file, 'utf8') || '{}';
+    const content = JSON.parse(rawContent);
 
     const newContent = sortObject({ [key]: value, ...content });
 
--- a/modules/i18n/server/controllers/backend.server.controller.js
+++ b/modules/i18n/server/controllers/backend.server.controller.js
@@ -1,0 +1,46 @@
+const path = require('path');
+const fs = require('fs-extra');
+
+/**
+ * We want to process only one unsaved translation at the time.
+ * Otherwise the output files will be inconsistent.
+ * We make a promise chain: The next promise will start executing after the previous one finished
+ */
+const queue = { promise: Promise.resolve() };
+function enqueue(func, ...args) {
+  queue.promise = queue.promise.then(() => func(...args));
+}
+
+/**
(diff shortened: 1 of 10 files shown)


simison:
Workflow questions:

Wouldn't we accumulate non-needed strings as we keep modifying strings in development mode? As opposite of just having a script which parses all the JS files and picks translations from `t()` functions?

Tho I realise that might not be how this was intended to use, and the API is the way to go. :-) I just personally have experience only from i18n system where you need to do the scraping to produce language files for translation.

mrkvon:
@simison valid points. :+1:

Having a working parser would probably be a preferred way for me, too.

In particular the cumulation of old translations is an issue I'm aware of. I can imagine ways to tackle it, although that is for another discussion and for another time.

I built this because I was repeatedly frustrated with the scraper tools i18next provide. Both [i18next-parser](https://github.com/i18next/i18next-parser) and [i18next-scanner](https://github.com/i18next/i18next-scanner) are out of date, buggy and don't pick the correct namespaces. The project owners reply to issues: we accept PRs, but don't have time for doing anything ourselves.
I guess this is the downside of the community project.

I18next also advertise their commercial product locize.com a lot, and it uses the API method. So they're probably interested in maintaining it. They don't have stakes in the parsers.

I also tested [i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter). It converts to `.po` smoothly, and nice free tools like `poedit` can be used for translating.

I wish to take the path to _make something work_ to know that the translations will be useful eventually. Well, in the ideal world, we'll need an easy, fully automated workflow, which is easy to follow and frustration-free for non-programmers. What is the internationalization good for if there is nobody to translate?

simison:
Your reasoning makes total sense, thanks for elaborating! 👍 

So yeah, static analysis would indeed be the best but sounds like tooling isn't quite there for the time being. We can always iterate on the process. :-)

Lemme know when you'd like this to be reviewed!

modules/i18n/server/controllers/backend.server.controller.js
```diff
@@ -0,0 +1,54 @@
...
+
+  try {
+    // create file if it doesn't exist
+    await fs.ensureFile(file);
+
+    // read current translations or set a default
+    const rawContent = await fs.readFile(file, 'utf8') || '{}';
+    const content = JSON.parse(rawContent);
+
+    const newContent = sortObject({ [key]: value, ...content });
```
simison: Would lodash `sortBy` simplify this? https://lodash.com/docs/4.17.15#sortBy

modules/i18n/server/controllers/backend.server.controller.js
```diff
@@ -0,0 +1,54 @@
...
+ * So we can sort it alphabetically by keys
+ */
+const sortObject = object => Object.fromEntries(Object.entries(object).sort(([a], [b]) => a > b ? 1 : -1));
+
+/**
+ * Here we execute the standard request, not concerned with waiting
+ */
+async function processRequest(req, res) {
+  // get the needed variables
+  const { lng, ns } = req.params;
```
simison: `lng` is also used for "longitude" but I suspect it's the value used by default in the framework? Otherwise I'd appreciate using writing it out as `language` or `lang`. Not a big concern, just a thought.

modules/i18n/server/controllers/backend.server.controller.js
```diff
@@ -0,0 +1,54 @@
+const path = require('path');
+const fs = require('fs-extra');
+
+/**
+ * We want to process only one unsaved translation at the time.
+ * Otherwise the output files will be inconsistent.
+ * We make a promise chain: The next promise will start executing after the previous one finished
+ */
+const queue = { promise: Promise.resolve() };
```
simison: I've been using [`async.queue`](https://caolan.github.io/async/v2/docs.html#queue) helper for these but delighted to see how few lines custom solution is. :-)

nicksellen:
Great progress!

I also read the [localization tutorial](https://github.com/Trustroots/trustroots/blob/docs/localization/docs/l10n.md) and was going to comment on there, but seems more appropriate here.

If we extract to json then convert the json to gettext, we presumably cannot make use of any of the extra features that gettext provides (looking at https://docs.weblate.org/en/weblate-3.10.3/formats.html#translation-types-capabilities for the extra features).

Perhaps the backend can write .po files directly, and potentially include extra context? Comments and source location seem potentially useful, at least this is something that we ran into with karrot, where the strings sometimes lack context, and the translation ends up not quite right.

modules/i18n/server/controllers/backend.server.controller.js
mrkvon: Thank you for pointing out potentially simpler solutions.

Unfortunately `sortBy` won't help us here. The function works on objects, but seems to take only values into consideration, and returns an array. And we want to sort by keys.

If I misunderstood please kindly help me understand. :slightly_smiling_face:

modules/i18n/server/controllers/backend.server.controller.js
mrkvon: The [i18next documentation](https://www.i18next.com/overview/configuration-options#languages-namespaces-resources) uses `lng` and `ns`. However there is no need for us to follow that convention in our backend code. I've also replaced `ns` with more clear `namespace`.

66f2a3de28b949edfca502adcd7c941927a2a1b9

mrkvon:
@nicksellen That's an interesting problem, how to add more context to the translations.

Unfortunately, by default, with runtime extraction that we do here, the app sends missing key, value and time only. There is no meta information included, only namespace.
There are some additional gettext [tools](https://www.i18next.com/overview/plugins-and-utils) like gettext backend and po loader, but they're either not implemented or not maintained. I don't know how to progress in that.

I've linked the comment in #492 so we don't lose it.

mrkvon:
@nicksellen @simison What additional changes would you like to see before approving this PR?

PR #1059, opened 2019-02-28
Migrate "Report Member" link to React

mrkvon:
#### Proposed Changes

* Migrate `./modules/support/client/directives/tr-report-member.client.directive.js` to React.
* We keep the AngularJS directive, because it's used in multiple not-yet-migrated places

#### Testing Instructions

* Open the user profile (other than self) and see:
  - there is still a Report Member link under the profile basic view
  - there is still a Report Member link in `/messages/:username`
  - the link redirects to support form with reported user's name filled
  - the title shows as expected
  - see that i18n works (but we don't have translations to test this)

Continuous work on #1019


simison:
> We keep the AngularJS directive, because it's used in multiple not-yet-migrated places

I see just one?

https://github.com/Trustroots/trustroots/blob/15a8bcebee8286eca1591f950a120420b2e08dd3/modules/messages/client/views/thread.client.view.html#L201

https://github.com/Trustroots/trustroots/search?q=tr-report-member&unscoped_q=tr-report-member

modules/support/client/components/ReportMemberLink.component.js
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import '@/config/client/i18n';
+import { useTranslation } from 'react-i18next';
+
+export default function ReportMemberLink({ username }) {
+  const { t } = useTranslation('support');
+
+  return (
+    <span title={t('Report member {{username}} to support', { username })}>
```
simison: Why not just add this `title` to `a` tag?

simison: Oh I see, you're following the rendered output from Angular:
```
<small ng-if="::app.user._id !== profileCtrl.profile._id" tr-report-member="::profileCtrl.profile.username" title="Report member dalpus to support"><a ui-sref="support({report: username})" class="text-muted" href="/support?report=dalpus">Report member</a></small>
```

That's just unfortunate how directives work. We can make it better in React. :-)

mrkvon:
> I see just one?
Indeed, I thought I saw another 2 places, but one of them was in the original file comment.
Probably worth replacing immediately.

simison:
>Probably worth replacing immediately

Yup, tho not a blocker for this PR at all. :-)

modules/support/client/components/ReportMemberLink.component.js
mrkvon: removed unnecessary tags and moved `title` => `aria-label` in 2424f57fd4d27e9edee34bbe6fd34af987b5bb42

PR #1060, opened 2019-02-28
Migrate panel with avatar and basic user info to React

mrkvon:
#### Proposed Changes

Build the left panel in user profile. With avatar and basic user info. This component wraps Avatar and ProfileViewBasics.
Also takes care of the modal with bigger avatar image.

#### Testing Instructions

* See that the left panel hasn't changed during the migration.
* Click the avatar and a modal should show. The modal should close in standard ways.
* See that it behaves as expected on a small screen, too.

Continued work on #1019

### DIFF
--- a/modules/users/client/components/ProfileOverview.component.js
+++ b/modules/users/client/components/ProfileOverview.component.js
@@ -1,0 +1,50 @@
+/**
+ * A panel with basic user info in user profile.
+ * It wraps Avatar, a Modal with Avatar and ProfileViewBasics components.
+ * @param {Object} profile - displayed user's profile data
+ */
+
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import Avatar from './Avatar.component';
+import ProfileViewBasics from './ProfileViewBasics.component';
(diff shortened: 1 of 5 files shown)


mrkvon:
I believe this can be (reviewed and )merged.

modules/users/client/components/ProfileOverview.component.js
```diff
@@ -0,0 +1,57 @@
...
+      </div>
+    </div>
+
+    {/* modal with avatar */}
+    <Modal
+      show={isAvatarModalOpen}
+      onHide={() => setIsAvatarModalOpen(false)}
+      className="modal-avatar"
+    >
+      <Avatar
```
simison: Now when image is opened, it cannot be closed by re-clicking. It can only be closed by clicking at the area outside the image:

Screenshot 2019-11-10 at 20 51 22

Would just adding `onClick={handleClose}` in `Avatar` work? Or wrapping it in `<button>`?

mrkvon: Good catch! :+1:
fixed with  68f5a0e

PR #1061, opened 2019-03-01
Migrate bottom navigation for small screens to React

mrkvon:
#### Proposed Changes

Bottom Navigation for small screens
there is an issue with binding of contacts between angular and react, and a hacky solution

#### Testing Instructions

* See that the bottom navigation on small screens works (almost) the same as before
* Test that active tab is highlighted
* Test that a correct number of Contacts is shown. When user is not self and doesn't have contacts, Contacts tab should not be shown.

Continuous work on #1019

### DIFF
--- a/modules/users/client/components/BottomNavigationSmall.component.js
+++ b/modules/users/client/components/BottomNavigationSmall.component.js
@@ -1,0 +1,70 @@
+/**
+ * Bottom navigation for small screens
+ */
+
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+import classNames from 'classnames';
+
+export default function BottomNavigationSmall({ username, isSelf, contactCount }) {
+  // @TODO the default value for 'active' will need to be fetched in more React way
+  // when we have a router in place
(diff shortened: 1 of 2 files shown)


mrkvon:
IMO this can be merged, preferably after a review.

modules/users/client/components/BottomNavigationSmall.component.js
```diff
@@ -0,0 +1,70 @@
...
+      label: t('Hosting'),
+      link: `/profile/${username}/accommodation`
+    }
+  ];
+
+  // contacts tab is shown only sometimes
+  if (contactCount > 0 || isSelf) {
+    tabs.push({
+      key: 'contacts',
+      label: <>{t('Contacts')} <span className="badge">{contactCount}</span></>,
```
simison: This is fine to merge as is, just noting that another way would be to include `count: contactCount` and move the HTML to `return`

mrkvon: Interesting. I didn't know that the span was ignored without content. I like this solution, as it separates label from count (which are two different things).

PR #1062, opened 2019-03-03
Top user profile navigation for small screens - migrate to React

mrkvon:
#### Proposed Changes

* Migrate top profile navigation for small screens to React.

top-navigation-small

#### Testing Instructions

* Go to somebody's profile
* Make the browser smaller
* Check that the top profile navigation hasn't changed in looks and behaviour.
  - especially check that the correct links are displayed and that they behave as they did before
    - own profile: Edit your profile
    - not a contact: Send a message | Add contact
    - some of my contacts: Send a message | Remove contact
    - a contact request I sent: Send a message | Delete contact request
    - a contact request I received: Send a message | Delete contact request
    - additionally, when not own profile && references are enabled: Write a reference

Work on #1019

### DIFF
--- a/modules/users/client/components/TopNavigationSmall.component.js
+++ b/modules/users/client/components/TopNavigationSmall.component.js
@@ -1,0 +1,142 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { OverlayTrigger, Tooltip } from 'react-bootstrap';
+
+export default function TopNavigationSmall({ isSelf, username, contact, areReferences, isResolved }) {
+
+  let links;
+  console.log('*********************************', isSelf, username, contact, areReferences, isResolved);
+
+  if (isSelf) {
+    links = [
+      {
+        label: 'Edit your profile',
+        link: '/profile/edit'
+      }
+    ];
+  } else {
+    links = [
+      {
+        label: 'Send a message',
+        link: `/messages/${username}`
+      }
+    ];
+
+    if (areReferences) {
+      links.push({
+        label: 'Write a reference',
+        link: `/profile/${username}/references/new`
+      });
+    }
+
+    if (isResolved) {
+      const contactLink = (!contact._id) ?
+        {
+          label: 'Add contact',
+          link: `/messages/${username}`
+        }
+        :
+        (contact.confirmed) ?
+          {
+            label: labelWithTooltip({ label: 'Remove contact', tooltip: 'Contacts since {{ ::profileCtrl.contact.created | date:\'mediumDate\''}),
+            link: `/messages/${username}`,
+            tooltip: 'Contacts since {{ ::profileCtrl.contact.created | date:\'mediumDate\''
+          }
+          :
+          {
+            label: 'Delete contact request',
+            link: `/messages/${username}`,
+            tooltip: 'Request sent {{ ::profileCtrl.contact.created | date:\'mediumDate\' }}'
+          };
+
+      links.push(contactLink);
+    }
+  }
+
+
+
+  return (
+    <nav className="navbar navbar-white navbar-fixed-top navbar-fixed-top-below visible-xs-block">
+      <div className="container">
+        <ul className="nav navbar-nav" role="navigation">
+          {links.map(({ label, link }, index) => (
+            <li key={index}><a href={link}>{label}</a></li>
+          )
+          )}
+        </ul>
+      </div>
+    </nav>
+  );
+}
+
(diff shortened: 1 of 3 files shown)


modules/contacts/client/components/RemoveContact.js
```diff
@@ -32,10 +40,10 @@ export function RemoveContact({ t, contact, show, inProgress, onHide, onRemoveCo
   const labels = situationLabels[situation];
 
   return (
-    <Modal show={show} onHide={onHide}>
+    <Modal show={show} onHide={onCancel}>
       <div className="modal-content">
         <Modal.Header>
-          <button type="button" className="close" aria-hidden="true" onClick={onHide} ng-if="!removeContactModal.isLoading">&times;</button>
+          <button type="button" className="close" aria-hidden="true" onClick={onCancel} ng-if="!removeContactModal.isLoading">&times;</button>
```
simison: Remember to remove `ng-if` Angular bits. :-)

mrkvon: Thank you @simison! :+1:
Fixed in a98e3403d544775fdae6d8e06074382cdabc9a46 as part of #1121.

modules/users/client/controllers/profile.client.controller.js
```diff
@@ -23,11 +23,22 @@ function ProfileController(
   vm.contact = contact;
   vm.contacts = contacts;
 
-  // Exposed to the view
-  vm.hasConnectedAdditionalSocialAccounts = hasConnectedAdditionalSocialAccounts;
-  vm.isConnectedSocialAccount = isConnectedSocialAccount;
-  vm.socialAccountLink = socialAccountLink;
-  vm.isWarmshowersId = isWarmshowersId;
```
mrkvon: These are unrelated to the migration itself. Unused variables hanging around.

mrkvon:
I'll be grateful if we can get this old PR finished. I've rebased and fixed some bugs. Would be great if somebody can review and/or approve so we can merge. Thanks!

modules/users/client/components/TopNavigationSmall.component.js
```diff
@@ -0,0 +1,173 @@
...
+  username: PropTypes.string.isRequired,
+  selfId: PropTypes.string.isRequired,
+  userId: PropTypes.string.isRequired,
+  contact: PropTypes.object,
+  referencesEnabled: PropTypes.bool.isRequired,
+  isResolved: PropTypes.bool.isRequired,
+  onContactRemoved: PropTypes.func.isRequired,
+};
+
+function LabelWithTooltip({ id, label, tooltip }) {
```
simison: I think it's fine to leave tooltips out from mobile navigation; they never get shown since there's no _hover_ on mobile devices.

image

That'll simplify this a bit.

mrkvon: Is the small screen unlikely to be used on desktop?
Also, the tooltip is present in the original code. i.e. this change will degrade/simplify the original.
@simison If you're happy with the above points, I'll proceed with the change. I would appreciate a confirmation.

simison: I think we can consider everything under 768px as a touch screen, and very unlikely used on a desktop.

If it was on original code it's my bad. 😅 Let's clean out now that we have a chance while looking into it.

mrkvon: Removed the tooltip in ad97da390e46b965c2a789f0d030eb87dd361388.

modules/users/client/components/TopNavigationSmall.component.js
```diff
@@ -0,0 +1,123 @@
...
+export default function TopNavigationSmall({
+  username,
+  contact,
+  referencesEnabled,
+  isResolved,
+  selfId,
+  userId,
+  onContactRemoved,
+}) {
+  // @TODO this hacky fix should be removed and the data should be consistent
```
simison: This could warrant opening an issue as a follow up for some day.

mrkvon:
Thank you, @simison for the collaboration!

modules/users/client/components/TopNavigationSmall.component.js
mrkvon: Maybe a reminder to review all those migration-dependent TODOs... seems there will be more as we progress.

I love Typescript, because it helps so much with fixing inconsistencies like this one...

Issue #919, opened 2018-12-06
Report a member is invisible on small screens.


mrkvon:
**Describe the bug**
The Report member button is invisible on small screens. It's hidden under the bottom navigation in `/profile/:username`. It's not shown at all in messages.

**To Reproduce**
1. Go to https://trustroots.org/profile/:username.
2. Make the window with a browser small or use a mobile.
3. Scroll to the bottom.

**Expected behavior**
Report member is available on small screens.

**Screenshots**
![screenshot_2018-12-06_09-20-01](https://user-images.githubusercontent.com/7449720/49571271-1d9bc500-f939-11e8-8445-354a71acf74b.png)
Yes, it really is hiding behind there.

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser: Firefox
 - Version 62.0.2 64-bit

mrkvon:
We may need a placeholder for the navigation at the bottom of the screen. The same element with visibility:hidden and position:relative, or something like that.

simison:
>We may need a placeholder for the navigation at the bottom of the screen

Extra `padding` or `margin` somewhere should do the trick and avoids adding new DOM elements. 👍

PaulElis:
In the 'modules/users/clients/views/profile/profile-view.client.view.html' line 137 we have a class name 'hidden-xs' which makes the 'Report member' invisible on smaller screens. This is a wrapper for the 'Report member'.

Line 137:
`      <div class="col-sm-3 hidden-xs" ng-include="'/modules/users/views/profile/profile-view-basics.client.view.html'"></div>
`

PaulElis:
@mrkvon @simison
What Less file would be best to make changes to the padding?

mrkvon:
@PaulElis I haven't done any styling with less for Trustroots. Anyways, it seems to me that `modules/users/client/less/profile-view.less` may be a good spot. There is a class `.profile-view` which is our container in question.

If you feel comfortable with media queries, it would be nice if the padding is higher only for the small display, where the bottom navigation is shown. :slightly_smiling_face:

PR #1063, opened 2019-03-12
added a.text-muted padding-bottom

PaulElis:
Issue #919

@mrkvon @simison


petr-hajek:
With this fix it has 2 scrollbars
image

PR #1064, opened 2019-04-04
remove displayUsername

### LINKED ISSUE #860: changing username doesn't work for a profile, but it works after loading and saving the data from the profile
**Describe the bug**

User can't change their username. Has never changed their username.

**To Reproduce**

1. Create profile with profile creation date 2016-12-22.
2. Try to change username.
3. Can't change username. 

After changing date (locally, in database) to 2017-12-22 it is possible to change username.

**Additional context**
https://trustroots.zendesk.com/inbox/conversations/2268 (only accessible for those with access)

tg-x:
#### Proposed Changes

* remove displayUsername

it does not do any db migrations, perhaps a reversed version of [add-user-displayUsername.js](https://github.com/Trustroots/trustroots/blob/d887dc3aa3c74014407a692c44c3db1a6d28edd9/migrations/archived/1429124257492-add-user-displayUsername.js) would be necessary?

#### Testing Instructions

Completely untested, asking for review.

Fixes #860

### DIFF
--- a/modules/contacts/server/controllers/contacts.server.controller.js
+++ b/modules/contacts/server/controllers/contacts.server.controller.js
@@ -456,7 +456,6 @@ exports.contactListByUser = function (req, res, next, listUserId) {
           updated: '$user.updated',
           displayName: '$user.displayName',
           username: '$user.username',
-          displayUsername: '$user.displayUsername',
           avatarSource: '$user.avatarSource',
           avatarUploaded: '$user.avatarUploaded',
           locationFrom: '$user.locationFrom',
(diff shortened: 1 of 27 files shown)


tg-x:
cc @simison

simison:
Thank you!

simison:
>it does not do any db migrations, perhaps a reversed version of add-user-displayUsername.js would be necessary?

the query to remove these fields in production DB is very simple so no need migrations here, and leaving these to development databases is not a big deal. Thanks!

PR #1066, opened 2019-04-28
Update dependencies

simison:
#### Proposed Changes

* Update dependencies

#### Testing Instructions

* CI & manual poking around


package.json
```diff
@@ -34,7 +34,7 @@
     "ensure-uploads-dir-exists": "node ./bin/ensure-uploads-dir-exists.js",
     "fontello": "gulp fontello",
     "install-dependencies": "node ./bin/install-deps.js",
-    "lint:watch": "esw --cache --watch .",
+    "lint:watch": "esw --cache --watch --quiet .",
```
simison: The new version now supports `--quiet` which gets rid of repeated "all good" messages.

config/webpack/webpack.config.js
```diff
@@ -31,8 +31,9 @@ module.exports = merge(shims, {
           options: {
             presets: [
               ['@babel/preset-env', {
-                useBuiltIns: 'usage',
-                modules: 'commonjs'
+                corejs: 2,
```
simison: The new version now asks for defining `corejs` and it defaults to 2, but that might change with future versions to 3.

Should be pretty trivial to update to 3 as well but that's another PR.

PR #1071, opened 2019-05-06
Remove meta links

doktorhorst:
Commented out links to meta, since it is already down for several months. Replaced it by a link to the page linking to slack where useful.

### DIFF
--- a/docs/Support-Getting-Started.md
+++ b/docs/Support-Getting-Started.md
@@ -1,3 +1,3 @@
 # Getting started with support
 
-We have many support requests to handle. Check out the [meta thread](https://meta.trustroots.org/t/wiki-support-team/22) to get involved.
+We have many support requests to handle. Introduce yourself in the support channel of our [volunteer chat](Chat.md) to get involved. <!-- Check out the [meta thread](https://meta.trustroots.org/t/wiki-support-team/22) to get involved. -->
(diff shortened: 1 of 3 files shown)


doktorhorst:
> Looks good! Feel free to merge.
Should I choose 'squash' or 'rebase'? It's my first PR :flushed: 
> We should update also links at https://www.trustroots.org/team but that can be a separate PR.
To be honest, I would replace all the links at the top by one "Get active!" directly linking to https://team.trustroots.org/Volunteering.html, since the different volunteering topics are listed there anyway. Agree?

simison:
> Should I choose 'squash' or 'rebase'? 

Squashing is better since it leaves one clean commit vs many.

Congraz on first one. Well done!

> I would replace all the links at the top by one "Get active!" directly linking to https://team.trustroots.org/Volunteering.html, since the different volunteering topics are listed there anyway

good idea!

simison:
Merge went well. :-) You can now hit "delete the branch", too.

Hm, looks like the site generator didn't like those (it's Markdown but it's kinda random how things end up at the site) 

https://team.trustroots.org/

image

Mind remove those completely in the same PR with changes to `/team` page's links?

doktorhorst:
Oh, damn. Will do that.

Issue #1067, opened 2019-04-28
"Hosting offer map" uses Open street map instead of Mapbox


simison:
In the user's profile, when looking at their hosting profile, the map that loads is openstreet map:

<img width="437" alt="Screenshot 2019-04-28 at 19 15 55" src="https://user-images.githubusercontent.com/87168/56867130-72280580-69ea-11e9-954b-fa9118c9b93a.png">

When Mapbox is configured, it should load Mapbox layers instead. Here's a screenshot of same area from `/search`:

<img width="1192" alt="Screenshot 2019-04-28 at 19 16 38" src="https://user-images.githubusercontent.com/87168/56867133-7f44f480-69ea-11e9-835f-629d5cbc060a.png">

OSM layer is used just for development without Mapbox key and they themself also ask not to use these layers in production, as it just hammers their servers needlessly.

Likely introduced in https://github.com/Trustroots/trustroots/pull/1023

PR #1083, opened 2019-05-18
Centralize React map layer config to a component

simison:
#### Proposed Changes

* Change default config layer from our custom Mapbox layer to Mapbox's globally provided one: https://www.mapbox.com/maps/streets/ — already updated to all other maps in production via configuration
* Conditionally load either OSM or Mapbox in a component that provides map layers for React components (and hence for React-Leaflet)

#### Testing Instructions

See hosting  offer map on user profile:

image

Previously it would show up OSM map, which isn't very pretty nor it's supposed to be used in production sites: 

image

Fixes https://github.com/Trustroots/trustroots/issues/1067

### DIFF
--- a/config/env/local.sample.js
+++ b/config/env/local.sample.js
@@ -69,10 +69,17 @@ module.exports = {
   /*
   mapbox: {
     maps: {
+      // Old Trustroots main map layer (2014–05/2019)
+      //streets: {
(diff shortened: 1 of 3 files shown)


simison:
Thanks! In future, we can use this same component with other maps in the app and add more layers to it.

Issue #1080, opened 2019-05-15
Importing CouchSurfing profiles


guaka:
**Is your feature request related to a problem? Please describe.**
For new members who have complete couchsurfing profiles: some of these will be very happy to import their data from couchsurfing.

**Describe the solution you'd like**
CS data export there is currently possible at /preferences/privacy
It's a zip file. We can write some code to import the JSON and JPGs and automatically fill in the user's profile.

rumwerfer:
With the recent paywall decision: Can anyone with a CS account answer the question if they let you download your data now if you didn't pay them? (they'd probably have to for EU users at least)

Akronix:
I thought to send them an email about requesting my data, but it'd be interesting to know if someone has tried already.

simison:
We have some code for the importer written here: https://github.com/Trustroots/trustroots/pull/1086

simison:
Plenty more code here: https://github.com/simison/couchspinner/

Now this mostly would need bringing that over, and adding some UI.

PR #1086, opened 2019-05-19
CouchSurfing importer

simison:
WIP

See https://github.com/Trustroots/trustroots/issues/1080

Should be split into separate tasks and merged independently, for example;
- backend importer, perhaps start simple with only description but leave out images and other more complex stuff for now
- writing imported data to user profile
- adding simple UI in settings
- adding more guidance in UI for instructing how to export profile in CS
- adding UI in signup flow
- adding options to append or rewrite profile?
- writing tests for all of the above as you go :-)

#### Proposed Changes

*

#### Testing Instructions

*

Fixes #

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,4 @@
-const semver = require('semver');
+Ωconst semver = require('semver');
 
 // Converts semver range `~16.6.0` to strict version `16.6.0`
 const reactVersion = semver.coerce(require('./package.json').dependencies.react)
--- a/modules/users/server/controllers/users.import.server.controller.js
+++ b/modules/users/server/controllers/users.import.server.controller.js
@@ -1,13 +1,16 @@
+// const fs = require('fs');
 const path = require('path');
 const errorService = require(path.resolve('./modules/core/server/services/error.server.service'));
 const fileUpload = require(path.resolve('./modules/core/server/services/file-upload.service'));
+const decompress = require('decompress');
+const decompressUnzip = require('decompress-unzip');
 
 // log = require(path.resolve('./config/lib/logger')),
 
 /**
- * Middleware to validate+process avatar upload field
+ * Middleware to validate+process import file upload field
  */
-exports.csImportUploadField = function (req, res, next) {
+exports.csImportUploadField = (req, res, next) => {
   if (!req.user) {
     return res.status(403).send({
       message: errorService.getErrorMessageByKey('forbidden')
@@ -15,4 +18,31 @@ exports.csImportUploadField = function (req, res, next) {
   }
 
   fileUpload.uploadFile(['application/zip'], 'import-file', req, res, next);
+};
+
+/**
+ *
+ */
+exports.processCsImport = async (req, res) => {
+
+  // eslint-disable-next-line
+  console.log(req.file.path);
+
+  try {
+    // @TODO: Need to have some good destination folder here
+    await decompress(req.file.path, './tmp/unzipped', {
+      plugins: [
+        decompressUnzip()
+      ]
+    });
+  } catch (err) {
+    return res.status(400).send({
+      message: errorService.getErrorMessage(err)
+    });
+  }
+
+  // fs.unlink(req.file.path);
+  res.send({
+    message: 'Import file processed.'
+  });
 };
--- a/modules/users/server/controllers/users.import.server.controller.js
+++ b/modules/users/server/controllers/users.import.server.controller.js
@@ -1,0 +1,18 @@
+const path = require('path');
+const errorService = require(path.resolve('./modules/core/server/services/error.server.service'));
+const fileUpload = require(path.resolve('./modules/core/server/services/file-upload.service'));
+
+// log = require(path.resolve('./config/lib/logger')),
+
+/**
+ * Middleware to validate+process avatar upload field
+ */
+exports.csImportUploadField = function (req, res, next) {
+  if (!req.user) {
+    return res.status(403).send({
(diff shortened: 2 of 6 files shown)


modules/users/server/controllers/users.import.server.controller.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ */
+exports.processCsImport = async (req, res) => {
+
+  // eslint-disable-next-line
+  console.log(req.file.path);
+
+  try {
+    // @TODO: Need to have some good destination folder here
+    await decompress(req.file.path, './tmp/unzipped', {
```
simison: since the filename is random, maybe just `${filename}-output` as the directory name where files can be unzipped. Then after the process is done, delete both; temp file and the folder.

modules/users/server/controllers/users.import.server.controller.js
```diff
@@ -0,0 +1,48 @@
...
+      plugins: [
+        decompressUnzip()
+      ]
+    });
+  } catch (err) {
+    return res.status(400).send({
+      message: errorService.getErrorMessage(err)
+    });
+  }
+
```
simison: Next up: 
- Read Json file
- Read profile from json file
- Write profile contents from the file to the current user's profile (append or rewrite?)

simison:
rebased

.eslintrc.js
```diff
@@ -1,4 +1,4 @@
-const semver = require('semver');
+Ωconst semver = require('semver');
```
simison, line 1: 🦆

guysoft:
Hey question - why are you asking the users to upload a json file which they have a hard time getting, if you could just scrape the data off their page which is publicly available?
It might be an issue for it changing, but it has not done that in over a few good years.
You can even figure out who their friends are etc.

simison:
>Hey question - why are you asking the users to upload a json file which they have a hard time getting, if you could just scrape the data off their page which is publicly available?

Hi, thanks for asking! A few reasons:

- CS already provides a feature to get this zip/json in canonical way:
   image
- Above canonical way is much less likely to change over time, vs the site or APIs are likely changing a lot. We'd have to 1) monitor and detect, and then quickly act on any breakage.
- CS has historically blocked any such scrapers (servers and users) as accessing their site in such way goes against their Terms of use (and it's very easy for them to block such access by IP)
- Profiles might be public or not depending on user settings, so it would be more complicated to build and maintain something that goes around that limitation. We'd also need to have an authenticated user to access the site or ask for the user's username+password, which at leats I'm not comfortable asking for security-reasons.

Does that answer your question?

The most ideal setup would be that CouchSurfing would let us connect profiles to their site via oAuth or similar mechanism, much as we can connect to Twitter or Facebook. That way we could access their APIs in an authenticated manner.

simison:
One more bonus; if other hospitality exchange networks format their export-data in a similar format, we could have a cross-network way to import/export your hospitality exchange profiles.

guysoft:
Hey, a change that might be relevant since this PR was proposed last year is, that you can't get that JSON anymore unless you pay CS. Although this is somewhat less expected, when this PR was made, most users can't download this zip anymore. The url  https://www.couchsurfing.com/preferences/account redirects now to https://www.couchsurfing.com/contribution .

So in my case, for example, and likely most users in CS now, you can't get your profile zip file, but you can see the profile using the url https://www.couchsurfing.com/people/[USERNAME] .

simison:
Yep, that's a little unfortunate indeed. I believe it's still possible to request the data via form (can't find the link for that but I'll look it up.)

I feel like going on with anything else than this would make this much bigger project than it is fun to maintain over time, because of points I wrote out above. So I'm still in favor of just proceeding with a simpler feature where we process the export. That way we'll get some sense/info on the usage of this already and can think of next more complicated steps if need.

Much of the code and designs for this feature would be useful for any more complex system as part of it anyway.

Does that make sense?

guysoft:
Makes sense. I made a data request myself by emailing support.  We have a group of people who did. We got an automated message that it might take up.to 30 days. Will uodate, but that kinda seen not as great. I went and copy-pasted my profile in the end manually. 30 days is too much.

So what is required now for this to be pulled?

simison:
>  30 days is too much.

I agree that's a lot, I hope in reality it'll be sooner than that! Let us know how it goes. 

Perhaps someone in the community comes up with a separate "Couchsurfing exporter" tool which does the scraping you described and produces a file in the same format much faster than the wait via their support.

Let's hope their support is able to procude these faster later too, once the storm settles down a bit after their paywall change.

guysoft:
@simison Do you have an example json file like this? Or just tell me what fields are being used? I looked at this PR and I only see the download and unzipping part, no the JSON parsing bit where you look for specific fields.

simison:
Here's my profile mostly anonymized for example: https://gist.github.com/simison/2c46cbe2655104247f19d4f633f01f68 

I have data in some of the array fields I emptied for example but they're not really relevant for this right now (forum posts, guests, etc)

The zip also contains a bunch of photos in a quite low format.

I'd start very simple and just see what it takes to import the profile description first.

guysoft:
@simison Is there any code where you actually take fields? I want to see what fields you are actually using.

simison:
No not yet, this pr was at the stage where I was only processing the archive.

Next steps would be to read the field(s) and update profile with it.

simison:
Turns out this approach actually isn't the best — we don't nee to progress these in the backend, we can do it fully clientside as seen at https://github.com/simison/couchspinner/

PR #1100, opened 2019-07-06
Change welcome sequence names

simison:
#### Proposed Changes

* Change names to volunteers who actually reply to emails sent as replies to welcome sequence mails

@doktorhorst @noahsmindfuck would you be fine with having your names in these emails that go out to every new member?

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -52,6 +52,7 @@ module.exports = {
   sessionCollection: 'sessions',
   domain: 'localhost:3000',
   supportEmail: 'support@trustroots.org', // TO-address for support requests
+  supportVolunteerNames: ['Dario', 'Noah'], // Used as "from" name to send some automated emails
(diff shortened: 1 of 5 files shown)


doktorhorst:
Sure, mine is already in there since Novemeber

simison:
Oh yeah 👍  So there are 3 emails in total, which previously had names:
- Kasper
- Natalia
- Dario

This will switch them all to be randomly either "Noah" or "Dario"

noahsmindfuck:
Yep, great! :)

simison:
Let me also know if you think we should adjust welcome sequence time intervals. To me it seems quite ok right now but it might look different in the support queue. We might as well experiment with different days if you'd like to see if it changes what kind of replies people write back.

doktorhorst:
I just did a rough count:

|e-mail|subject |# replies in 2019 |
|:-|:- |- |
|1|Welcome to Trustroots|10|
|2|Meet new people at Trustroots| 23|
|3|How is it going| 37|

It doesn't say anything about quality of the feedback, though. Just the "effectiveness".

PR #1107, opened 2019-09-01
Update magic number for splash screen

ahstro:
The home screen was acting up in Chromium, showing double scrollbars, a white margin at the bottom (not apparent in screenshot because I didn't think to consider that GitHub's background color is white), and not scrolling.

Since this is part of the Angular code and I'm new to the code base, I didn't want to put too much effort into a nice solution, so I just went with updating the magic number so the issue went away :innocent:

Seems to work fine with different screen sizes, but it's probably a good idea if someone else tests the code as well.

#### Testing Instructions

* Open [Trustroots](https://www.trustroots.org) in Chromium (76.0.3809.132)
* Try to scroll

Before:
www trustroots org_withnotes
After:
localhost_3000_

### DIFF
--- a/modules/pages/client/controllers/home.client.controller.js
+++ b/modules/pages/client/controllers/home.client.controller.js
@@ -37,7 +37,7 @@
     var vm = this;
 
     // Exposed to the view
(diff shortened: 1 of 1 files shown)


simison:
Nice catch!

PR #1121, opened 2019-11-11
Refactor RemoveContact logic and Contact component

mrkvon:
#### Proposed Changes
- move logic for removing contacts from Contact to RemoveContactContainer
(and we should choose consistent naming and file structure conventions)
- refactor Contact to be a functional component
- small fix of passing data to an event

#### Testing Instructions

* open my Contacts
* try to reject or revoke a contact request
* see that it works as before

Is a prerequisity for finishing #1062

### DIFF
--- a/modules/contacts/client/components/RemoveContact.js
+++ b/modules/contacts/client/components/RemoveContact.js
@@ -3,7 +3,15 @@ import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
 import { Modal } from 'react-bootstrap';
 
-export function RemoveContact({ t, contact, show, inProgress, onHide, onRemoveContact, situation='confirmed' }) {
+function getSituation(contact, selfId) {
+  return (contact.confirmed === false && contact.userFrom === selfId && 'unconfirmedFromMe')
+    || (contact.confirmed === false && contact.userTo === selfId && 'unconfirmedToMe')
+    || 'confirmed';
+}
+
+export function RemoveContact({ t, contact, show, inProgress, onRemove, onCancel, selfId }) {
+
+  const situation = getSituation(contact, selfId);
 
   // parse contact.created to Date
   const created = new Date(contact.created);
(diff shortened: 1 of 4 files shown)


modules/contacts/client/components/RemoveContact.js
```diff
@@ -3,7 +3,15 @@ import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
 import { Modal } from 'react-bootstrap';
 
-export function RemoveContact({ t, contact, show, inProgress, onHide, onRemoveContact, situation='confirmed' }) {
+function getSituation(contact, selfId) {
```
simison: This could be divided into two helpers: `isConfirmed()` and `isFromMe()` for better code clarity, or simply accessing `confirmed` directly and having a helper for just `isFromMe()`

Then you'd use it as so:

```js
const { confirmed } = contact;
if(confirmed && isFromMe(selfId)) {
 // ...
}
```

It just feels like that would be more ergonomic and less fuzzy than "situation" nomenclature.

Not a blocker for merging tho. :-)

mrkvon: I like clarifying the logic like this. :+1: Thank you @simison!
Incorporated it into this PR and added more fixes in a98e340.

mrkvon:
@simison Have you tested it from Contact list?

image

Elsewhere (in headers) we still use the old angular component, as of this PR.

image

simison:
Ahh you're right. I now tested also that contact list version and that works as expected in this PR. :thumbsup: Good that you spotted it. :-)

Akronix:
Just a little language question, what is more appropriate: "Remove contact request", "Cancel contact request" or "Delete contact request"?
I heard once that delete is rather a word used by geeks.

doktorhorst:
I'm not a native English speaker, but I would say that *cancel a request* is the proper expression. It's also the option with most results on Google.

Akronix:
I agree with @doktorhorst

simison:
I agree "delete" is very technical — anyone would like to PR a quick change? :-)

Akronix:
Me! :raising_hand_man:

PR #1123, opened 2019-11-14
Migrate tribe's JoinButton to React

mrkvon:
#### Proposed Changes

Rewrite `JoinButton` to React.

image

#### TODO

- [x] leave tribe popup
- [x] tooltip when Leave
- [ ] update number of people in the tribe +-1 (better done later down the road)
- [ ] google analytics (WONTDO)

#### Testing Instructions

* see that everything works the same

Part of #1133

### DIFF
--- a/modules/tribes/client/api/tribes.api.js
+++ b/modules/tribes/client/api/tribes.api.js
@@ -1,0 +1,9 @@
+import axios from 'axios';
+
+export async function join(tribeId) {
+  await axios.post(`/api/users/memberships/${tribeId}`, { tribeId });
+}
+
+export async function leave(tribeId) {
+  await axios.delete(`/api/users/memberships/${tribeId}`);
+}
--- a/modules/tribes/client/api/tribes.api.js
+++ b/modules/tribes/client/api/tribes.api.js
@@ -1,7 +1,7 @@
 import axios from 'axios';
 
 export async function join(tribeId) {
-  const { data } = await axios.post(`/api/users/memberships/${tribeId}`, { tribeId });
+  const { data } = await axios.post(`/api/users/memberships/${tribeId}`);
   return data;
 }
 
--- a/modules/tribes/client/api/tribes.api.js
+++ b/modules/tribes/client/api/tribes.api.js
@@ -1,9 +1,11 @@
 import axios from 'axios';
 
 export async function join(tribeId) {
-  await axios.post(`/api/users/memberships/${tribeId}`, { tribeId });
+  const { data } = await axios.post(`/api/users/memberships/${tribeId}`, { tribeId });
+  return data;
 }
 
 export async function leave(tribeId) {
-  await axios.delete(`/api/users/memberships/${tribeId}`);
+  const { data } = await axios.delete(`/api/users/memberships/${tribeId}`);
+  return data;
 }
--- a/modules/tribes/client/components/JoinButton.component.js
+++ b/modules/tribes/client/components/JoinButton.component.js
@@ -1,0 +1,76 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+
+import * as api from '../api/tribes.api';
+
+function JoinButtonPresentational({ isMember, onToggle, isLoading, joinLabel='Join', joinedLabel='Joined', tribe, isLogged }) {
+  const { t } = useTranslation('tribes');
+
+  const ariaLabel = (isMember) ? t('Leave Tribe') : t(`${joinLabel} ({{label}})`, { label: tribe.label });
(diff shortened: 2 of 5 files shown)


modules/tribes/client/components/JoinButton.component.js
```diff
@@ -0,0 +1,108 @@
...
+
+// @TODO this can (and should) be replaced by other container, when we finish the migration; when we start using redux etc.
+export default function JoinButton({ tribe, user, onUpdated, ...rest }) {
+  // isLeaving controls whether the tribe delete modal is shown
+  const [isLeaving, setIsLeaving] = useState(false);
+
+  const [isUpdating, setIsUpdating] = useState(false);
+
+  const isSelfMember = user && user.memberIds && user.memberIds.indexOf(tribe._id) > -1;
+  const [isMember, setIsMember] = useState(isSelfMember);
```
nicksellen: I was wondering what the _self_ meant in `isSelfMember` I guess just to distinguish it from `isMember`? If so
`isMemberInitial` seems clearer to me, as the value passed into `useState()` is the initial value, and on subsequent executions of the JoinButton function, it will use stored state.

modules/tribes/client/components/JoinButton.component.js
```diff
@@ -0,0 +1,108 @@
...
+
+  return <>
+    <LeaveTribeModal show={isLeaving} tribe={tribe} onConfirm={handleLeave} onCancel={handleCancelLeave}/>
+    <JoinButtonPresentational tribe={tribe} isLogged={!!user} isMember={isMember} isLoading={isUpdating} {...rest} onToggle={handleToggleMembership} />
+  </>;
+}
+
+JoinButton.propTypes = {
+  tribe: PropTypes.object.isRequired,
+  user: PropTypes.object.isRequired,
```
nicksellen: I think user is not required (in the case of logged out user).

modules/tribes/client/components/JoinButton.component.js
```diff
@@ -0,0 +1,108 @@
...
+
+import * as api from '../api/tribes.api';
+
+function JoinButtonPresentational({ isMember, isLoading, joinLabel='Join', joinedLabel='Joined', tribe, isLogged, onToggle }) {
+  const { t } = useTranslation('tribes');
+
+  const ariaLabel = (isMember) ? t('Leave Tribe') : t(`${joinLabel} ({{label}})`, { label: tribe.label });
+  const title = (isMember) ? t(joinedLabel) : t(joinLabel);
+
+  if (!isLogged) {
```
nicksellen: `isLoggedIn` seems clearer to me, `isLogged` sounds like it's something related to login to my brain.

modules/tribes/client/components/JoinButton.component.js
mrkvon: That's indeed the case. Will rename!

mrkvon: Done in c5c33f984a7cdb75721ceed50cd222bc2b55b327

modules/tribes/client/components/JoinButton.component.js
mrkvon: Done in c5c33f984a7cdb75721ceed50cd222bc2b55b327

modules/tribes/client/components/JoinButton.component.js
mrkvon: That's true. Was causing warnings when logged out. Fixed in c5c33f984a7cdb75721ceed50cd222bc2b55b327.

modules/tribes/client/api/tribes.api.js
```diff
@@ -0,0 +1,11 @@
+import axios from 'axios';
+
+export async function join(tribeId) {
+  const { data } = await axios.post(`/api/users/memberships/${tribeId}`, { tribeId });
```
nicksellen: You don't need to pass `tribeId` as a parameter here, in the angular version the parameter is just used to replace the path parameter:

```js
  function UserMembershipsService($resource) {
    return $resource('/api/users/memberships/:tribeId?', {
      tribeId: '@tribeId'
    }, {
      post: {
        method: 'POST'
      },
      delete: {
        method: 'DELETE'
      },
      get: {
        method: 'GET'
      }
    });
  }
```

The server gets the `tribeId` from params (which is path params):

https://github.com/Trustroots/trustroots/blob/303e23277de3fce2eafb32988bbe286de276dc03/modules/users/server/controllers/users.profile.server.controller.js#L904

mrkvon: Careful observation. :+1: Removed body from the request and tested that it still works in fa91c259bb1bc13ed8b98c1aca7f10e085b12e21

mrkvon:
@nicksellen Thank you for the reviews!

PR #1126, opened 2019-11-15
Convert codebase to ES6

simison:
#### Proposed Changes

* Convert all code (server and client) to ES6:
  * Manually fix a few var → let/const situations
  * Remove [`use strict`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) using https://github.com/5to6/5to6-codemod and enable `no-strict` rule
  * Convert `var` → `const`/`let` (using `eslint --fix` and a few manual fixes)
  * Other things that `eslint --fix` can do
* Reorganise testutils to server and client subfolders for easier linting via separate server & client side overrides
* Update Eslint config:
  *  Lint everything as es6
  * Refactor to be more specific for each override section; each section has their own globals instead of keeping them together for the whole app
  * Separate client-side and server-side, as well tests and app files.
  *  Use `eslint:recommended` as a base; remove rules in our custom set which would be duplicate with recommended.
  * Add `one-var` linter rule since we're already touching all these lines. This should make e.g. automatically replacing require→import easier later on.
   * Removed all redundant Eslint disabled rules

#### Testing Instructions

* Please review `.eslintrc.js` — there are so many other changes that you might just want to skim through a bit and smoke test the site. 
* `npm start` works?
* `npm run lint` green?
* Confirm that clientside built bundles don't have `const`/`let` in sources and that they're ES5 code.


simison:
There's one Eslint error that I didn't yet figure out of it's legit or not:
```
modules/references/server/controllers/reference.server.controller.js
  415:5  error  Possible race condition: `req.reference` might be reassigned based on an outdated value of `req.reference`  require-atomic-updates
```

https://github.com/Trustroots/trustroots/blob/51d467494cfff583b6753f1222f1c73b7d477ad6/modules/references/server/controllers/reference.server.controller.js#L417

https://eslint.org/docs/rules/require-atomic-updates

simison:
I'm marking this already ready for review; just a couple lint issues to fix but otherwise ready.

.eslintrc.js
```diff
@@ -22,24 +22,19 @@ const defaultRules = {
...
-  'no-undef': 2,
   'no-unneeded-ternary': 2,
-  'no-unreachable': 2,
   'no-underscore-dangle': 0,
   'no-unused-expressions': 0,
-  'no-unused-vars': 2,
   'no-use-before-define': [1, 'nofunc'],
-  'no-var': 0,
   'object-curly-spacing': [2, 'always'],
-  'one-var': [0, 'never'],
```
simison: This is a new rule.

.eslintrc.js
```diff
@@ -22,24 +22,19 @@ const defaultRules = {
   'no-caller': 2,
   'no-console': 2,
   'no-else-return': 0,
-  'no-empty-character-class': 2,
```
simison: All (except `one-var`) removed rules are still applied because they're in `eslint:recommend` config. Just removing as duplicates.

mrkvon:
@simison wrote:

> There's one Eslint error that I didn't yet figure out of it's legit or not

There is an [open issue for this behaviour on eslint](https://github.com/eslint/eslint/issues/11899).

There is nothing racing in our case. We may just
```js
// eslint-disable-next-line require-atomic-updates
req.reference = whatever()
```

mrkvon:
Probably not related to this PR, but `eslint-plugin-angular` in package.json may want to go to `devDependencies` instead of `dependencies`.

.eslintrc.js
```diff
@@ -22,24 +22,19 @@ const defaultRules = {
...
   'no-multi-spaces': 2,
   'no-param-reassign': 0,
```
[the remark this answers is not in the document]
simison: Valid point! I simply didn't think about it in this PR and you're prolly right that they date back long ago. :-)

They're ripe for reviewing+removing. 👍

.eslintrc.js
```diff
@@ -55,192 +50,199 @@ const defaultRules = {
...
+        'angular/typecheck-number': 2,
+        'angular/typecheck-object': 2,
+        'angular/typecheck-string': 2,
+        'angular/window-service': 2
+      },
+      settings: {
+        angular: 1,
+      },
+      env: {
+        browser: true,
```
simison: We actually might want to turn this `browser` env off and simply allow `document` and `window` globals.

That will stop nasty bugs like where someone uses `find()` and forgets to import it from `lodash`, causing `window.find` to be referenced and nothing would warn about it — it would just silently break.

Good for another PR.

.eslintrc.js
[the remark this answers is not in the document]
simison: Aight! Removed all these in https://github.com/Trustroots/trustroots/pull/1126/commits/a5b5135eb256f8cb6e95a21fe1ddefd81162aa54 and flipped  `no-strict` from `0` to `2` since it was relevant for this PR; https://github.com/Trustroots/trustroots/pull/1126/commits/8456afdfdb3a2eb5e7a639fcfa505c1145a22708

simison:
@mrkvon this no longer throws eslint warnings  (I simply inline-ignored the two in bin script https://github.com/Trustroots/trustroots/pull/1126/commits/421ae54261e252a2e79dfc738dc80114658c0dad and disabled another rule in eslint since it'll likely pop up quite a bit when we transform more backend code to promises. https://github.com/Trustroots/trustroots/pull/1126/commits/c3f4ce9289fe2d68fcf68e5ff0543f81bce26a97

Good for one more spin before merging!

.eslintrc.js
```diff
@@ -50,197 +34,204 @@ const defaultRules = {
   }],
   'space-in-parens': [2, 'never'],
   'spaced-comment': [2, 'always'],
-  strict: 0,
+  strict: [2, 'never'],
```
mrkvon: What is the reasoning behind adding this rule? Do we actually want to use [sloppy mode](https://developer.mozilla.org/en-US/docs/Glossary/Sloppy_mode)? (I guess not?) Or do we address the issues in a different way?

simison:
Thanks, folks!

Yeah, if something gets merged without being rebased and introduces ES5, it's fine, we can jump in and those will be super simple fixes.

.eslintrc.js
simison: So ES6 modules are always in strict mode automatically, so the declaration is redundant. That said, we have heaps  of old Angular code that would prolly benefit from having it, but in my understanding, we'll be safe just by nature of us modifying and adding mostly new  ES6 code and _removing_ old Angular "ES5" code.

mrkvon: I see. Makes sense.
We could still benefit from strict mode in api code and tests, couldn't we? AFAIK it uses `require` and `module.exports` for exporting and importing. Have we used strict mode until now?

PR #1127, opened 2019-11-15
Fix eslint `no-prototype-builtins` issues

simison:
This is a smaller iteration step from https://github.com/Trustroots/trustroots/pull/1126 where I'm converting eslint config to extend [`eslint:recommended`](https://eslint.org/docs/user-guide/configuring#extending-configuration-files).

That ruleset includes [`no-prototype-builtins`](https://eslint.org/docs/rules/no-prototype-builtins) rule and there were a few warnings so I set to fix those. Instead of using elaborate `Object.prototype.hasOwnProperty.call(foo, "bar")` I'm just using Lodash's `has` here instead.

I generally like using vanilla JS over Lodash especially in frontend code but this is all server-side and just feels more readable in these cases. We're also already using Lodash in all these files.

#### Proposed Changes

* Use `_.has()` over `hasOwnProperty` checks

#### Testing Instructions

* Does the app start?
* Do the tests pass?

### DIFF
--- a/modules/references/server/controllers/reference.server.controller.js
+++ b/modules/references/server/controllers/reference.server.controller.js
@@ -41,14 +41,14 @@ function validateCreate(req) {
 
   // Values of interactions must be boolean
   ['met', 'hostedMe', 'hostedThem'].forEach(function (interaction) {
-    if (req.body.interactions && req.body.interactions.hasOwnProperty(interaction) && typeof req.body.interactions[interaction] !== 'boolean') {
+    if (_.has(req, ['body', 'interactions', interaction]) && typeof req.body.interactions[interaction] !== 'boolean') {
       valid = false;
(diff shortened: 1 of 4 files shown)


modules/references/server/controllers/reference.server.controller.js
```diff
@@ -41,14 +41,14 @@ function validateCreate(req) {
 
   // Values of interactions must be boolean
   ['met', 'hostedMe', 'hostedThem'].forEach(function (interaction) {
-    if (req.body.interactions && req.body.interactions.hasOwnProperty(interaction) && typeof req.body.interactions[interaction] !== 'boolean') {
+    if (_.has(req, ['body', 'interactions', interaction]) && typeof req.body.interactions[interaction] !== 'boolean') {
```
simison: Good win especially here since we're checking for a path two levels deep.

PR #1130, opened 2019-11-20
Replaced "Delete Request" by "Decline Request" contact requests

Akronix:
#### Proposed Changes

* Replace text of "Delete Request" by "Decline Request" which is more "human" friendly.

#### Testing Instructions

* Send a contact request and see changed text.

Fixes discussion in #1121

### DIFF
--- a/modules/contacts/client/components/ContactPresentational.js
+++ b/modules/contacts/client/components/ContactPresentational.js
@@ -44,7 +44,7 @@ export function ContactPresentational({ t, className, contact, avatarSize=128, h
           {t('Confirm Request')}
         </a>
         <a className="btn btn-xs btn-warn" onClick={onClickRemove}>
-          {t('Delete Request')}
+          {t('Decline Request')}
         </a>
       </div>}
     </div>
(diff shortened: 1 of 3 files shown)


Akronix:
Btw, It would be a good idea to have a number red icon with the number of non-answered requests in the "Contacts" drop-down menu item on the top right corner. What do you think @simison ? do we have any issue regarding this?

simison:
@Akronix Yeah I think that would be helpful, but only to highlight requests that I can act on. It's not useful (rather annoying) to highlight requests I've sent and the other party hasn't replied to.

Would it be enough to have the contact number highlighted red, or did you mean to add an additional number? I think additional number could be confusing any anyway most of the time just have a number 1.

We should use these colors for it:

https://github.com/Trustroots/trustroots/blob/55fd5a0357ce223f57ca184f72b6d82e0159904b/modules/core/client/less/global-variables.less#L54-L56

image

Akronix:
Agree with the first paragraph.

I actually was thinking in this dropdown menu:
image

In the tab menu you are talking about I would add another number within parentheses or after a `+` symbol, or otherwise I'd leave it like how it is now.

Anyway, let's treat this as a separated issue.

simison:
Oh yeah, that menu is great for something like this too! Just for that the data fetching might be little more involved work.

Issue #1032, opened 2019-02-04
Reply rate: ignore messages from past X hours/days


simison:
**Observation:**

My profile says my reply rate is 0% — Is that just because I haven't had any conversations in a while? I did get one message just now that I haven't replied, tho...

**Issue:**
It was precisely because you had that one unreplied request and it's your only received request within last 3 months.

**Solution:**
 It would be nicer if we ignore new messages for the first 3 days or until they're replied. 

**The code is here:**

https://github.com/Trustroots/trustroots/blob/79539a08c2755b94266d2a4d3ec21f23bdcd02a8/modules/messages/server/services/message-stat.server.service.js#L231-L248

PR #1132, opened 2019-11-26
Issue #1032 Enhancement Fix: ignore messages from past 3 days

suljicv:
#### Proposed Changes

* In response to issue #1032, this fixes the problem where response rate automatically drops to 0 in the case of no immediate reply for users that haven't been messaging in a while. It waits three days before setting reply rate to 0.

### DIFF
--- a/modules/messages/server/services/message-stat.server.service.js
+++ b/modules/messages/server/services/message-stat.server.service.js
@@ -231,14 +231,14 @@ exports.readMessageStatsOfUser = function (userId, timeNow, callback) {
   async.waterfall([
     /**
      * Get the data from the database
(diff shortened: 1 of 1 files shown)


mrkvon:
I love this. @suljicv Since I was implementing the original logic, let me know if you wish to have some support.

guillaumebrunerie:
Looking at the code, this pull request does not seem to do the right thing. It only handles the case where all messages received in the last three months were actually received in the last three days.
But if I responded to a message a month ago and then receive a new message today, my reply rate should not immediately drop to 50%, it should wait three days before doing so. With this code, it will still immediately drop to 50%.

Issue #1133, opened 2019-11-26
Rewrite Tribes page to React


mrkvon:
Rewrite `/tribes` page in React.

This is a roadmap and overview for the task.

See #1019 for inspiration.

PR #1134, opened 2019-11-27
Migrate Tribe to React

mrkvon:
#### Proposed Changes

Make tribe into a react component.

image

#### Testing Instructions

* See that design and functionality survived the migration
* See that the tests and lint pass

Part of #1133 
~Depends on #1123, should be reviewed after that one is merged to the feature branch `tribes-react`~


mrkvon:
The member count doesn't update. Info broadcasted to angular is not listened to. (I actually don't know what that means in angularjs world...)
May be fixed on React level as the migration proceeds.

mrkvon:
The number of members doesn't update, but it's being taken cared of in #1145. So please bear with that... :slightly_smiling_face:

modules/tribes/client/components/helpers/getTribeBackground.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {string} [quality=lighter] - Set background image quality
+ *   Options: normal, better, best, lighter (default), lightest
+ *   See https://uploadcare.com/documentation/cdn/#operation-quality
+ *
+ * @param {boolean} [isProgressive=false] - Set progressive image loading
+ *   See https://uploadcare.com/documentation/cdn/#operation-progressive
+ *
+ * @returns {object}
+ */
+export default function getTribeBackground(tribe, { quality='lighter', dimensions='1024x768', isProgressive=false }={}) {
```
nicksellen: minor point: `getTribeBackgroundStyle` would be slightly clearer to me, it wasn't immediately obvious that it was going to return me a style object I can use in a `style` attribute.

modules/tribes/client/less/tribes-grid.less
```diff
@@ -7,7 +7,7 @@
     background-color: @brand-primary;
     background-size: cover;
     height: 250px;
-    width: 288px;
+    width: 100%;
```
nicksellen: I wondered what this was about, but I understand now, as the react component ends up in a nested element, it doesn't get it's width set by the angular grid thing (which sets pixel widths directly on each element), so our inner element can just be full width :+1:

modules/tribes/client/components/helpers/getTribeBackground.js
mrkvon: Renamed in 2c37f3f38c0504bcd17b6f27000296e0d66fe448

modules/tribes/client/less/tribes-grid.less
mrkvon: @nicksellen Would you like to have a comment added there?

mrkvon:
There is still a failing propType on JoinButtonPresentational. It gets fixed in the further PRs, I promise! :slightly_smiling_face:

modules/tribes/client/less/tribes-grid.less
nicksellen: Nope, seems fine without :)

PR #1135, opened 2019-11-27
Tribes react suggest tribe

mrkvon:
#### Proposed Changes
Migrate `SuggestTribe` component to React
This one is easy: It has no state or props.

image

#### Testing Instructions

* See that it behaves as before

Part of #1133


mrkvon:
Removed the dependency on #1134. Ready for review.

PR #1136, opened 2019-11-28
React

mrkvon:
#### Proposed Changes

- Highly experimental branch
- Building the React app from root as opposed to building from edges.
- Allows us to set up state management with `redux` and routing with `react-router-dom`
- Radical separation into `server` and `client`
- App created with `create-react-app`

#### Testing Instructions

*


nicksellen:
Hey! You asked for some comments on this approach. I have to say I didn't look in detail, but I my first thought it that it is very useful to try out what a fresh `create-react-app` app looks like, as a target for what we can approach.

As for using it, it's whether we could ever migrate the whole site to react in one go, without getting tired of working on a mega rewrite branch for months on end. I don't think it's impossible actually, and also I don't think it invalidates any (or much) of the work done migrating to react so far (as those components can just be used...). It's actually what I'm involved in doing for another project (a full frontend rewrite of https://github.com/Freegle/iznik (backbone and friends) to https://github.com/Freegle/iznik-nuxt (nuxt/vue)). I didn't decide to do that, I don't know if I would have or not.

There seems definitely an advantage to `create-react-app` especially as one of main drawbacks I saw before (needing to eventually _eject_ and lose all the nice simplicity and up gradability) seems like not very necessarily now (a lot more stuff is supported, and there are tools to override some things if you really need... e.g. [craco](https://github.com/gsoft-inc/craco)). On the other hand, I am never that bothered by the mountains of webpack/babel config, once it's setup it just sits there (ok, until the next big breaking upgrade... but still, I've done quite a few of them and not sooo bothered).

My little experiment to try and add react router into the existing app failed miserably, as it's really not a good idea to have two routes on the page at once (angular ui router, and react router), they fight each other. I sort of wanted to try and get react-router to just manage a subroute of the other one. There is also [@uirouter/react-hybrid](https://github.com/ui-router/react-hybrid) which lets you route to a react component or angular stuff, but I don't think it offers much over what we're already doing (it's main selling point is you can use their link components, but we don't want that, we'd only have to rip them out later...).

I think it wouldn't be too hard to add a redux store into the existing page quite soon though (perhaps after evaluating if it is a good way forward in general - seems probably so, although there are other state libraries and approaches (e.g. contexts), I briefly skimmed [this post](https://daveceddia.com/context-api-vs-redux/) which goes through many of the approaches).

In some way replacing the existing site bit by bit maybe makes it more solid if it's getting deployed regularly in between, errors should hopefully become known about quite soon. I don't know if it is actually being deployed much though... I will get involved with that more soon!

... but I also think it comes down a lot to how frustrating it might be to work on either approach, and that is quite personal to each developer. I wonder what your main pain points with the current piece by piece approach are @mrkvon?

mrkvon:
I'm discontinuing this for now. I may reopen or restart in the near or more distant future.

mrkvon:
@nicksellen Thank you for your thoughts on this topic! I must have missed them before.

I closed this PR because it's going to be impossible to rebase it, and I'm not working on it atm.

My main frustration is the following: the piece-by-piece work doesn't really allow to go deep, even as little as to routing; or it does, but temporary fixes are needed, which consume time and leave unclean state of affairs. For example this emitting events into angular; to update the client-side data model.

This is closed, but the idea is not. I'll come back to it.

PR #1137, opened 2019-11-28
Header for Tribes List

mrkvon:
#### Proposed Changes

Migrate to React the following:

image

#### TODO

- [x] broadcast the copyright of the photo (Board)

#### Testing Instructions

- see that the header on /tribes page works as before
  - both signed in and signed out
- see that the credits for the photo are updated

Work on #1133

### DIFF
--- a/modules/core/client/components/Board.js
+++ b/modules/core/client/components/Board.js
@@ -1,5 +1,6 @@
-import { Children, cloneElement, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import { selectPhoto } from '../services/photos.service';
 
 /**
@@ -16,8 +17,10 @@ function selectName(names) {
  * modules/core/client/directives/tr-boards.client.directive.js
  *
  * @TODO implement tr-boards-ignore-small directive
+ * @TODO implement primary, inset, error and maybe other attributes, which are currently board classes
+ *  and which could become attributes <Board primary inset error names="bokeh" />
  */
-export default function Board({ names='bokeh', children, onDisplayPhoto=() => {}, onHidePhoto=() => {} }) {
+export default function Board({ names='bokeh', children, onDisplayPhoto=() => {}, onHidePhoto=() => {}, className, ...rest }) {
 
   const [photo, setPhoto] = useState(null);
 
@@ -40,7 +43,11 @@ export default function Board({ names='bokeh', children, onDisplayPhoto=() => {}
   }, []);
 
   const style = photo ? { backgroundImage: `url("${photo.imageUrl}")` } : null;
-  return Children.map(children, child => cloneElement(child, { style }));
+  return (
+    <section style={style} className={classNames('board', className)} {...rest}>
+      {children}
+    </section>
+  );
 }
 
 Board.propTypes = {
@@ -48,6 +55,8 @@ Board.propTypes = {
     PropTypes.string,
     PropTypes.arrayOf(PropTypes.string)
   ]).isRequired,
+  className: PropTypes.string,
+  children: PropTypes.node,
   onDisplayPhoto: PropTypes.func,
   onHidePhoto: PropTypes.func
 };
--- a/modules/core/client/components/Board.js
+++ b/modules/core/client/components/Board.js
@@ -5,7 +5,7 @@ export default function Board({ name='bokeh', children, onNameChanged }) {
 
   // update
   useEffect(() => {
-    onNameChanged();
+    onNameChanged({ name: selectPhoto(name) });
   }, [name]);
 
   const photo = selectPhoto(name);
--- a/modules/core/client/components/Board.js
+++ b/modules/core/client/components/Board.js
@@ -1,0 +1,14 @@
+import { Children, cloneElement, useEffect } from 'react';
+import { selectPhoto } from '../services/photos.service';
+
+export default function Board({ name='bokeh', children, onNameChanged }) {
+
+  // update
+  useEffect(() => {
+    onNameChanged();
+  }, [name]);
+
+  const photo = selectPhoto(name);
+  const style = { backgroundImage: `url("${photo.imageUrl}")` };
+  return Children.map(children, child => cloneElement(child, { style }));
+}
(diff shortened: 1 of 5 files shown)


modules/core/client/components/Board.js
```diff
@@ -0,0 +1,14 @@
+import { Children, cloneElement, useEffect } from 'react';
+import { selectPhoto } from '../services/photos.service';
+
+export default function Board({ name='bokeh', children, onNameChanged }) {
```
nicksellen: I don't see anywhere where the name actually changes? I guess you implemented it for cases where it does change dynamically, although the current `tr-boards` directive seems to let you pass an array of names, and it will randomly select one and keep to that one, so no dynamic updating.

modules/core/client/components/Board.js
```diff
@@ -0,0 +1,14 @@
...
+
+  // update
+  useEffect(() => {
+    onNameChanged({ name: selectPhoto(name) });
+  }, [name]);
+
+  const photo = selectPhoto(name);
```
nicksellen: .. and I guess missing support for:
- `tr-boards-ignore-small`
- `window.innerWidth <= 480`

... which in this case is because they're not used on the tribes header, but maybe add a note that it doesn't exactly provide the same functionality as the `tr-boards` directive.

modules/tribes/client/components/TribesHeader.component.js
```diff
@@ -0,0 +1,37 @@
...
+            </div>}
+          </div>
+        </div>
+      </div>
+    </section>
+  </Board>;
+}
+
+TribesHeader.propTypes = {
+  isLogged: PropTypes.bool.isRequired,
```
nicksellen: Same comment as on the other PR about `isLogged` vs `isLoggedIn` :)

modules/tribes/client/components/TribesHeader.component.js
```diff
@@ -0,0 +1,37 @@
...
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+import Board from '../../../core/client/components/Board';
+
+export default function TribesHeader({ isLogged, onBoardChanged }) {
+
+  const { t } = useTranslation('tribes');
+
+  return <Board name="tribes-1" onNameChanged={onBoardChanged}>
+    <section className="board tribes-header" tr-boards="'tribes-1'">
```
nicksellen: Can remove `tr-boards="'tribes-1'"` now it's in react :)

modules/core/client/components/Board.js
```diff
@@ -0,0 +1,14 @@
...
+  const style = { backgroundImage: `url("${photo.imageUrl}")` };
+  return Children.map(children, child => cloneElement(child, { style }));
```
nicksellen: I wonder whether it would be simpler to just make it output a `<section>` element, instead of manipulating the background image of its children, as in all the uses I can see, it's always on a `<section>` element...

It feels kind of magical for it to modify the children.

I think the logic for it to be a directive before is that in angular you had to choose between an something with a template and controller (which felt a bit heavy), and something that just fiddles with the element (a directive), whereas in react they are equally lightweight (something like that anyway, I've slightly forgotten about angular...)

modules/tribes/client/components/TribesHeader.component.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+import Board from '../../../core/client/components/Board';
```
nicksellen: There a nice webpack alias to avoid all the `../../../` noise :)

modules/core/client/components/Board.js
mrkvon: > pass an array of names, and it will randomly select one

:+1: That's a correct observation, I seem to have ignored the _select random photo_ feature. I'll add a TODO comment for it. (We'll need it at Home page I think.)

> cases where it does change dynamically

Is it possible to implement a React component which doesn't change dynamically when its prop changes?

---

With the `useEffect` + `onNameChanged` some ancestor should tell another component to display credits of the photo.
- Now doing it angular way: Broadcasting/emitting some info to a $scope. (I don't understand details of what that means. Seems to work.)
- As the migration progresses, it will inform some shared state that this photo is displayed. It should also inform this state when the photo is not displayed anymore. I haven't implemented that.

So this component is somewhat confused. I'll have a closer look and fix it.

Does the above make sense?

mrkvon: - allows array of photo names as input
- improve emitting info about displayed photos for credits (display/hide info; select photo when component mounts only)
- added the TODO for tr-boards-ignore-small

Implemented in 76d84e9c3e396256b181a84f3b55836b3edcac1a

modules/core/client/components/Board.js
mrkvon: added the TODO for future implementation of tr-boards-ignore-small in 76d84e9

modules/tribes/client/components/TribesHeader.component.js
mrkvon: fixed in 76d84e9c3e396256b181a84f3b55836b3edcac1a

modules/tribes/client/components/TribesHeader.component.js
mrkvon: fixed in 76d84e9c3e396256b181a84f3b55836b3edcac1a

modules/tribes/client/components/TribesHeader.component.js
mrkvon: Moved the whole `<section>` inside `Board` in fc4f0f4bdf7fa8a47447c9f638ac7c9c74ff6623

modules/core/client/components/Board.js
mrkvon: Simplified in fc4f0f4bdf7fa8a47447c9f638ac7c9c74ff6623

modules/core/client/components/Board.js
nicksellen: > Is it possible to implement a React component which doesn't change dynamically when its prop changes?

No, I doubt it, it's just that if it will never change, then the emitting the name change would never happen (apart from the initial call), but it's a bit odd that you tell it the name, then it emits back to you what the name is! ... however ....

> As the migration progresses, it will inform some shared state that this photo is displayed.

... because of this, I don't think we need to address that right now.

So, yes! Resolves issue! :star2:

mrkvon:
@nicksellen Thank you for the careful reviews! :tada:

PR #1138, opened 2019-11-28
Fix script test:server:watch to watch again

mrkvon:
#### Proposed Changes

* `npm run test:server:watch` was failing after the tests ran. Fix.

#### Testing Instructions

* see that before this change, `npm run test:server:watch` doesn't watch, but fails after the first round of tests
* see that after this change, `npm run test:server:watch` runs again

#### To do later
Some tests weirdly fail in the second and further rounds of tests. Won't fix in this PR.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -182,7 +182,7 @@ gulp.task('watch:server:run-tests', function watchServerRunTests() {
   // Add Server Test file rules
   gulp.watch([
     'modules/*/tests/server/**/*.js',
-    defaultAssets.server.allJS,
+    ...defaultAssets.server.allJS,
     defaultAssets.server.migrations
   ],
   gulp.series('test:server'))


Akronix:
Supposedly, the test should be re-run when the source code of a test or the source code of the app is modified, right?
Because this is not happening to me :confused:

mrkvon:
@Akronix I confirm that the tests don't rerun when they're still in progress. One has to wait until all are run before they get rerun. That is a bug. However I didn't intend to fix that in this PR. Would you consider opening a bug report (or seeing if there is one already) for this?

Akronix:
I mean now if I run the script `npm run test:server:watch`, I wait until all tests are done and then I change some source code or test code, the tests don't get rerun.

I thought that the intended purpose of this script is that tests get rerun when some code is modified.

mrkvon:
@Akronix said:
> purpose of this script is that tests get rerun when some code is modified

Indeed that's the purpose. The behavior you describe is unexpected, i.e. a bug.

What's your setup? (docker, ...) Is any error logged in your terminal at the end of the tests? What file do you update?

My tests still reload when I save a js file. I can't reproduce what you describe. However I've already experienced various bugs. The `watch` behavior seems to be broken on multiple levels.

- no reloading while tests are running
- complains about a port in use
- tests stop when any test fails in the first round
- ...

Intent of this PR is not to fix all of the above. It is to fix a simple bug, e.g. to spread an array which was not spread before, into another array in `.gulpfile`.

@Akronix Can you try to run the `npm run test:server:watch` without this fix and see if they watch or if the script fails? Can you compare with this fix?

What you experience would either mean that the behavior of .gulpfile is inconsistent across different setups; or would be good for a separate issue.

Akronix:
OK. So I've been doing some tests.

This reloads only when server code is modified, but not for client code, is that alright?

Akronix:
I just read that the script is `test:server:watch` and I was modifying client code :sweat:
Modifying server code works OK for me :+1:

mrkvon:
I take liberty to merge this.

simison:
Thanks for fixing it!

PR #1139, opened 2019-12-02
Remove app store buttons in native app

Akronix:
#### Proposed Changes

* Remove app store buttons when inside the native app

#### Testing Instructions

* Use mobile expo and app store buttons shouldn't be there anymore

[Fixes one blocking point to publish iOS](https://github.com/Trustroots/trustroots-expo-mobile/issues/10#issuecomment-528238998)


Akronix:
ping @simison

simison:
Now that this is deployed to production and I tested via Android app, the badges were still visible. 🤔 

I wonder if it's `::` which stops Angular from re-rendering when the prop changes (it might be false on the first render).

PR #1143, opened 2019-12-10
Migrate "Join Trustroots" section on /tribes page to React.

mrkvon:
It's the section which appears at the bottom of /tribes page when user is signed out.

#### Proposed Changes

* migrate the following to React (/tribes):

image

#### Testing Instructions

* Sign out
* go to `/tribes`
* see that the above mentioned section is present and behaves as expected
* see that the section is not present when signed in

Work on #1133


mrkvon:
@nicksellen thank you for the review!

Merged to feature branch `tribes-react`.

PR #1144, opened 2019-12-10
Migrate list of tribes to React

mrkvon:
#### Proposed Changes

Migrate to React the following:

image

The [angulargrid](https://yarnpkg.com/en/package/angulargrid) was replaced with custom css, using [styled-components](https://yarnpkg.com/en/package/styled-components).

#### Testing Instructions

* See that the list of tribes behaves as expected.
  Especially check that the grid behaves more or less as expected.

Work on #1133 
~Depends on #1134 and #1135. It's really just the last commit. The dependencies should be merged to `tribes-react` branch before reviewing of this starts.~ (done)


modules/tribes/client/components/Tribe.js
```diff
@@ -2,22 +2,27 @@ import React from 'react';
...
 import JoinButton from './JoinButton';
 
 import getTribeBackgroundStyle from './helpers/getTribeBackgroundStyle';
 
+const Container = styled.div`
+  position: relative;
+`;
+
 /**
  * @TODO maybe rename to Tribe
```
nicksellen: You did it!

modules/tribes/client/components/TribesList.component.js
```diff
@@ -0,0 +1,56 @@
...
+  margin: -5px -5px 15px -5px;
+  line-height: 0;
+`;
+
+const Item = styled.li`
+  line-height: 1.42857143; // set back the line-height set by bootstrap
+  padding: 5px;
+  display: inline-block;
+  vertical-align: bottom;
+  // background-color: red;
```
nicksellen: Can remove commented out style

modules/tribes/client/components/TribesList.component.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @media (min-width: 992px) {
+    width: 33.3333%;
+  }
+`;
+
+export default function TribesList({ tribes, user, onMembershipUpdated }) {
+  return (
+    <List className="list-unstyled tribes-grid">
```
nicksellen: I wonder how to handle existing CSS classes as we move forward? Should it all move into styled-component definitions?

package.json
```diff
@@ -170,6 +170,7 @@
     "serve-favicon": "~2.5.0",
     "speakingurl": "~14.0.1",
     "stathat": "0.0.8",
+    "styled-components": "~4.4.1",
```
nicksellen: I wonder why styled-components as opposed to other react styling libraries. I took a quick look, and seems a lot of them have been deprecated over time, so styled-components is looking good from that perspective.

The only other one that caught my eye from a brief look is something based on CSS-modules (e.g. https://github.com/gajus/babel-plugin-react-css-modules).

I don't have a strong opinion about these things, really, although maybe others do! I can see advantages to having style defined without the file, or in seperate files.

simison: The choice these days [seems to be](https://css-tricks.com/the-fragmented-but-evolving-state-of-css-in-js/) mostly between [styled-components](https://github.com/styled-components/styled-components) (the most popular) or [Emotion](https://github.com/emotion-js/emotion). They have cross-compatible API so migrating from one to another [is fairly simple](https://github.com/WordPress/gutenberg/pull/17963#issuecomment-552842261).

I don't think the differences in our use would be significant; Emotion might be nicer when publishing components to NPM as it can isolate styles more effectively and it's also a little more performant than Styled components. Emotion can be themed easily (not sure about styled-components) and Styled components plays nicely with React Native (not sure about Emotion). There might be some gotchas in each library with server-side-rendering from w hat I've heard. None of these are our concerns, anyway.

CSS modules are really nice too but if you prefer something like this, probably better to clean-cut and go all in CSS-in-JS. :-) 

_That said_, just importing style files directly from JS already improves the developer experience significantly, together with more solid CSS selector naming conventions. Would that still be enough?

There's going to be a performance penalty with CSS-in-JS — which might be ok if dev experience improves, but just something to keep in mind.

[Facebook announced](https://www.youtube-nocookie.com/embed/9JZHodNR184) they're working on a solution that's supposedly solving performance issues; we might want to move to that eventually. Even migrating to it from any other CSS-in-JS library should be smooth enough, since APIs across different frameworks are typically quite similar. I'm sure someone will even write Codemods to help to migrate automatically.

modules/tribes/client/components/JoinButton.js
```diff
@@ -59,7 +59,9 @@ export default function JoinButton({ tribe, user, onUpdated, ...rest }) {
 
   const [isUpdating, setIsUpdating] = useState(false);
 
-  const isMemberInitial = user && user.memberIds && user.memberIds.indexOf(tribe._id) > -1;
+  // This should always be true or false, even when user is null.
+  // Therefore we convert the output to boolean.
+  const isMemberInitial = Boolean(user && user.memberIds && user.memberIds.indexOf(tribe._id) > -1);
```
simison: Nitpick; 
```
user.memberIds.indexOf(tribe._id) > -1
```
can also written with little more modern [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes): 
```
! user.memberIds.includes(tribe._id)
```
...up to you. :-)

modules/tribes/client/components/TribesList.component.js
```diff
@@ -0,0 +1,56 @@
...
+  line-height: 1.42857143; // set back the line-height set by bootstrap
+  padding: 5px;
+  display: inline-block;
+  vertical-align: bottom;
+  // background-color: red;
+  @media (max-width: 615.9999px) {
+    width: 100%;
+  }
+
+  @media (min-width: 616px) and (max-width: 991.9999px) {
```
simison: Looking at values like `991.9999px` makes me feel like it should come from a centralized variable somewhere.

Would it make sense to combine these styles with something like [`react-responsive`](https://www.npmjs.com/package/react-responsive)?

modules/tribes/client/components/TribesList.component.js
```diff
@@ -0,0 +1,56 @@
...
+import Tribe from './Tribe';
+import SuggestTribe from './SuggestTribe';
+
+const List = styled.ul`
+  margin: -5px -5px 15px -5px;
+  line-height: 0;
+`;
+
+const Item = styled.li`
+  line-height: 1.42857143; // set back the line-height set by bootstrap
```
simison: Similar concerns with this value than in https://github.com/Trustroots/trustroots/pull/1144/files#r362089900

Does the comment mean that Bootstrap sets different line-height for `li` than rest of elements?

modules/tribes/client/components/TribesList.component.js
nicksellen: react-response looks like it's for selectively rendering whole components based on media queries, but here it's about setting css based on media queries.

As styled-components are just strings could use something similar to https://jsramblings.com/2018/02/04/styled-components-media-queries.html.

https://github.com/morajabi/styled-media-query is another approach, but doesn't seem much improvement on the other approach, and needs an extra library.

modules/tribes/client/components/TribesList.component.js
mrkvon: The line height had been set to this value by bootstrap before the migration. Here we took that responsibility away from bootstrap, so we entered the value manually to keep the original looks.
https://stackoverflow.com/questions/19982651/why-does-bootstrap-set-the-line-height-property-to-1-428571429
I'll try to improve the comment. Or do you have ideas how else to improve this, @simison?

simison: Oh, hm, I'd really prefer we inherit those kind of styles from UI library (currently bootstrap) instead of manually entering them everywhere. :-/

Can we use Bootstrap classnames and inherit right base styles?

mrkvon: I don't understand why i set `line-height: 0` on `ul` in the first place. Removing that removes the necessity of specifying this weird value here.

I don't like writing this either. It was a lazy fix.

modules/tribes/client/components/TribesList.component.js
mrkvon: I don't know. What do you think?
I'm fine with leaving until we have the definitive decision whether to go forward with separate stylesheets or styled-components or other way of managing styles. Less work for now... :slightly_smiling_face:

nicksellen: Yup, I agree to leave it for now.

modules/tribes/client/components/JoinButton.js
mrkvon: What do you think about `(_.)get(user, ['memberIds'], []).includes(tribe._id)`? Are we ok with `import get from lodash/get` and such?

modules/tribes/client/components/TribesList.component.js
mrkvon: I just found we don't have to specify this fractional `max-width` because the css will care about the latest rule only.

So we could just do
```css
/* @media (min-width: 0) is just for clarity that width values are overwritten by the next one */
/* it can be removed. preferences? */
@media (min-width: 0) {
  width: 100%;
}
@media (min-width: 616px) {
  width: 50%;
}
@media (min-width: 992px) {
  width: 33.3333%;
}
```

When it comes to these breakpoints, they are pretty arbitrary (just taken by observing behavior of angular-grid). The real device breakpoints will [likely be different](https://www.freecodecamp.org/news/the-100-correct-way-to-do-css-breakpoints-88d6a5ba1862/). I would prefer not to put the breakpoints here to outside variables, as they're most likely not going to be reused anywhere.

Thank you for the linked articles, I've learned from them and the knowledge will be useful later.

What do you think? Does it make sense?

modules/tribes/client/components/JoinButton.js
simison: Oh, actually we could enable optional chaining JS proposal to make this simpler: https://github.com/Trustroots/trustroots/pull/1158

```js
(user?.memberIds || []).includes(tribe._id))
```

What do you think?

Not opposed to including Lodash in frontend if it makes code more readable but nice if we can avoid it like this.

modules/tribes/client/components/Tribe.js
mrkvon: Fixed in 9fcdd0ab1e89d508057425b433571a1394c42565

modules/tribes/client/components/TribesList.component.js
mrkvon: Fixed in 9fcdd0ab1e89d508057425b433571a1394c42565

modules/tribes/client/components/JoinButton.js
mrkvon: I like this very much!

I just noticed I had used `import get from lodash/get` in the frontend before, so it will be great to replace that with this new syntax, too! :tada:

nicksellen: fyi, for frontend lodash imports `lodash.get` means it doesn't have to load the whole of lodash.

modules/tribes/client/components/TribesList.component.js
simison: > `/* @media (min-width: 0) is just for clarity that width values are overwritten by the next one */`

My preference would be to not include it, since overriding with media queries is so typical in CSS. So;

```css
width: 100%;

@media (min-width: 616px) {
  width: 50%;
}
```

mrkvon: I removed the `max-width` parts of `@media` in 9fcdd0ab1e89d508057425b433571a1394c42565. It works because width is always overwritten by the next media query.
Please give a feedback, especially:
- Does `@media (min-width: 0)` have any clarity benefit, or should we remove it altogether?
- Can we leave the breakpoints hardcoded?

modules/tribes/client/components/TribesList.component.js
mrkvon: Fixed in 9fcdd0ab1e89d508057425b433571a1394c42565.

modules/tribes/client/components/TribesList.component.js
simison: I'd prefer not `min-width:0` and it's fine to leave them hardcoded. 👍

package.json
mrkvon: For completeness: We discussed this on Slack.

Consensus among people involved in the discussion seemed to be to keep the `styled-components` library in experimental mode and see if we like it. If we don't like it we will remove it in the future. Or we will replace it with the fancy, soon-to-come new proposals for css-in-js mentioned above.

modules/tribes/client/components/TribesList.component.js
mrkvon: removed min-width and added some comments in 8f13a627b6f84a5f8af89177045725

modules/tribes/client/components/JoinButton.js
mrkvon: improved with the optional chaining and `array.includes` in 8f13a627b6f84a5f8af89177045725.

I'm happy to have learned the new syntax of #1158! :tada:

mrkvon:
As far as I'm concerned, the review comments are resolved. Thank you for your feedback @nicksellen and @simison! If you like, have another look. Otherwise I'll merge to `tribes-react`after 24 hours.

simison:
These are things good for follow up, but wanted to record some thoughts;

- Instead of splitting columns manually  by percentages, we could use [CSS grid](https://css-tricks.com/snippets/css/complete-guide-grid/) 
- We could even start using some grid component for this and other grids that we have (e.g. https://material-ui.com/components/grid/)

PR #1147, opened 2019-12-12
Add react tests

nicksellen:
~Initial playing around with adding tests for react components. Just a quick play for now...~

This adds react testing in line with the philosophy outlined in the [comment below](https://github.com/Trustroots/trustroots/pull/1147#issuecomment-565741145).

Currently, it means we have two test runners, the existing karma one, and a new jest one. Over time we need to put karma in fire, and just use jest for the client tests.

So, there are two test path patterns now:
* karma tests --> `modules/*/tests/client/*.js` (configured in `karma.conf.js` in `preprocessors` and `files` configuration)
* jest tests --> `modules/*/tests/client/components/*.test.js` (configured in `package.json` at path `jest.testMatch`)

Slightly related, I also fixed the `$templateCache` 404 errors in the tests. I think it had a broken setup for quite a while, the `ngHtml2JsPreprocessor` wasn't even being called (as no html files included in the _files_ section), but we also don't need/use that now, so I removed it. Templates are built in the gulp configuration (called before running the tests) and loaded from `dist/templates.js` (via `main.js`). So they are just always there. With the real templates, not empty strings.

## New/changed commands

```
# run both suites, first karma, then jest
npm run test:client

# run and watch (only) the karma tests
npm run test:client:watch

# run the jest tests
npx jest 

# run and watch jest tests
npx jest --watchAll

# run one specific jest test file
npx jest modules/core/tests/client/components/LanguageSwitch.test.js 

# run and watch one specific jest test file
npx jest --watch modules/core/tests/client/components/LanguageSwitch.test.js 
```

(in the future might put the jest command behind an npm command too...)

## How to review/test

* `git checkout react-tests` (or `git switch react-tests` if you're into the new command, since git v2.23, see https://git-scm.com/docs/git-switch)
* run full client test suite: `npm run test:client` (or full backend too if you want `npm run test`)
* run jest tests: `npx jest`
* optionally try with watching jest tests: `npx jest --watchAll`
* have a peek at how the tests look...
* make _at least_ one important|nitpicky comment for me to change something

## TODO

- [x] investigate https://github.com/testing-library/react-testing-library
- [x] include react tests in main test suite

## TODO later (in other PRs)

- [ ] migrate existing react enzyme tests to this new approach
- [ ] consider migrating ~mocha~ karma/mocha/jasmine -> jest? see https://github.com/skovhus/jest-codemods
- [ ] test a few more react things
- [ ] add useful docs and links into the docs page


nicksellen:
A bit of research about the latest in react testing philosophies.

One thing that seems totally consistent is using jest as the test running, all the rest of this stuff is basically helpers to use within tests... so about philosophy of testing, and the tools that help you there.

This article - [Revisiting React Testing in 2019](https://codeburst.io/revisiting-react-testing-in-2019-ee72bb5346f4) - refers us to 4 libraries:
1. [enzyme](https://airbnb.io/enzyme/) (from airbnb)
2. [react test utils](https://reactjs.org/docs/test-utils.html) (from react team)
3. [react testing library](https://github.com/testing-library/react-testing-library) (from Kent C Dodds)
4. [react test renderer](https://reactjs.org/docs/test-renderer.html) (from react team)

2 now recommends using 3 (or 1 as alternative), so we can eliminate 2 now I think. One of the selling points of 4 is that it also works for react native (as it renders to objects, no DOM involved), which to me, seems like a downside.

Philosophy generally summed up as:
> testing using user facing features instead of implementation details

It points to two articles by Kent C Dodds:
- [Write tests. Not too many. Mostly integration.](https://kentcdodds.com/blog/write-tests)
- [Why I Never Use Shallow Rendering](https://kentcdodds.com/blog/why-i-never-use-shallow-rendering)

Another article - [Testing React Components: The Mostly Definitive Guide (2019)](https://www.valentinog.com/blog/testing-react/) - says:
- you can use [react test renderer](https://reactjs.org/docs/test-renderer.html) to do JSON-based snapshot testing, once a component is stabilized (using react test renderer)
- don't assert on internal state, assert what the user will see
- feels more real to use DOM, so uses [react test utils](https://reactjs.org/docs/test-utils.html)
- do not call a real API in your tests, stub API responses
- suggests looking at react testing library
- this stuff has been for unit testing
- but likes [https://www.cypress.io/](cypress) for functional/e2e testing

## My feeling

Given 2 recommended 3, and the second article mentioned using 2. Using 3 (react testing library) seems a good approach. Everybody seems to be into Kent C Dodds philosophy on testing (which seems reasonable to me too).

I think we could also add cypress tests (and as they are e2e they would presumably work easily for all the existing angular stuff too).

Currently there is karma/mocha/jasmine, but for this PR I would just ignore that and leave them as they are, also leaving off cypress tests for another time/consideration/PR. I did not really look at https://github.com/skovhus/jest-codemods yet... but I suspect it would be quite fiddly, esp going from real browser to fake one. But let's see still worth trying to unify them as soon as possible!

_So in summary_: plan = add jest and react testing library, read Kent C Dodds and focus on testing what the user will see/use, not implementation, mock API call layer if needed. Consider unifying with existing tests if possible.

simison:
>  One of the selling points of 4 is that it also works for react native (as it renders to objects, no DOM involved), which to me, seems like a downside.

Seems relevant mostly for component libraries and design systems, totes not relevant to us.

nicksellen:
... I failed to notice the existing react tests (with enzyme), and everybody was too polite (or forgetful) to remind me ;) (actually @simison did mention something in https://trustroots.slack.com/archives/C0A3Q15SS/p1576176881030500?thread_ts=1576025025.027400&cid=C0A3Q15SS)

e.g. https://github.com/Trustroots/trustroots/blob/master/modules/references/tests/client/CreateReference.client.component.tests.js

Fortunately, there are not so many there, most of the new react components don't have tests yet.

... _and_ if we buy into the logic of the internet people, we should not rely on shallow tests so much. I quite like how these jest + react testing library tests are looking, see https://github.com/Trustroots/trustroots/blob/38316b0c600f748977e9379a1218dc0398c815f9/modules/core/tests/client/components/LanguageSwitch.test.js

Thoughts?

simison:
Ohh so sorry, I also wasn't sure if there were any (I start forgetting what's here and what's elsewhere).

Anyway, I think what you describe is much more powerful and meaningful as far as testing goes so if you all don't either mind ditching Enzyme shallow render tests, rewriting them or just leaving them in for now (my recommendation). It's still worth introducing better setup and maybe just do shallow renders when it actually is meaningful and faster than other kind of testing. Does that make sense?

nicksellen:
Yup, I like the new approach too. I think I would go with the approach to try and get this merged without much further change, just fix up a few broken/failing bits.

mrkvon:
I totally forgot that I was writing those enzyme tests for references! Happy to remove them or rewrite them or have them rewritten to this new approach in near future.

package.json
```diff
@@ -232,5 +236,13 @@
   "bugs": {
     "url": "https://github.com/Trustroots/trustroots/issues"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "jest": {
+    "moduleNameMapper": {
+      "@/(.*)$": "<rootDir>/$1"
```
mrkvon: This must have been a challenge to figure out! :open_mouth:

package.json
```diff
@@ -232,5 +236,13 @@
   "bugs": {
     "url": "https://github.com/Trustroots/trustroots/issues"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "jest": {
```
mrkvon: As the package.json is getting longer, this could also go to a [separate config](https://jestjs.io/docs/en/configuration) (jest.config.js).
Just an info. Optional. :)

simison: I also think separate config file would be helpful for better discoverability.

modules/core/tests/client/components/LanguageSwitch.test.js
```diff
@@ -0,0 +1,67 @@
...
+beforeEach(() => i18n.changeLanguage('en'));
+
+describe('default presentation', () => {
+
+  it('has a menu item for each language', async () => {
+    const { getAllByRole } = render(<LanguageSwitch/>);
+    const items = getAllByRole('menuitem');
+    expect(items).toHaveLength(3);
+    const names = items.map(item => item.innerHTML);
+    expect(names).toEqual(['English', 'česky', 'suomi']);
```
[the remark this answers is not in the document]
nicksellen: I changed to load the list from the `locales.json` file :) I left the explicitly selected languages the same, as that only depends on having the language available, which seems fine to have fixed in the test.

karma.conf.js
```diff
@@ -12,8 +12,7 @@ module.exports = (karmaConfig) => {
     frameworks: ['jasmine'],
     preprocessors: {
       'config/webpack/entries/main.js': ['webpack'],
-      'modules/*/tests/client/**/*.js': ['webpack'],
-      'modules/*/client/views/**/*.html': ['ng-html2js']
+      'modules/*/tests/client/*.js': ['webpack']
```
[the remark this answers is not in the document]
nicksellen: Nope, I removed that bit, as the `../client/component/` subdirectories are test by jest, from this bit of jest config:

```
  testMatch: [
    "<rootDir>/modules/*/tests/client/components/*.test.js"
  ]
```

The others are the karama/mocha-run tests.

config/webpack/entries/main.js
```diff
@@ -8,7 +8,7 @@
 import '@/public/dist/uib-templates';
 import angular from 'angular';
 
-if (process.env.NODE_ENV === 'production') {
+if (process.env.NODE_ENV !== 'development') {
   require('@/public/dist/templates');
```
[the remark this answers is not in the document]
nicksellen: There are three environments, `development`, `test`, and `production` - I changed it here so we include the built templates for the test environment too.

The `npm run test:client` command was already building the templates (by running `gulp angular-templatecache`) but wasn't loading them. Hence all the 404 errors whilst running the tests - it used to mock specific templates that were loaded for specific tests, but that didn't catch all the cases where the templates were loaded.

package.json
nicksellen: Done!

config/webpack/entries/main.js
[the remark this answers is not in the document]
simison: Aside; I wish we could slowly move away from `test` env to simplify things. Thanks for sorting out template loading!

package.json
nicksellen: Haha :) I usually have some other project to refer to for checking those... I think it took a while to setup jest for karrot at various points. Jest promises "configuration free"... but it's not really true... although it does a lot without any config.

package.json
```diff
@@ -48,7 +48,7 @@
     "start:worker": "gulp worker:dev",
     "start": "concurrently --raw --kill-others --kill-others-on-fail 'npm:lint:watch' 'npm:start:develop' 'npm:start:worker' 'npm:dashboard:mail'",
     "test:client:watch": "gulp angular-templatecache && npm run check-environment && NODE_ENV=test karma start",
```
simison: Should you add `jest --watch` here in `test:client:watch`? It should run in parallel with Karma so likely would need to use `concurrently`.

nicksellen: Possibly, I left it out intentionally, but not for a particularly good reason, just to avoid having to think about it :) ... I'll try it out (it would be `jest --watchAll` though).

config/webpack/entries/main.js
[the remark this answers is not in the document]
nicksellen: It's part of jest, so even if we wanted to (I like having the extra env), it would involve fighting jest:

https://github.com/facebook/jest/blob/9419034d6b575fe2c157453fe8b7d2000be66aad/packages/jest-cli/bin/jest.js#L12-L14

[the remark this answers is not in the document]
simison: Oh, interesting! I meant specifically that in our assets/config/Gulp setup we have the 3 envs separately — `test` being basically just copy of `development`: https://github.com/Trustroots/trustroots/blob/6c072ceab0be79b95f893600dc6bb8f1c50a7f53/config/assets/test.js

package.json
simison: Following up in separate PR is fine, totally not a blocker for merging. 👍 

Oh right, I thought `test:client:watch` re-runs only tests that changed but looks like it indeed re-runs everything. :-)

config/webpack/entries/main.js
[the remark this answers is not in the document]
nicksellen: Currently we have configurations that apply to:
- `development` only
- `production` only
- `development` or `test`

There are `development` only things that need to be off in `test`, so only have `production` only (and the inverse) wouldn't be sufficient. 

Perhaps we could just have `development` only (and the inverse).

Although once we get rid of karma/mocha _and_ gulp then becomes less relevant as the jest tests aren't put through webpack.

I'd suggest we just keep it in mind that at some point we could try and switch to only two varieties of env.

package.json
nicksellen: The way jest outputs the results isn't very compatible with running it concurrently with the karma tests as jest clears the console, assuming it's the only thing in the terminal. I looked for some option that made it just output linearly, but no joy.

simison: Ahh, bummer! Thanks for checking.

Suppose folks can just run `npx jest --watch` directly in the root when working on those tests? 😁

karma.conf.js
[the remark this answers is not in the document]
mrkvon: I see. The "old" client tests are indeed all in the root directory. :+1:

package.json
mrkvon: > run `npx jest --watch` directly
Yes. Who wants to run all the tests when doing BDD anyways? :slightly_smiling_face:

I would enjoy having a script in package.json, then. Like `"test:react": "jest"`. Maybe also script for `test:react:watch`, even though it can be achieved with "npm run test:react -- --watchAll".

Not a blocker. :slightly_smiling_face:

nicksellen: I'm slightly hesitant to add extra scripts, as eventually it would use the existing ones (`npm run test:client` and `npm run test:client:watch`) the jest/react ones would only be temporary until the test runners are unified.

Also, `npm run test:client` already runs the jest tests, it's just the watching that doesn't.

So, unless someone would _really_ like extra scripts, I would leave it as it is for now, and hopefully get the test runners unified in not _forever_...

simison: > hopefully get the test runners unified in not forever...

Haha, deal!

modules/core/tests/client/components/LanguageSwitch.test.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  it('can change language', async () => {
+    const { getByRole, getByText } = render(<LanguageSwitch presentation="select"/>);
+    expect(selectedOption(getByRole('combobox'))).toHaveTextContent('English');
+    fireEvent.change(getByRole('combobox'), { target: getByText('česky') });
+    expect(selectedOption(getByRole('combobox'))).toHaveTextContent('česky');
+  });
+
+  it('can save the language to the API', async () => {
```
mrkvon: This test fails when run on its own. (`it` => `it.only` or `fit`)

My debugging suggests the following:
1. The mocked `api.users.update` still remembers having been called with `{ locale: 'cs' }` earlier and that's why it passes when the other test has run.
2. The line 62 (`fireEvent.change(getByRole('combobox'), { target: getByText('česky') });`) does change the select option, but fails to run the `select`'s `onChange` event (which is responsible for calling the api).

I'm going to sleep and look into this tomorrow (or asap later). To the points above:
1. First of all we need to ensure the test fails consistently, i.e. when called together and separately: I think we may be able e.g. to mock only the function `users.update` `beforeEach` and then [restore](https://jestjs.io/docs/en/mock-function-api#mockfnmockrestore) it in `afterEach`. Not sure what's the exact syntax. I've just found it in the jest docs.
2. How to fire the event on `select` so the `onChange` gets fired, I haven't found.

Does this make sense? Is it reproducible?

nicksellen: Ah, good catch! Because the first save to API test makes the call, the mock check passes...

Actually, the error is quite simple, the save to API test for the select is missing the `saveToAPI={true}` option... which is why it never saves.

It's a good idea to reset the mocks too... I'll fix it up tomorrow, unless you want to before then. Gotta sleep now.

mrkvon: The `saveToAPI={true}` was necessary but not sufficient. Additionally, `fireEvent.change(getByRole('combobox'), { target: { value: 'cs' } });` did the trick for me.

nicksellen: Ah, indeed! I fixed it up now, I used:

```js
fireEvent.change(getByRole('combobox'), { target: { value: getByText('česky').value } });
```

... so we are still choosing the value based on the human readable label. To me, that code looks like it should work as:

```js
fireEvent.change(getByRole('combobox'), { target: getByText('česky') });
```

... basically, an object with a value property, but it doesn't work that way.

Select change events seem a bit of a weak point for this testing-library (I read though [this issue](https://github.com/testing-library/react-testing-library/issues/322) when I was originally writing the test, but still got it wrong).

PR #1150, opened 2019-12-14
Move css/less building to webpack, replace livereload with webpack hot reloading

nicksellen:
I wanted to just rip out css/less and move to webpack, but then it lost live reloading, so I added it back in with webpack hot reloading, but that was tricky for the react components, but it works now! (just using very experimental library for now!).

#### Before

Various entries specified in `config/assets/default.js`, split into app and lib categories.

#### After

All via imports which are processed by webpack.

So in main.js there is:
```js
import './main.less';
```
Inside `main.less` it imports all the css files, plus less files from libraries (i.e. bootstrap), and also our global overrides.

Back in `main.js` it then loads all the `*.less` within `modules/`, using this fancy webpack context:
```js
importAll(require.context('../../../modules/', true, /\.less$/));
function importAll(r) {
  r.keys().forEach(r);
}
```
(this makes it similar to how we load the react components just below there)

#### Before

They were all absolute paths which corresponds to the `public/` directory, e.g.:

```css
background-image: url(/img/avatar.png);
```

.. would be loaded by the browser from `<root>/public/img/avatar.png`

#### After

That (should) still works as before, _but_ as the files are imported via webpack it wants to load and process them too, and we use an alias to refer to `public/img` directory (configured in `config/webpack/webpack.config.js`. So, we might write:

```css
background-image: url(~img/avatar.png);
```

Now, webpack will pass that through it's loaders, and make it available at something like `public/assets/images/avatar-3a74587.png` (the hash is generated during the webpack processing, and the original css/less `url()` changed to match.

Also, it will inline the images using data-uri scheme if they are smaller than 10k (configurable, check webpack config again...). Not sure what the best value for that is.

So, everything in `public/assets` (served at `/assets`) is webpack output. In dev mode no files will actually be written, it's all served by webpack-dev-server. In production build, it will write files into `public/assets`.

#### Before

The trustroots server (the node.js express app running the backend) was running on port 3000, and you directly connect to it.

#### After

The node.js app is running on port 3001, and on 3000 is webpack dev server which does two things:
1. any requests to the js/css files it _would_ build, it will serve directly (from memory)
2. anything else it will proxy through to port 3001

#### No refresh on angular template change

Currently we no longer refresh on angular template changes (but will still do full page refreshes for other angular changes, controllers, etc), this will be fixed in the future when we load angular templates through webpack too.

#### Random fix to server test watching

Whilst removing livereload I fixed an issue! It had some code that ran when the server code changed, and it tried to check if it was a server test, and if so, only run that specific test. It didn't really work. It does now.

#### Bleeding edge react hot reloading plugin

The old way of hot loading react is going away (see [this thread](https://github.com/facebook/react/issues/16604)), but the new way didn't really get official webpack support yet. I tried using the old way, but it didn't really work with how we load the react components. So I tried the new way, it worked!

But the plugin to connect the new stuff with webpack (`@pmmmwh/react-refresh-webpack-plugin`) is so new, it has some rough edges, one of them makes incremental compilation _really_ slow, there is [an issue](https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/20#issuecomment-565470955) for it, but for now you, if you want reasonable update speed, you have to comment out the `this.addDependency(path.resolve('./runtime/RefreshModuleRuntime'));` line in `node_modules/@pmmmwh/react-refresh-webpack-plugin/src/loader.js`. I could switch to one of it's forks (e.g. https://github.com/WebHotelier/webpack-fast-refresh), but I think this is a fast moving game... need to wait for it to settle, but maybe this hack is ok for now... it's still faster than the full refresh before probably and I think a fix will come soon.

## How to test

* `git checkout build-styles-in-webpack` (ok I think the work slightly expanded in scope since I picked the branch name!)
* `npm install`
* optional, comment out the `this.addDependency(path.resolve('./runtime/RefreshModuleRuntime'));` line in `node_modules/@pmmmwh/react-refresh-webpack-plugin/src/loader.js` (horrible, hack, see comment just above this for explanation)
* `npm start`
* see if everything seems ok!
* check hot reloading
  * visit http://localhost:3000/volunteering
  * open `./modules/pages/client/components/Volunteering.component.js`
  * edit something, e.g. `<h2>{t('Volunteering')}</h2>` -> `<h2>{t('Volunteering2')}</h2>`
  * see it update in your browser without a refresh!
  * if it doesn't work read the log output (wherever you ran `npm start`), and/or pop open a console in your browser and see if anything looks interesting
  * to try style reloading, visit http://localhost:3000/messages
  * open `./modules/messages/client/less/inbox.less`
  * fiddle around a bit... e.g. a nice `border: 10px solid red` on the `.threadlist`

If you use docker, the same should apply... (except editing the addDependency thing will be more painful.... as node_modules is only inside the container).

## TODO

- [x] integrate into main `npm start` scripts, so it JustWorks
- [x] remove unused less imports (a few I ended up importing higher up the chain... so can be removed from lower down)
- [x] ensure it spits assets out correctly during production builds (`publicPath` settings for style sometimes gets confusing)
- [x] check I removed now-unused libraries from package.json
- [x] check gulp is not watching files it doesn't need to be any more (now webpack does a lot of the watching)
- [x] ~document the need to install the react dev tools for the hot reload (and check first if that's actually true, I think it is...)~ it was a lie, you don't need react dev tools for it
- [x] make sure the docker setup works, re: ports, etc...
- [x] add back the webpack build for the service worker

## TODO in another PR

- [ ] load angular templates through webpack
- [ ] in order to support all the cool webpack splitting options and hashed filenames we need to tell the server which assets to load, I can steal/adapt what I did in foodsharing webpack config - it writes a json file with all the assets, then the server reads it, so it can generate the right js/css tags in the index.html (not needed right now if we stick to static filenames, so probably leave this out for now)
- [ ] we don't have many/any css/js optimization options set, maybe need some more of those...


config/env/development.js
```diff
@@ -13,6 +13,9 @@ module.exports = {
     reference: true,
     i18n: false
   },
+  // in dev we have webpack-dev-server on 3000, and the real server on 3001
+  // @TODO NS consider the docker case
+  port: 3001,
```
simison: You could have this as `parseInt(process.env.FOOBAR || 3001, 10)` so that it's easy to just pass an environment variable to Docker container if it needs to be changed?

config/webpack/webpack.config.js
```diff
@@ -8,21 +14,77 @@ const basedir = join(__dirname, '../..');
...
 
+const isDevelopment = process.env.NODE_ENV === 'development';
 const isProduction = process.env.NODE_ENV === 'production';
 
+const ifDevelopment = (...items) => {
+  return isDevelopment ? items : [];
+};
+
+const ifProduction = (...items) => {
+  return ifProduction ? items : [];
```
simison: Should check `isProduction`, not `ifProduction`?

config/webpack/webpack.config.js
```diff
@@ -39,29 +101,59 @@ module.exports = merge(shims, {
...
+          ...styleLoaders,
+          {
+            loader: 'less-loader'
+          }
+        ]
       }
     ]
   },
   plugins: [
+    ...ifProduction(new MiniCssExtractPlugin({
```
simison: Another pattern I've been using is with [`lodash.compact`](https://lodash.com/docs/4.17.15#compact):

```js
plugins: compact( [
  isProduction && new MiniCssExtractPlugin(),
  new webpack.DefinePlugin(),
  ! isProduction && new ReactRefreshWebpackPlugin(),
] )
```

Up to you which you prefer — personally I like not having helper internal functions and relying on Lodash instead. :-)

It would also simplify if we treat `isProduction` / `isDevelopment` as one constant, so that `! isProduction` would always be `development`?

config/webpack/webpack.config.js
```diff
@@ -8,21 +14,77 @@ const basedir = join(__dirname, '../..');
...
+  }
+];
+
 module.exports = merge(shims, {
   mode: isProduction ? 'production' : 'development',
+  devtool: isProduction ? 'source-map' : 'cheap-module-eval-source-map',
   entry: require.resolve('./entries/main'),
   output: {
-    path: join(basedir, 'public/assets')
+    path: join(basedir, 'public/assets'),
```
simison: I'm not 100% sure but I think Windows is happier with `join(basedir, 'public', 'assets')`

config/env/development.js
nicksellen: I thought about using env variables, but given the configuration is already done via the files, I didn't want to add an additional mechanism - actually that TODO is out of date as I didn't need to change anything for the docker case...

config/webpack/webpack.config.js
nicksellen: I switched to the `compact` approach, I'm pretty agnostic, but like to try new things!

`isProduction` is not the opposite of `isDevelopment`, because of test environment (the react refresh stuff broke stuff when loaded in test environment iirc).

config/webpack/webpack.config.js
nicksellen: I removed the functions entirely.

config/webpack/webpack.config.js
nicksellen: Windows also accepts forward slashes in most places, but I switched it to use multiple args in `join()` anyway.

Dockerfile
```diff
@@ -50,7 +50,6 @@ COPY . /trustroots
 # - Nginx proxy     80
 # - Nodemon server  3000
 # - Node debug      5858
-# - LiveReload      35729
```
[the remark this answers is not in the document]
nicksellen: Thanks, fixed!

gulpfile.js
```diff
@@ -318,15 +216,14 @@ function fontello() {
...
-    .pipe(plugins.refresh());
+    .pipe(gulp.dest('modules/core/client/fonts/fontello'));
 }
 
 function mocha(done) {
   // Open mongoose connections
   const mongooseService = require('./config/lib/mongoose');
   const agenda = require('./config/lib/agenda');
-  const testSuites = changedTestFiles.length ? changedTestFiles : 'modules/*/tests/server/**/*.js';
+  const testSuites = changedTestFiles.length > 0 ? changedTestFiles : 'modules/*/tests/server/**/*.js';
```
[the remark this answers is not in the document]
nicksellen: I switched for clarity, I can never quite remember what is truthy or not, and this seemed clearer to me. I changed it when I was confused what was going on (as it wasn't only re-running the changed files...). This change in itself though doesn't make any practical difference.

config/webpack/webpack.config.js
simison: Cheers! This is one thing I've meant to double-check someday, it's just memory-rule I have from somewhere. I haven't heard anyone try run TR on Windows for a long time so it's not a biggie to keep guarding compatibility. :-)

modules/core/server/views/layout.server.view.html
```diff
@@ -10,14 +10,17 @@
...
+
   {% if env === 'development' %}
-    {% for cssFile in cssFiles %}<link rel="stylesheet" href="/{{cssFile}}?c={{appSettings.commit}}">
-    {% endfor %}
-    <link rel="stylesheet" href="/dist/application.css?c={{appSettings.commit}}">
-  {% endif %}
-  {% if env === 'production' %}
-    <link rel="stylesheet" href="/dist/application.min.css?c={{appSettings.commit}}">
+  {# in dev load js in head, as it also contains CSS, so we prevent flash of unstyled page #}
+  {% for jsFile in jsFiles %}
```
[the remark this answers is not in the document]
nicksellen: I fixed it manually for now... I think some of the indentation style was to make the _output_ html file look nicely indentation, but that's not consistently done either, and doesn't really matter...

gulpfile.js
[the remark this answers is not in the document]
simison: Some folks (and linter presets like `xo`) prefer to check length against zero instead of truthy because in theory `object.length` could be whatever. Not a biggie either way. :-)

nicksellen:
>  I wonder whether this is development only. Then I suppose it's a feature. Otherwise it's an annoyance.

Yes, that would only be for development.

> The dev app fails when there is a linter error. That can be a feature. Anyways it's different from before.

Yes, that is a change (by adding the the eslint-loader as a `enforce: 'pre'` in the webpack config). Some people might find it very annoying though, and want to lint later, if so we could make it a configuration option that you can override if you want.

modules/core/client/less/layout/board.less
```diff
@@ -80,10 +86,10 @@
   .page-header {
     margin: 0;
     h1 {
-      .h1();
+      // .h1(); @TODO find out what happened to this, not defined anywhere
```
simison: Could be coming from Bootstrap less files.

config/lib/express.js
```diff
@@ -55,9 +54,12 @@ module.exports.initLocalVariables = function (app) {
   app.locals.siteAnnouncement = config.siteAnnouncement || { enabled: false };
 
   // Assets
-  if (process.env.NODE_ENV !== 'production') {
+  if (process.env.NODE_ENV === 'production') {
+    app.locals.jsFiles = ['assets/main.js'];
+    app.locals.cssFiles = ['assets/main.css'];
```
simison: Can this be simplified further by just hardcoding these paths in rendered layout?

nicksellen: It could, but I anticipate loading them dynamically in the future, currently we cannot enable webpack file splitting or hashed filenames, because we have no way of telling the backend which files to load.

The way to solve it is to have webpack output a json file listing all the asset files needed, and the backend can read this and render the layout correctly. So, that would keep it kind of similar to what is there, except loading the array from a json file.

Make sense?

modules/core/client/less/layout/board.less
nicksellen: I had looked in bootstrap, but couldn't find it. I found all the other references to bootstrap mixins. Perhaps an older version of bootstrap did define them? Or it's just a misunderstanding (as bootstrap defines .h1, .h2 as _classes_, but not mixins).

I can't actually see anywhere that defines a `.page-header` inside a `.board`, so suspect it's not even being used, but don't want to attempt a CSS cleanup right now...

config/lib/express.js
simison: Yep! Sounds good. :-)

PR #1156, opened 2019-12-31
Refactor OfferLocation React map

simison:
Refactor `OfferLocation` component and create a new base component for all the maps.

**Biggest changes;**
- I think the API of the new library is nicer and more capable, which will be important for refactoring the main search feature to React.
- I think this component structure makes more sense; to create a map you now just need to import our base `Map` component and it takes care of all the basic layers, components, etc. Then you'll add your "content" to the map either as HTML or by creating new `BaseLayer`s:
    ```jsx
    <Map zoom={5} className="my-map">
        <div>
            My content
            ...could be e.g. BaseLayer providing map markers or simple <SVGOverlay />.
        </div>
    </Map>
    ```
    ...you can see this in action in `<OfferLocation>` and `<OfferLocationOverlay>` components.
- Vector tiles 🎉 — much faster and smoother experience compared to bitmap tiles 

Screen Recording 2020-03-15 at 16 28 23 2

#### Proposed Changes

* Create a new base `Map` component and use a different library (([`react-map-gl`](https://www.npmjs.com/package/react-map-gl) instead of [`react-leaflet`](https://www.npmjs.com/package/react-leaflet)).
  * Has fallback capability to render OSM tiles when Mapbox API key isn't provided, so we don't need to be vendor-locked because of this.
  * Use vector tiles instead of bitmap tiles. Vectors look visually better while zooming and it's also possible to rotate (or tilt) the map, although I've disabled that feature for now.
  * Adds [Mapbox attribution](https://docs.mapbox.com/help/how-mapbox-works/attribution/) which we apparently should've had previously.
* Refactor `OfferLocation` component to use above base component (used at profiles' offer section and at "meet" list)
* Adds scale control to OfferLocation map

#### Testing Instructions

1. **Hosting map**
    * Create a hosting offer
    * Open your profile
    * Map should show up under the "Accommodation" section
    Screenshot 2020-03-15 at 15 46 54
    Screenshot 2020-03-15 at 15 44 26

2. **Meet map**
    * Create a meet offer
    * Go to "meet" section and see the map rendered

    Screenshot 2020-03-15 at 15 44 35

3. **Additionally...**
    - Try different screen widths, "yes" and "maybe" hosting offer, try zooming in and out by scrolling, clicking controls, double-clicking...
    - Compared to the maps at the live site
    - When changing screen size (E.g. change browser's width), the map doesn't refresh to new dimensions. I'd like to leave fixing that to a follow-up PR.


modules/core/client/components/withMapControls.component.js
```diff
@@ -0,0 +1,59 @@
...
+import get from 'lodash/get';
+import React, { Component } from 'react';
+
+
+/**
+ * Higher Order Component that configures Trustroots map layers:
+ * - Adds Mapbox API token
+ * - Adds Navigation control
+ */
+const withMapControls = (WrappedComponent) => {
```
simison: Note filename `components/withMapControls.component.js` — other components start with a capital case.

modules/core/client/components/withMapControls.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Higher Order Component that configures Trustroots map layers:
+ * - Adds Mapbox API token
+ * - Adds Navigation control
+ */
+const withMapControls = (WrappedComponent) => {
+  // @TODO: what's up with this?
+  // eslint-disable-next-line react/display-name
+  return class extends Component {
```
nicksellen: Do you need to use the class component form for HOCs?

config/webpack/entries/main.js
```diff
@@ -34,7 +34,7 @@ function importComponents(r) {
   r.keys().forEach(path => {
     const Component = r(path).default;
     const name = extractComponentNameFromPath(path);
-    if (name !== Component.name && process.env.NODE_ENV !== 'production') {
+    if (![name, '_class'].includes(Component.name) && process.env.NODE_ENV !== 'production') {
```
mrkvon: I guess the `_class` as a component name comes from the HOC, which is nameless, does it?

FYI there seems to be [a convention with example](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) for setting a `displayName` of a HOC. However it does not fit the convention for `Component.name` here.

I've also found an old hack which we used to rename `withTranslation(s?)` HOC from `react-i18next` to fit exactly this check. It's not used anymore, because we replaced the HOC with `useTranslation(?s)` hook from the same library. The hack goes as follows:

```jsx
/*


...............A definition of Avatar component here...........................


 */

const AvatarHOC = withNamespaces('user')(Avatar);

AvatarHOC.propTypes = {
  user: PropTypes.object.isRequired,
  size: PropTypes.number,
  source: PropTypes.string,
  link: PropTypes.bool
};

Avatar.propTypes = {
  ...AvatarHOC.propTypes,
  t: PropTypes.func.isRequired
};
// |||||||||||||||||| this is what i mean
// VVVVVVVVVVVVVVVVVV
Object.defineProperty(AvatarHOC, 'name', { value: 'Avatar' });

export default AvatarHOC;
```
Maybe it will be useful for you, perhaps to include in the HOC; so the `Component.name` would be more meaningful.
Anyways this will eventually not matter, as we remove the angularjs completely. And we may just want to stick with [the convention linked above](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging).
I also see this is still a draft PR, so this is just FYI. :slightly_smiling_face:

simison: >FYI there seems to be a convention with example for setting a displayName of a HOC. However it does not fit our convention for Component.name here.

Yep, on todo, I think we'll just need first [`babel-plugin-proposal-class-properties`](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties).

modules/core/client/components/Map/osm.json
```diff
@@ -0,0 +1,23 @@
+{
```
simison: This file defines OSM tile layer.

modules/core/client/components/Map/osm.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": 8,
+  "name": "OpenStreetMap",
+  "metadata": {},
+  "sources": {
+    "osm": {
+      "attribution": "<a href=\"https://www.openstreetmap.org/about/\" target=\"_blank\" rel=\"noopener\">&copy; OpenStreetMap</a>",
+      "mapbox_logo": true,
```
simison: We wouldn't _need_  to show the Mapbox logo with this layer, but I've left it here so that it's visible while developing and we know to take into account as if we'd be on production site.

public/locales/en/languages.json
```diff
@@ -1,3 +1,3 @@
 {
     "Polish": "Polish"
-}
\ No newline at end of file
+}
```
simison: 🤷‍♂

modules/offers/client/views/offer-meet-list.client.view.html
```diff
@@ -46,7 +46,10 @@ <h3>Your meetups</h3>
   >
     <div class="panel-body">
       <!-- The map (React component) -->
-      <offer-location offer="offer"></offer-location>
+      <offer-location
+        location="offer.location"
+        offerType="offer.type"
```
simison: Note how we're more explicit with what we pass here.

modules/core/client/components/Map/osm.json
mrkvon: > I'm not planning on showing OSM in production anyway.

:+1: [OSM tile policy](https://operations.osmfoundation.org/policies/tiles/) disallows distributing an app which use OSM tiles and other heavy uses.

simison: Yup, exactly. :-)

mrkvon:
image

The bitmap tiles from OSM (development) look a bit unfocused. Maybe it doesn't matter, since we won't use bitmap tiles in production anyways?
If we wish to use bitmap tiles as some kind of fallback, this would be worth looking into.

simison:
> The bitmap tiles from OSM (development) look a bit unfocused. 

I noticed the same but wasn't too worried since it's for development. Would be nice to solve it tho but I figured it's ok for a follow up too.

modules/core/client/components/Map/map.less
```diff
@@ -0,0 +1,52 @@
+
+@import '~/node_modules/mapbox-gl/dist/mapbox-gl.css';
```
nicksellen: nitpick: I don't think you need `node_modules` in the path here, it resolves inside, maybe just `~mapbox-gl/dist/mapbox-gl.css`.

modules/offers/client/components/OfferLocationOverlay.js
```diff
@@ -0,0 +1,70 @@
+// External dependencies
+import { BaseControl } from 'react-map-gl';
+import { SVGOverlay } from 'react-map-gl';
```
nicksellen: nitpick: imports can be combined
```js
import { BaseControl, SVGOverlay } from 'react-map-gl';
```

simison: Oh silly me! We better enable [`no-duplicate-imports`](https://eslint.org/docs/rules/no-duplicate-imports).

modules/core/client/components/Map/map.less
simison: Fixed 👍

simison: Fixed 👍

modules/offers/client/components/OfferLocationOverlay.js
simison: PRed: https://github.com/Trustroots/trustroots/pull/1306

PR #1157, opened 2020-01-01
Convert react enzyme tests into jest + testing-library ones

nicksellen:
This converts the react enzyme tests to use jest and testing-library. They are pretty much 1-to-1 conversions, although a few more cases of using label text to interact.

As part of it I also added some "role" attributes which allows the testing library to find the elements and also improves accessibility.


modules/references/client/components/create-reference/Navigation.js
```diff
@@ -11,6 +11,7 @@ export function Navigation({ t, disabled, tab, tabs, tabDone, onBack, onNext, on
   const backButton = (
     <button
       type="button"
+      role="button"
```
simison: I think `<button>` elements might already be getting button role by default? 

See ["Permitted ARIA roles" list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button).

nicksellen: Yes, you're right, let me see if it finds it without it explicitly set. I couldn't see where it said that in your link, but in this one it says it clearly: https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#the-button-element

modules/references/tests/client/components/CreateReference.client.component.test.js
```diff
@@ -0,0 +1,147 @@
...
+} from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import * as references from '@/modules/references/client/api/references.api';
+const api = { references };
+
+jest.mock('@/modules/references/client/api/references.api');
+afterEach(() => jest.clearAllMocks());
+
+import CreateReference from '@/modules/references/client/components/CreateReference.component';
```
simison: Nit; I think it would be nice to group imports on top before any other code as a convention? I see benefit of having related imports+code grouped together, too, but I'd lean towards grouping imports anyway.

Not a biggie either way.

modules/references/client/components/create-reference/Navigation.js
nicksellen: Yup, still works without that, so I removed that, thanks!

docs/React.md
```diff
@@ -100,7 +101,7 @@ Read [a manual](i18n.md).
 
 ## Testing
 
-Testing is done with `Jasmine` and [`Enzyme`](https://github.com/airbnb/enzyme).
+Testing is done with [`jest`](https://jestjs.io/en/) and [`React Testing Library`](https://testing-library.com/docs/react-testing-library/intro).
 
 TODO add an example when references are in master.
```
simison: > TODO add an example when references are in master.

Could be removed — I think the idea was to link to some examples. Maybe writing what's the typical path for test files would be enough?

simison: Did you have a nice summary about testing somewhere in chat or Github? It might have some links or other useful bits to add here?

modules/references/tests/client/components/CreateReference.client.component.test.js
nicksellen: Agree, even better we can add https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/first.md and https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md

docs/React.md
nicksellen: Yup, added a bit more now (from https://github.com/Trustroots/trustroots/pull/1147#issuecomment-565741145)

modules/references/tests/client/components/CreateReference.client.component.test.js
nicksellen: I added extending from `plugin:import/errors` and upgrading `import/first` to an error. In the future we can tighten it up with `import/order` too, but that makes too many unrelated changes for this PR.

docs/React.md
nicksellen: Yup, I added the path matcher and a link to a test.

PR #1158, opened 2020-01-02
Build: Add support for optional chain and null coalesce

simison:
This came up at https://github.com/Trustroots/trustroots/pull/1144#discussion_r362422022

#### Proposed Changes

Enables two stage 3 proposals:
- [**Optional chaining**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) via [`@babel/plugin-proposal-optional-chaining`](https://babeljs.io/docs/en/next/babel-plugin-proposal-optional-chaining).
- [**Nullish coalescing operator**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) via [`@babel/plugin-proposal-nullish-coalescing-operator`](https://babeljs.io/docs/en/next/babel-plugin-proposal-nullish-coalescing-operator).

This should reduce the use of `lodash.has()` and `lodash.get()` (and `demo && demo.something && demo.something.else` style tests) in the codebase, which is important especially for frontend code to avoid bundle ballooning bigger.

#### Testing Instructions

* Test that you can use examples from MDN articles in the clientside code.

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -38,7 +38,9 @@ module.exports = merge(shims, {
               ['@babel/preset-react']
             ],
             plugins: [
+              '@babel/plugin-proposal-nullish-coalescing-operator',
               '@babel/plugin-proposal-object-rest-spread',
+              '@babel/plugin-proposal-optional-chaining',
               'angularjs-annotate'
             ]
           }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1339,6 +1339,16 @@
         "@babel/plugin-syntax-json-strings": "^7.7.4"
       }
     },
+    "@babel/plugin-proposal-nullish-coalescing-operator": {
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-nullish-coalescing-operator/-/plugin-proposal-nullish-coalescing-operator-7.7.4.tgz",
+      "integrity": "sha512-TbYHmr1Gl1UC7Vo2HVuj/Naci5BEGNZ0AJhzqD2Vpr6QPFWpUmBRLrIDjedzx7/CShq0bRDS2gI4FIs77VHLVQ==",
+      "dev": true,
+      "requires": {
+        "@babel/helper-plugin-utils": "^7.0.0",
+        "@babel/plugin-syntax-nullish-coalescing-operator": "^7.7.4"
+      }
+    },
     "@babel/plugin-proposal-object-rest-spread": {
       "version": "7.7.7",
       "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-object-rest-spread/-/plugin-proposal-object-rest-spread-7.7.7.tgz",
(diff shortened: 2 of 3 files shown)


mrkvon:
Optional chaining is [Stage 4](https://tc39.es/proposal-optional-chaining/) ([Finished](https://tc39.es/process-document/)).
How safe is it to use Stage 3? What are the odds that it will land to EcmaScript?

simison:
Oh, it already made to 4, cool!

[It's not common to withdraw things at stage 3 anymore (but it does happen)](https://stackoverflow.com/a/48150001), nor they're very likely to change. [Via](https://babeljs.io/docs/en/babel-preset-stage-3):

> changes should only be made in response to critical issues raised by the implementations and their use.

I'd personally be pretty confident with this considering that refactoring back would likely be very straightforward. Happy to limit only to optional chaining too if you prefer. :-)

nicksellen:
I like these, although one risk is they get sprinkled everywhere in a lazy way "just in case"... which can just end up hiding some underlying data/config errors.

simison:
> one risk is they get sprinkled everywhere in a lazy way "just in case"... which can just end up hiding some underlying data/config errors.

True; let's keep reviewing each others code for this. :-)

mrkvon:
Do we have eslint rules already in?

simison:
Sorry, which eslint rules you were thinking specifically? I didn't think
we'd need any for this. :-)

mrkvon:
I thought linter would be confused and complain about a syntax error. Will check in a while; or tomorrow.

simison:
@mrkvon I see now what you meant, this indeed fails linting when running `npm run lint`, even if in our Eslint config we have `parser: 'babel-eslint'` so I'd expect it to be able lint code just like it's able to parse it when building.

This kind of code doesn't file in my Atom editor probably because it uses `atom-eslint` or something like that, which might work differently.

I just read that `babel-eslint@11` will behave more like I'd expect: https://github.com/babel/babel-eslint/tree/249a2947a24428ac9306d2780121201a9e27b77d#breaking-change-in-v11xx

> As of the v11.x.x release, babel-eslint now requires Babel as a peer dependency and expects a valid Babel configuration file to exist. This ensures that the same Babel configuration is used during both linting and compilation.

...but even if I update `babel-eslint`, it didn't help. Not entirely sure how to make Eslint happy. :-/

nicksellen:
I have no linting errors, note in master now there is a `babel.config.js`, which came from one of my PRs, as I needed to share the babel config with webpack and jest, so a proper babel config is better.

simison:
> I have no linting errors

Uff, that was my bad, my test string was faulty. 😝
```
  6:14  error  Parsing error: modules/admin/client/components/Admin.component.js: Unexpected token (6:14)

  4 |
  5 | export default function Admin() {
> 6 |   const window?.test;
    |               ^
  7 |   return (
  8 |     <>
  9 |       <AdminHeader />

✖ 1 problem (1 error, 0 warnings)
```

Adding this to any client-side JS file and running `npm run lint` shows no errors so all should be well, I think:

```js
const test = window.test?.test; // eslint-disable-line no-unused-vars
```

Otherwise this would show "Unexpected token".

mrkvon:
Impressive! Our linter already knows how to handle this syntax!

simison:
Nulliash seems to be at stage four and going to be on by default in latest Babel: https://github.com/babel/babel/issues/10813

PR #1159, opened 2020-01-02
More webpack and jest

nicksellen:
#### Proposed Changes

This moves the last bits of frontend asset loading into webpack, and moves the existing tests into jest. Maybe it seems like they are unrelated, but for jest I needed to move to a lot of explicit imports, and so why not explicitly import the templates too!

It's a bit of a beast, as it totally changes how the files are loaded.

It's also a bit messy, as I went through a few variations on how to do it en-route...

As it now depends on explicit importing, it means the client modules need to import all their controllers, this is a bit painful to do manually, so I wrote a script to do code generation. We could not include the script, and just do it as a one-time thing, I guess there aren't going to be many more angular things created. Up for discussion.

#### TODO

- [ ] do the TODOs I added...
- [x] remove karma stuff entirely
- [ ] tidy up
- [ ] rebase to remove a bit of mess in history
- [ ] think/discuss about the code generation stuff
- [x] convert the remaining ng-includes to webpack imports (and check they work...)
- [ ] investigate missing templates

#### Testing Instructions

* for dev: `npm start` then try using it, everything should work as normal!
* for tests: `npm test` / `npm test:client` / `npm test:client:watch` should all work as expected
* for prod: `npm build` then `npm run start:prod` should work as expected, no console errors


modules/core/client/app/config.js
```diff
@@ -1,82 +1,107 @@
...
+import 'chosen-js/chosen.jquery.js';
+import 'angular-chosen-localytics/dist/angular-chosen.js';
+
 import ngreact from 'ngreact';
 
+// @TODO can I import it properly in tests? seems tricky as it expects this.angular to exist
+if (process.env.NODE_ENV === 'test') {
+  angular.module('zumba.angular-waypoints', []);
+} else {
+  require('angular-waypoints/dist/angular-waypoints.all');
```
nicksellen: I think mocking this away with jest would be a better solution.

nicksellen:
Closed in favour of https://github.com/Trustroots/trustroots/pull/1167

PR #1160, opened 2020-01-02
Move around utils

simison:
Mainly wanted to pull some unrelated changes out from https://github.com/Trustroots/trustroots/pull/1161

#### Proposed Changes

* Move generic utils to the core module
* Simplify `isWarmshowersId()` input attribute
* Small ES6'ifying

#### Testing Instructions

* Profile works as usual

### DIFF
--- a/modules/users/client/components/ProfileViewBasics.js
+++ b/modules/users/client/components/ProfileViewBasics.js
@@ -4,7 +4,7 @@ import { withTranslation } from 'react-i18next';
 import { Trans } from 'react-i18next';
 import PropTypes from 'prop-types';
 import * as languages from '@/config/languages/languages';
-import { hasConnectedAdditionalSocialAccounts, isWarmshowersId, socialAccountLink } from './utils/networks';
+import { hasConnectedAdditionalSocialAccounts, isWarmshowersId, socialAccountLink } from '../utils/networks';
 
 
 export function ProfileViewBasics({ t, profile }) {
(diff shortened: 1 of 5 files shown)


modules/users/client/utils/networks.js
```diff
@@ -2,34 +2,30 @@
...
-    return 'https://github.com/' + providerData.login;
+    return `https://github.com/${providerData.login}`;
   } else {
     return '#';
   }
 }
-
-
-
-module.exports = { hasConnectedAdditionalSocialAccounts, socialAccountLink, isWarmshowersId };
```
simison: We should probably be careful with this kind of object exports. It works because in ES6 I can pull keys from an object. I don't think these are equal in terms of code-splitting some day:

```js
const foo;
export { foo };
// ...
import { foo } from '.';
```

```js
export const foo;
// ...
import { foo } from '.';
```

...not entirely sure about this tho and didn't start dig deeper.

modules/users/client/components/ProfileViewBasics.js
```diff
@@ -4,7 +4,7 @@ import { withTranslation } from 'react-i18next';
 import { Trans } from 'react-i18next';
 import PropTypes from 'prop-types';
 import * as languages from '@/config/languages/languages';
-import { hasConnectedAdditionalSocialAccounts, isWarmshowersId, socialAccountLink } from './utils/networks';
```
simison: There was an existing pattern for `modules/**/client/utils/**`

https://github.com/Trustroots/trustroots/blob/70ebf948764773832dad6287e4bf15f077093bce/.eslintrc.js#L171

PR #1161, opened 2020-01-02
Add "read more" component

simison:
I PRed some related file-movings and things in https://github.com/Trustroots/trustroots/pull/1160 which can be merged first. (update: done!)

#### Proposed Changes

* Extract "Read more" feature from profile to its own component. Later own this will be useful in some offer components (there's a similar mechanism in https://github.com/Trustroots/trustroots/pull/1045) 
* Simplify `AboutMe` component
* Switch text to say "Read more…" instead of "Show more..."

#### Testing Instructions

* Create over 2400 chars long profile description
* Open profile, and observe this functionality work just fine:
  Screenshot 2020-01-03 at 00 22 17
  Screenshot 2020-01-03 at 00 22 21

* Test also with an empty description or short description.

### DIFF
--- a/modules/core/client/components/ReadMorePanel.js
+++ b/modules/core/client/components/ReadMorePanel.js
@@ -20,11 +20,6 @@ export class ReadMorePanel extends Component {
     this.setState({ showMore: true });
   }
 
-  componentDidCatch(error, errorInfo) {
-    // You can also log the error to an error reporting service
-    console.error(error, errorInfo); //eslint-disable-line
-  }
-
   render() {
     const { showMore } = this.state;
     const { content, id } = this.props;
--- a/modules/core/client/components/ReadMorePanel.js
+++ b/modules/core/client/components/ReadMorePanel.js
@@ -2,6 +2,9 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 
+// Internal dependencies
+import { plainTextLength } from '../utils/filters';
+
 export class ReadMorePanel extends Component {
   constructor(props) {
     super(props);
@@ -28,7 +31,7 @@ export class ReadMorePanel extends Component {
       return;
     }
 
-    if (showMore || content.length <= limit) {
+    if (showMore || plainTextLength(content) <= limit) {
       return <div dangerouslySetInnerHTML={{ __html: content }} />;
     }
 
--- a/modules/core/client/components/ReadMorePanel.js
+++ b/modules/core/client/components/ReadMorePanel.js
@@ -45,5 +45,4 @@ export default function ReadMorePanel({ content, id }) {
 ReadMorePanel.propTypes = {
   content: PropTypes.string.isRequired,
   id: PropTypes.string.isRequired,
-  t: PropTypes.func.isRequired,
 };
--- a/modules/core/client/components/ReadMorePanel.js
+++ b/modules/core/client/components/ReadMorePanel.js
@@ -1,7 +1,7 @@
 // External dependencies
 import { withTranslation } from 'react-i18next';
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 
 // Internal dependencies
 import '@/config/client/i18n';
@@ -9,51 +9,36 @@ import { plainText, plainTextLength } from '../utils/filters';
 
 const LIMIT = 2000;
 
-export class ReadMorePanel extends Component {
-  constructor(props) {
-    super(props);
-    this.toggleMore = this.toggleMore.bind(this);
-    this.state = {
-      showMore: false
-    };
+function ReadMorePanel({ content, id, t }) {
+  const [showMore, setShowMore] = useState(false);
+
+  if (content.length === 0) {
+    return;
   }
 
-  toggleMore() {
-    this.setState({ showMore: true });
+  if (showMore || plainTextLength(content) <= LIMIT) {
+    return <div dangerouslySetInnerHTML={{ __html: content }} />;
   }
 
-  render() {
-    const { showMore } = this.state;
-    const { content, id, t } = this.props;
-
-    if (content.length === 0) {
-      return;
-    }
-
-    if (showMore || plainTextLength(content) <= LIMIT) {
-      return <div dangerouslySetInnerHTML={{ __html: content }} />;
-    }
-
-    return (
-      <div className="panel-more-wrap">
-        <div
-          className="panel-more-wrap"
-          dangerouslySetInnerHTML={{ __html: `${plainText(content).substr(0, LIMIT)} …` }}
-          id={id}
-          onClick={this.toggleMore}
-        />
-        <div
-          aria-controls={id}
-          aria-expanded="false"
-          className="panel-more-fade"
-          onClick={this.toggleMore}
-          type="button"
-        >
-          {t('Read more…')}
-        </div>
+  return (
+    <div className="panel-more-wrap">
+      <div
+        className="panel-more-wrap"
+        dangerouslySetInnerHTML={{ __html: `${plainText(content).substr(0, LIMIT)} …` }}
+        id={id}
+        onClick={() => setShowMore(true)}
+      />
+      <div
+        aria-controls={id}
+        aria-expanded="false"
+        className="panel-more-fade"
+        onClick={() => setShowMore(true)}
+        type="button"
+      >
+        {t('Read more…')}
       </div>
-    );
-  }
+    </div>
+  );
 }
 
 ReadMorePanel.propTypes = {
--- a/modules/core/client/components/ReadMorePanel.js
+++ b/modules/core/client/components/ReadMorePanel.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 
 // Internal dependencies
-import { plainTextLength } from '../utils/filters';
+import { plainText, plainTextLength } from '../utils/filters';
 
 export class ReadMorePanel extends Component {
   constructor(props) {
@@ -39,7 +39,7 @@ export class ReadMorePanel extends Component {
       <div className="panel-more-wrap">
         <div
           className="panel-more-wrap"
-          dangerouslySetInnerHTML={{ __html: content.substr(0, limit) }}
+          dangerouslySetInnerHTML={{ __html: plainText(content).substr(0, limit) }}
           id={id}
           onClick={this.toggleMore}
         />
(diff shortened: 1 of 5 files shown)


modules/core/client/components/ReadMorePanel.js
```diff
@@ -0,0 +1,71 @@
...
+
+    if (showMore || plainTextLength(content) <= limit) {
+      return <div dangerouslySetInnerHTML={{ __html: content }} />;
+    }
+
+    return (
+      <div className="panel-more-wrap">
+        <div
+          className="panel-more-wrap"
+          dangerouslySetInnerHTML={{ __html: plainText(content).substr(0, limit) }}
```
simison: Cutting out plain text instead of HTML will avoid us breaking HTML tags from the middle.

modules/core/client/components/ReadMorePanel.js
```diff
@@ -0,0 +1,71 @@
...
+          onClick={this.toggleMore}
+        />
+        <div
+          aria-controls={id}
+          aria-expanded="false"
+          className="panel-more-fade"
+          onClick={this.toggleMore}
+          type="button"
+        >
+          {/* TODO: translate */}
```
simison: I actually added `withTranslate` here but it was failing with `t is not function`. Will have to look tomorrow again with fresh eyes.

modules/core/client/components/ReadMorePanel.js
```diff
@@ -0,0 +1,71 @@
+// External dependencies
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+
+// Internal dependencies
+import { plainText, plainTextLength } from '../utils/filters';
+
+export class ReadMorePanel extends Component {
```
nicksellen: Why a class component instead of a function?

simison: Just to be able to have `state.showMore `

simison:
Rebased and simplified tad bit by removing customizable "limit" prop. Just having limit hardcoded in 2000 chars seemed fine.

modules/core/client/components/ReadMorePanel.js
simison: @mrkvon haha, I'm again with sleepy eyes. I pushed the translation code now and still getting `t is not a function` — can you spot what I'm doing wrong?

modules/core/client/components/ReadMorePanel.js
nicksellen: Doesn't the `useState` hook address that? It's used in the newer react components @mrkvon has done. Or am I missing something?

simison: Oh cool! Nah you're right, it's just me who hasn't learned React hooks yet and rolls like it's still 2019. 😁

simison: Updated in https://github.com/Trustroots/trustroots/pull/1161/commits/662494a65c2a0033e1b0f83e8b1079a186481104 Thanks for the heads up.

modules/core/client/components/ReadMorePanel.js
simison: Yay, this started working in https://github.com/Trustroots/trustroots/pull/1161/commits/662494a65c2a0033e1b0f83e8b1079a186481104.

modules/core/client/components/ReadMorePanel.js
```diff
@@ -0,0 +1,50 @@
...
+
+function ReadMorePanel({ content, id, t }) {
+  const [showMore, setShowMore] = useState(false);
+
+  if (content.length === 0) {
+    return null;
+  }
+
+  if (showMore || plainTextLength(content) <= LIMIT) {
+    return <div dangerouslySetInnerHTML={{ __html: content }} />;
```
simison: Just noting that things are sanitized at the backend before outputting via APIs.

modules/core/client/components/ReadMorePanel.js
```diff
@@ -0,0 +1,50 @@
...
+        onClick={() => setShowMore(true)}
+      />
+      <div
+        aria-controls={id}
+        aria-expanded="false"
+        className="panel-more-fade"
+        onClick={() => setShowMore(true)}
+        type="button"
+      >
+        {t('Read more…')}
```
mrkvon: The production now says `Show more...`. Is the change here intentional?

modules/users/client/components/AboutMe.component.js
```diff
@@ -1,54 +1,41 @@
...
-  }
-  changeProfileDescriptionToggle(){
-    this.setState((prevState) => ({
-      profileDescriptionToggle: !prevState.profileDescriptionToggle,
-    }));
-  }
+// Internal dependencies
+import '@/config/client/i18n';
+import { plainTextLength } from '@/modules/core/client/utils/filters';
+import { withTranslation } from '@/modules/core/client/utils/i18n-angular-load';
```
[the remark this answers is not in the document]
simison: Starting to love these hooks! :-) Switched to using it.

modules/core/client/utils/filters.js
```diff
@@ -1,9 +1,12 @@
-export const limitTo = (text, length) => {
-  return text.substring(0, length);
+export const plainText = (html) => {
+  if (!html || typeof html !== 'string' || typeof document !== 'undefined') {
```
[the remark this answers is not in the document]
simison: Yess, indeed! Fixed.

modules/core/client/components/ReadMorePanel.js
simison: Yup, I switched to "read more" as it felt more natural in this context where a member wants to see more text to read more.

modules/core/client/components/ReadMorePanel.js
```diff
@@ -0,0 +1,50 @@
...
+
+  if (showMore || plainTextLength(content) <= LIMIT) {
+    return <div dangerouslySetInnerHTML={{ __html: content }} />;
+  }
+
+  return (
+    <div className="panel-more-wrap">
+      <div
+        className="panel-more-wrap"
+        dangerouslySetInnerHTML={{ __html: `${plainText(content).substr(0, LIMIT)} …` }}
```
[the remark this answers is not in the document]
simison: Good question! It's not desirable from a user experience / visual point of view. It's kinda mandatory if trying to avoid breaking HTML tags in the middle, tho. :-/

Consider e.g. breaking anywhere in the middle of `<a href="#">Hello</a>` and how destructive that might be. There could be some XSS type scenarios as well, although I can't think of any very harmful examples right now.

I'd like to somehow at least avoid breaking `<p><br/>` tho — I think that's avoidable quite simply.

That said, your suggestion to cut it with CSS is the "right" way to go about this: https://github.com/Trustroots/trustroots/pull/1161#pullrequestreview-338366175

> We could consider a different approach: Show the whole content, but wrap it in a block element with limited max-height and overflow: hidden. Then we would hide content based on content height, not its length.
>
> [This library](https://yarnpkg.com/en/package/react-overflow) could help with detecting whether the content overflows. However it has few people using it.

It started seeming a little more fiddly when considering different screen widths/heights/mobile affecting the panel width and calculating different height on each of those cases. It basically would need some kind of "max-height limit + overflow detector" to bail out from showing the "read more" if the text isn't overflowing, as you mention.

The actual logic for the component you linked is fairly simple, so just copying relevant bits to this component seems simple enough:

https://github.com/nickuraltsev/react-overflow/blob/b6ff220a23f60e84adf472882901eee6f149634a/src/OverflowDetector.jsx#L37-L48

Do you think it would be fine to merge this as is and loop back to bringing back the formatting within a few weeks?

modules/core/client/components/ReadMorePanel.js
simison: Related convo: https://github.com/Trustroots/trustroots/pull/1161#discussion_r363119558

modules/core/client/components/ReadMorePanel.js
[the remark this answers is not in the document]
simison: Good question! It's not desirable from a user experience / visual point of view. It's kinda mandatory if trying to avoid breaking HTML tags in the middle, tho. :-/

Consider e.g. breaking anywhere in the middle of `<a href="#">Hello</a>` and how destructive that might be. There could be some XSS type scenarios as well, although I can't think of any very harmful examples right now.

I'd like to somehow at least avoid breaking `<p><br/>` tho — I think that's avoidable quite simply.

That said, your suggestion to cut it with CSS is the "right" way to go about this: https://github.com/Trustroots/trustroots/pull/1161#pullrequestreview-338366175

> We could consider a different approach: Show the whole content, but wrap it in a block element with limited max-height and overflow: hidden. Then we would hide content based on content height, not its length.
>
> [This library](https://yarnpkg.com/en/package/react-overflow) could help with detecting whether the content overflows. However it has few people using it.

It started seeming a little more fiddly when considering different screen widths/heights/mobile affecting the panel width and calculating different height on each of those cases. It basically would need some kind of "max-height limit + overflow detector" to bail out from showing the "read more" if the text isn't overflowing, as you mention.

The actual logic for the component you linked is fairly simple, so just copying relevant bits to this component seems simple enough:

https://github.com/nickuraltsev/react-overflow/blob/b6ff220a23f60e84adf472882901eee6f149634a/src/OverflowDetector.jsx#L37-L48

Do you think it would be fine to merge this as is and loop back to bringing back the formatting within a few weeks?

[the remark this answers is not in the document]
mrkvon: > Do you think it would be fine to merge this as is and loop back to bringing back the formatting within a few weeks?

If we merge to production in the meantime, users will be confused. Otherwise it's fine I think. So I guess decide based on the production plans... :wink:

modules/users/client/components/AboutMe.component.js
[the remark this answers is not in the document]
mrkvon: This `withTranslation` is still hiding here. :wink: (not blocking)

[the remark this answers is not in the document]
simison: Ha! Cleaned up https://github.com/Trustroots/trustroots/pull/1161/commits/c611666a130d867ae695e6f34cb48ab409b90ead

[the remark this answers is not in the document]
simison: Ha! Cleaned up https://github.com/Trustroots/trustroots/pull/1161/commits/c611666a130d867ae695e6f34cb48ab409b90ead

simison:
Thanks, folks!

PR #1162, opened 2020-01-04
Member search: split into separate reusable components

simison:
#### Proposed Changes

* Split member search component into smaller, more generic, easier to re-use components:
  - Add `LoadingIndicator` component
  - Add `NoContent` component
  - I want to eventually merge [`<ContactPresentational />`](https://github.com/Trustroots/trustroots/blob/master/modules/contacts/client/components/ContactPresentational.js) with `<UserSummary />`
* Use CSS grid instead of Bootstrap columns and manually splitting arrays into two
* Simplified some components while at it as well.

#### Testing Instructions

* Go to `http://localhost:3000/search/members` and test that results work (try e.g. with your logged-in username) and that "no results" work as well. Try with data that give 3+ results.
    Screenshot 2020-01-19 at 21 48 42

* Go to profile's contacts list tab (add some contacts first) — for the search bar to appear you need to have a bunch) and confirm that list and filtering works
    Screenshot 2020-01-19 at 21 47 04

* No visual changes nowhere


modules/users/client/components/UserSummary.js
```diff
@@ -0,0 +1,43 @@
+// External dependencies
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from 'styled-components';
+
+// Internal dependencies
+import Avatar from '@/modules/users/client/components/Avatar.component';
+
+export default function UserSummary({ user }) {
+  const UserSummary = styled.div.attrs({
```
nicksellen: This caused me to do a double take, a UserSummary inside a UserSummary? I can't think of a better name, maybe just Container? What are the other styled components called?

simison: Haha, you're right. I saw `Container` in @mrkvon 's PR as well; seems indeed like a good convention to have.

mrkvon: [Try to define the styled component outside the component it's used in.](https://www.styled-components.com/docs/faqs#why-should-i-avoid-declaring-styled-components-in-the-render-method)

modules/search/client/components/UsersResults.js
```diff
@@ -0,0 +1,30 @@
...
+    return (
+      <NoContent icon="users" message="No members found by this name." />
+    );
+  }
+
+  return (
+    <div className="row">
+      <div className="col-xs-12">
+        <h4 className="text-muted">
+          {users.length === 1 ? 'One member found.' : `${users.length} members found.`}
```
mrkvon: We could add translations here, and use https://www.i18next.com/translation-function/plurals.

modules/users/client/components/UsersList.js
```diff
@@ -0,0 +1,33 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from 'styled-components';
+
+// Internal dependencies
+import UserSummary from './UserSummary';
+
+export default function UsersList({ users }) {
+  const UsersList = styled.div`
```
mrkvon: Now, this is nitpick: We could consider using `ul` and `li` to identify a list in html.

modules/users/client/components/UsersList.js
```diff
@@ -0,0 +1,33 @@
...
+    @media (min-width: 616px) {
+  	   grid-template-columns: 1fr 1fr;
+    }
+  `;
+
+  return (
+    <UsersList>
+      {users.map((user) => (
+        <UserSummary
+          className="user-summary"
```
mrkvon: I believe this `className` won't have any effect, because `UserSummary` doesn't expect it nor pass it down to its html.

mrkvon:
Tests great! :tada:

I wonder why we have no links to this feature from other places. Is that on purpose?
AFAIK there was a broken database index so it didn't work for a long time, right? But now it does and i love it.
Has some privacy implications though. Is that why it's still hidden?

simison:
Had a look if NoContent component could have `aria-role="alert"` but this doesn't make it sound like a good fit:

> The alert role should read out content that has changed, or bring the user's attention to it immediately, so it should not be used for static content or used regularly. Alerts, by definition, are disruptive. Lots of alerts at once or unnecessary alerts willl create a bad user experience.  

via https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role

modules/users/client/components/UsersList.js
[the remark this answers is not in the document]
simison: Thanks for the heads up! :-) 

Updated.

modules/users/client/components/UsersList.js
simison: Good catch! Fixed.

modules/search/client/components/UsersResults.js
simison: Updated in https://github.com/Trustroots/trustroots/pull/1162/commits/a4cecb07cb9711984bc3ef3cc2ee0c5821336add

modules/search/client/components/UsersResults.js
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react';
+
+// Internal dependencies
+import UsersList from '@/modules/users/client/components/UsersList';
+import NoContent from '@/modules/core/client/components/NoContent';
+
+export default function UsersResults({ users }) {
+  if (!users || users.length === 0) {
+    return (
+      <NoContent icon="users" message="No members found by this name." />
```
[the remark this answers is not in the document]
simison: Translated 👍

nicksellen:
Maybe `role="status"` fits better? https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22.html

> Example 1: Including a search results message
> 
> After a user presses a Search button, the page content is updated to include the results of the search, which are displayed in a section below the Search button. The change to content also includes the message "5 results returned" near the top of this new content. This text is given an appropriate role for a status message. A screen reader will announce "Five results returned".
> 
> 
> 	<div role="status">5 results returned.</div>

Although [this guide](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/alert/) makes `alert` sound the right one:

> In our opinion, the only appropriate use for role="alert" is to announce a change that happened immediately as a result of a user action.
> 
> For example, when letting the user filter a list of elements using a text input. In this case, after the user typed a letter "T", the list is reduced to elements that contain a "T" in their name, and the screen reader would immediately alert "2 of 5 elements filtered".

modules/contacts/client/components/ContactList.component.js
```diff
@@ -19,18 +21,16 @@ export default class ContactList extends React.Component {
   }
 
   render() {
-
     const { appUser, contacts, onContactRemoved } = this.props;
+    const { t } = useTranslation('contact');
```
nicksellen: https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both

> You can’t use Hooks inside of a class component

simison: Ahh! Good catch — I'm still new to these.

simison: Addressed in https://github.com/Trustroots/trustroots/pull/1162/commits/9c8ba0193b74aea0e4b05be0b080b3d4fa75fdbf

simison:
Rebased! Ready for final look.

simison:
Rebased.

simison:
Nick approved in a video call ;-)

simison:
> I wonder why we have no links to this feature from other places. Is that on purpose?

It was tricky to get UI right so that different searches (hosts and users) wouldn't get confused. I think we could add a link to this page e.g. from the "contacts" page on profiles.

> AFAIK there was a broken database index so it didn't work for a long time, right? But now it does and i love it.

Yup! Someone built the indices for it.

> Has some privacy implications though. Is that why it's still hidden?

Wanna elaborate more? :-)

mrkvon:
:tada:

@simison 



> > Has some privacy implications though. Is that why it's still hidden?
>
> Wanna elaborate more? :-)

I see my comment was vague. By privacy implications I meant: Now we're able to find users by name, therefore moving from more privacy to less privacy.

I was just wondering what was the reason for the feature to be hidden. I don't have concerns.

PR #1163, opened 2020-01-04
Eslint: add comma-dangle always multi-line rule

simison:
#### Proposed Changes

* Switch `comma-dangle` Eslint rule from `never` to `always multi-line`
* Apply `eslint --fix` to all files

(I separated these steps into two commits)

This is kinda paving the way to eventually start using Prettier in the codebase (https://github.com/Trustroots/trustroots/pull/932), although it's possible to turn off [`trailingComma` option](https://prettier.io/docs/en/options.html#trailing-commas) in Prettier so this isn't strickly necessary. 

That said, it's a good convention to have since it's a bit more convenient to re-arrange different lists (variables, imports, etc) alphabetically without comma-hassle. 😁 

This will also cause one less line-change in diffs when modifying the end of multi-line lists.

#### Testing Instructions

* `npm run lint` passing?


mrkvon:
You're fast, guys :grin:

Anyways, let's get other PRs rebased! This affects almost everything.

I like this, too. :+1:

simison:
@mrkvon yup, and you can run `npm run lint -- --fix` to get commas in your PRs after rebasing.

PR #1148, opened 2019-12-12
Migrate Tribes page to React (feature branch)

mrkvon:
#### Proposed Changes

* This is a feature branch, which will be rebased and merged to master when the following PRs are merged here:
- [x] #1123 
- [x] #1134 
- [x] #1135 
- [x] #1137 
- [x] #1143 
- [x] #1144 
- [x] #1145 

Fixes #1133 

Follow up #1169

PR #1164, opened 2020-01-04
Tribes: Format the Tribe member count according to current language

mrkvon:
#### Proposed Changes

In #1134 we show the number of tribes in the format `1234567890 members`, but before the migration it was `1,234,567,890` in English.
So we fix this and add also support for different number formats based on current i18n language. Based on [`number.toLocaleString(languageCode)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString).
image
image

#### Testing Instructions
- git switch or checkout to a branch `tribes-react-fix-translation`
- run the app (e.g. `npm start`)
- go to http://localhost:3000/tribes
- see that the numbers are formatted as announced here
- switch the language and see that the number format changes (e.g. from 50,000 to 50 000).

Now, if the membership of the tribes is too low, you may need to select the `Tribe` react component with _React Developer Tools_ and change the tribe.count (sub)property of that component.

image

Continuous work on #1148. When it's approved, we'll squash it into #1134 because that's where it belongs.

### DIFF
--- a/config/client/i18n.js
+++ b/config/client/i18n.js
@@ -44,6 +44,10 @@ function format(value, format, languageCode) {
     return moment(value).format(format);
   }
 
+  if (typeof value === 'number' && format === 'number') {
+    return value.toLocaleString(languageCode);
+  }
+
   return value;
 }
 
--- a/modules/tribes/client/components/TribeItem.component.js
+++ b/modules/tribes/client/components/TribeItem.component.js
@@ -15,7 +15,7 @@ export default function TribeItem({ tribe, user, onMembershipUpdated }) {
 
   const countInfo = (tribe.count === 0)
     ? t('No members yet')
-    : t('{{count}} members', { count: tribe.count });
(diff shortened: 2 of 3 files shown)


config/client/i18n.js
```diff
@@ -44,6 +44,10 @@ function format(value, format, languageCode) {
     return moment(value).format(format);
   }
 
+  if (typeof value === 'number' && format === 'number') {
```
mrkvon: Can you think of a better name for a number format than `number`?

simison: I'm used to "number" in this context elsewhere. :-) 

https://codex.wordpress.org/Function_Reference/number_format_i18n
https://www.npmjs.com/package/@wordpress/i18n#_n

mrkvon:
Unfortunately, this commit edits on `Tribe.js` component which isn't in master yet.

PR #1167, opened 2020-01-07
Load all resources via imports

nicksellen:
#### Proposed Changes

(this is a subset of what https://github.com/Trustroots/trustroots/pull/1159 was)

This change switches to loading all client resources via explicit imports, using webpack loaders to process any content as required.

The old way was using glob paths in `config/assets/*` to specify which resources should be loaded in the client. These were then included via some webpack hacks in `webpack.shims.js`.

The new way is to use explicit import where things are needed. This make some things more verbose, but I think clearer what is using what.

The module files (e.g. `modules/contacts/client/contacts.client.module.js`) are now responsible for importing anything that is needed (i.e. all the controllers, directives, etc). I generated those files using [this script](https://gist.github.com/nicksellen/9a602e05624938a7ae1c11fa394714b5). As we won't likely be adding more angular code, I think this should be sufficient (don't need to include the generator code in the repo).

The templates are loaded via a custom webpack loader (the existing ones had some issues and are old), and ng-includes are also processed via webpack using code such as `ng-include="~modules/users/client/views/profile/activate.client.view.html"`.

All the angular ui bootstrap imports are consolidated in `config/client/bootstrap.js` - previously we could use wildcard imports for the templates, now it is explicit imports only.

This is preparation both for switching to jest and removing gulp and also removes some of the longstanding webpack shims which have always been destined for removal.

I also made a slightly unrelated change to use `.tests.js` suffix for the jest tests (to make it consistent).

#### Testing Instructions

* everything should just work as before


modules/users/client/config/users.client.routes.js
```diff
@@ -192,25 +220,28 @@
...
-        templateUrl: '/modules/users/views/profile/profile-view-accommodation.client.view.html',
+        templateUrl: profileviewAccommodationTemplateUrl,
         requiresAuth: true,
         noScrollingTop: true,
         data: {
           pageTitle: 'Profile accommodation',
         },
       }).
+      /*
+      @TODO find out what is wrong with this, template path did not seem to exist
```
nicksellen: The template `modules/offers/client/views/offer-host-view.client.view.html` does not exist and the URL does not show anything on production either (e.g. https://www.trustroots.org/profile/mikael/meet), but now the templates are explicitly imported we can't get away with invalid templates paths.

Any idea what that is about? @simison / @mrkvon?

modules/users/client/config/users.client.routes.js
```diff
@@ -364,23 +395,26 @@
...
+          templateUrl: profileReferencesTemplateUrl,
           requiresAuth: true,
           noScrollingTop: true,
           abstract: true,
           data: {
             pageTitle: 'Profile references',
           },
         }).
+        /*
+        @TODO find out what is up with this, template did not exist
```
nicksellen: I guess this missing template is because https://github.com/Trustroots/trustroots/pull/918 is still open?

modules/users/client/config/users.client.routes.js
mrkvon: `git log -- modules/offers/client/views/offer-host-view.client.view.html` shows this was once present, but got moved to a directive.

modules/users/client/config/users.client.routes.js
mrkvon: Yes, it was never present, and is unlikely to be in the future. The route is used, though.

I'd suggest to replace the `templateUrl: .....` with `template: ''`. Would that work for you here?

config/assets/default.js
```diff
@@ -1,74 +1,4 @@
...
-        'node_modules/angular-simple-logger/dist/angular-simple-logger.js', // Required by angular-leaflet-directive
-        'node_modules/ui-leaflet/dist/ui-leaflet.js',
-        'node_modules/leaflet-active-area/src/leaflet.activearea.js',
-        'node_modules/angular-waypoints/dist/angular-waypoints.all.js',
-        'node_modules/ng-file-upload/dist/ng-file-upload.js',
-        'node_modules/chosen-js/chosen.jquery.js',
-        'node_modules/angular-chosen-localytics/dist/angular-chosen.js',
-        'node_modules/angular-loading-bar/build/loading-bar.js',
-        'node_modules/angular-trustpass/dist/tr-trustpass.js',
-        'node_modules/mailcheck/src/mailcheck.js',
```
mrkvon: on signup, i got the error

```
TypeError
el.mailcheck is not a function
Call Stack
 link/<
  node_modules/angular-mailcheck/angular-mailcheck.js:59:16
 dispatch
  node_modules/jquery/dist/jquery.js:5237:27
 add/elemData.handle
  node_modules/jquery/dist/jquery.js:5044:28
```
could it be related to this change?

config/assets/default.js
```diff
@@ -1,74 +1,4 @@
...
-        'modal',
-        'popover',
-        'progressbar',
-        'tabs',
-        'tooltip',
-        'typeahead',
-      ],
-      js: [
-        // Non minified versions
-        'node_modules/jquery/dist/jquery.js',
```
mrkvon: How does this one get loaded?

config/assets/default.js
```diff
@@ -1,74 +1,4 @@
...
-        'node_modules/angular-ui-bootstrap/src/popover/popover.js',
-        'node_modules/angular-ui-bootstrap/src/position/position.js',
-        'node_modules/angular-ui-bootstrap/src/progressbar/progressbar.js',
-        'node_modules/angular-ui-bootstrap/src/stackedMap/stackedMap.js',
-        'node_modules/angular-ui-bootstrap/src/tabs/tabs.js',
-        'node_modules/angular-ui-bootstrap/src/tooltip/tooltip.js',
-        'node_modules/angular-ui-bootstrap/src/typeahead/typeahead.js',
-        'node_modules/moment/moment.js',
-        'node_modules/angular-moment/angular-moment.js',
-        'node_modules/medium-editor/dist/js/medium-editor.js',
```
mrkvon: How does this get loaded? (It works, anyways!)

config/webpack/entries/main.js
```diff
@@ -1,24 +1,32 @@
 /*
  *  Our main js entrypoint :)
  *
- *  It looks kind of empty because almost everything is happening via webpack.shims.js for now.
- *
  */
 
-import '@/public/dist/uib-templates';
```
mrkvon: Does this ever get imported anywhere now? Why do or don't we need it?

modules/search/client/search.client.module.js
```diff
@@ -1,3 +1,22 @@
-// Use application configuration module to register a new module
-// The core module is required for special route handling; see /core/client/config/core.client.routes
+import AppConfig from '@/modules/core/client/app/config';
+
+import '@/modules/core/client/core.client.module';
+
 AppConfig.registerModule('search', ['core']);
+
+// config
```
mrkvon: I like how these things become explicit...

config/assets/default.js
nicksellen: Ah yes, good point! I had meant to go through these one by one making sure I imported them elsewhere, but ended up only doing the angular modules specifically...

modules/users/client/config/users.client.routes.js
nicksellen: Ah, when the _meet_ type was added, it looks like there was originally going to be a page on the profile showing the users _meets_ and this `/profile/<username>/meet` route would have been the mobile page to show that, but even in the original incarnation the template url was just the same as for the _host_ type (`offer-host-view.client.view.html`), so wasn't implemented correctly. 

See https://github.com/Trustroots/trustroots/blob/afad74c0a5a3a7d91c9b851d671878f39dacf6d5/modules/users/client/config/users.client.routes.js#L206-L214

Conclusion: I should just remove this route.

modules/users/client/config/users.client.routes.js
nicksellen: Yup, that works, thanks!

config/assets/default.js
nicksellen: This is loaded in `config/webpack/webpack.shims.js` as a global thing. I should be able to switch that to an import.

config/assets/default.js
nicksellen: This is loaded in `config/webpack/webpack.shims.js` which makes it available globally as `jQuery`, `$`, and `window.jQuery`. I will leave this one like this, as it's very fundamental, lots of things expect it in those places.

config/webpack/entries/main.js
nicksellen: Those are now imported in `config/client/bootstrap.js`. All the ones ending in `.html` in there are angularjs templates, which get processed through the webpack loaders. Previously there was a gulp task that built the templates into `public/dist/uib-templates`.

config/assets/default.js
nicksellen: Ok, fixed. Just missing the `mailcheck` import above. This pre-dependency-aware libraries are painful!

modules/search/client/search.client.module.js
nicksellen: Yes, it's either ugly and verbose, or explicit and clear, depending on perspective.

modules/users/client/config/users.client.routes.js
simison: Sounds about right. :-)

modules/core/client/app/config.js
```diff
@@ -1,82 +1,103 @@
...
+];
+
+/**
+ * Load different service dependency for Angulartics depending on environment
+ * @link https://github.com/angulartics/angulartics
+ */
+if (appEnv === 'production') {
+  // @link https://github.com/angulartics/angulartics-google-analytics
+  appModuleVendorDependencies.push('angulartics.google.analytics');
+  require('angulartics-google-analytics/lib/angulartics-ga');
```
simison: This is fine like this and no changes needed; I just wanted to highlight Webpack dynamic imports which sometimes is a nifty feature for situations like this.

https://webpack.js.org/guides/code-splitting/#dynamic-imports

nicksellen: Yup, actually, related, these should be chosen a way that works with webpack define plugin and javascript minification that removes the other branches in production... I suspect right now all 3 are included in the production build (but still the correct one used). I will leave that till after merge...

I think using a dynamic import here would risk the library not having loaded by the time angular is bootstrapped as they are async and return a promise (so execution continues below, before the import is complete), whereas traditional `require` should still be synchronous. Just basing things on my stored knowledge, could be wrong.

simison: Yup, bundle size was my concern as well and valid points about Angular. It's all code going away eventually, anyway, and not a ton of KBs.

I think I've seen some Webpack plugin that's smart about these env checks and can do the codesplitting. 🤔  I don't think it's necessarily worth setting up for this, tho.

nicksellen: Splitting is built into webpack now, but currently we have no way to tell the server to serve files other than main.js/css, but I have a plan for that... I wrote about it somewhere else already.

The env checks work because the DefinePlugin literally rewrites the code so that (depending on how you've configured it), `process.env.NODE_ENV === 'test'` gets turned into `'production' === 'test' (for example), and a normal javascript minifier can do dead code elimination. I _think_ the way we have used the DefinePlugin will make this not work (`JSON.stringify`ing the whole of process.env), it's on my list to investigate what actually happens...

PR #1168, opened 2020-01-07
Switch all client tests to jest

nicksellen:
#### Proposed Changes

Now we have jest for frontend tests, we can remove karma. This makes the minimal changes needed to get the existing client tests to run under jest.

Built on top of #1167 so that should be merged first.

#### Testing Instructions

* run the tests


jest.config.js
```diff
@@ -1,9 +1,24 @@
 module.exports = {
   moduleNameMapper: {
-    '@/(.*)$': '<rootDir>/$1',
+    '^@/(.*)$': '<rootDir>/$1',
```
mrkvon: :astonished: Whatever the stuff in this config is doing, it works!

PR #1169, opened 2020-01-08
React tests for Tribes

mrkvon:
#### Proposed Changes

- testing tribes (modules/tribes/tests/client/components/TribesPage.component.test.js)

#### Testing Instructions

- `git switch tribes-react-tests`
- `npm run test:client`
- see that the tests pass
- suggest more things to test

Work on #1133

### DIFF
--- a/modules/tribes/tests/client/components/TribesPage.component.test.js
+++ b/modules/tribes/tests/client/components/TribesPage.component.test.js
@@ -1,0 +1,177 @@
+/* tests to write
+
+when clicking join and not logged in, redirect to tribe page
+when clicking join and logged in, send a request to join to API
+when clicking join and logged in, update the number of tribe members
+when clicking leave, send a request to leave to API
+when clicking leave and logged in, update the number of tribe members
+
+
+*/
+
+import React from 'react';
+import * as rtl from '@testing-library/react';
+import { render, fireEvent, wait } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import '@/config/client/i18n';
+
+import TribesPage from '@/modules/tribes/client/components/TribesPage.component';
+import * as api from '@/modules/tribes/client/api/tribes.api';
+
+jest.mock('@/modules/tribes/client/api/tribes.api');
+
+const dummyTribes = [{
+  _id: 'aaaa',
+  id: 'aaaa',
+  count: 20000,
+  slug: 'hitchhikers',
+  label: 'Hitchhikers',
+}, {
+  _id: 'bbbb',
+  id: 'bbbb',
+  count: 5000,
+  slug: 'snails',
+  label: 'Snails',
+}, {
+  _id: 'cccc',
+  id: 'cccc',
+  count: 1500,
+  slug: 'teddybears',
+  label: 'Teddy Bears',
+}];
+
+const dummyUser = {
+  _id: 'user0',
+  id: 'user0',
+  memberIds: ['cccc', 'aaaa'],
+};
+
+api.read.mockImplementation(async () => dummyTribes);
+api.join.mockImplementation(async id => ({ tribe: dummyTribes.find(tribe => tribe._id === id) }));
+api.leave.mockImplementation(async id => ({ tribe: dummyTribes.find(tribe => tribe._id === id) }));
+
+
+describe('TribesPage', () => {
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('not signed in', () => {
+
+    it('fetch tribes from api and show tribes on page', async () => {
+      // first no api should be called
+      expect(api.read).toHaveBeenCalledTimes(0);
+
+      const { getAllByRole } = render(<TribesPage onMembershipUpdated={() => {}} />);
+
+      await wait();
+
+      // the tribes from api should be displayed
+      const tribes = getAllByRole('listitem');
+      expect(tribes).toHaveLength(dummyTribes.length + 1); // there is additionally the Suggest Tribe
+
+      tribes.slice(0, -1).forEach((tribe, i) => {
+        expect(tribe).toHaveTextContent(dummyTribes[i].label);
+      });
+
+      // during the test the api should be called only once
+      expect(api.read).toHaveBeenCalledTimes(1);
+    });
+
+    it('the join button should be a link to tribe page', async () => {
+      const { getAllByRole } = render(<TribesPage onMembershipUpdated={() => {}} />);
+
+      await wait();
+
+      // join button
+      const tribes = getAllByRole('listitem').slice(0, -1);
+
+      tribes.forEach((tribe, i) => {
+        const button = rtl.getByText(tribe, 'Join');
+        expect(button).toHaveAttribute('href', `/signup?tribe=${dummyTribes[i].slug}`);
+      });
+    });
+
+  });
+
+  describe('signed in', () => {
+
+    it('show tribes on page', async () => {
+      // first no api should be called
+      expect(api.read).toHaveBeenCalledTimes(0);
+
+      const { getAllByRole } = render(<TribesPage user={dummyUser} onMembershipUpdated={() => {}} />);
+
+      await wait();
+
+      // the tribes from api should be displayed
+      const tribes = getAllByRole('listitem');
+      expect(tribes).toHaveLength(dummyTribes.length + 1); // there is additionally the Suggest Tribe
+
+      tribes.slice(0, -1).forEach((tribe, i) => {
+        expect(tribe).toHaveTextContent(dummyTribes[i].label);
+      });
+
+      // during the test the api should be called only once
+      expect(api.read).toHaveBeenCalledTimes(1);
+    });
+
+    it('user is member of some tribes and not member of others', async () => {
+      const { getAllByText } = render(<TribesPage user={dummyUser} onMembershipUpdated={() => {}} />);
+
+      await wait();
+
+      // the tribes from api should be displayed
+      const join = getAllByText('Join');
+      const leave = getAllByText('Joined');
+      expect(join).toHaveLength(1);
+      expect(leave).toHaveLength(2);
+    });
+
+    it('click join and send api request', async () => {
+      expect(api.join).toHaveBeenCalledTimes(0);
+      const { getAllByText } = render(<TribesPage user={dummyUser} onMembershipUpdated={() => {}} />);
+
+      await wait();
+
+      // click the join button
+      const join = getAllByText('Join');
+      expect(join).toHaveLength(1);
+      fireEvent.click(join[0]);
+
+      await wait();
+
+      // api should be called
+      expect(api.join).toHaveBeenCalledTimes(1);
+    });
+
+    it('click leave with modal and send api request', async () => {
+      expect(api.leave).toHaveBeenCalledTimes(0);
+
+      const { getAllByText } = render(<TribesPage user={dummyUser} onMembershipUpdated={() => {}} />);
+
+      await wait();
+
+      // click the first Joined button
+      const buttons = getAllByText('Joined');
+      expect(buttons).toHaveLength(2);
+      fireEvent.click(buttons[0]);
+
+      await wait();
+
+      // confirmation modal should open
+      // we'll click Confirm
+      const confirm = getAllByText('Leave Tribe');
+      expect(confirm).toHaveLength(1);
+      fireEvent.click(confirm[0]);
+
+      await wait();
+
+      // and the api was called as expected
+      expect(api.leave).toHaveBeenCalledTimes(1);
+    });
+
+  });
+
+});
--- a/modules/tribes/tests/client/components/TribesPage.component.tests.js
+++ b/modules/tribes/tests/client/components/TribesPage.component.tests.js
@@ -13,25 +13,29 @@ jest.mock('@/modules/tribes/client/api/tribes.api');
 
 const onMembershipUpdated = jest.fn();
 
-const dummyTribes = [{
-  _id: 'aaaa',
-  id: 'aaaa',
-  count: 20000,
-  slug: 'hitchhikers',
-  label: 'Hitchhikers',
-}, {
-  _id: 'bbbb',
-  id: 'bbbb',
-  count: 5000,
-  slug: 'snails',
-  label: 'Snails',
-}, {
-  _id: 'cccc',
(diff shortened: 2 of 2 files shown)


modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+when clicking join and logged in, send a request to join to API
+when clicking join and logged in, update the number of tribe members
+when clicking leave, send a request to leave to API
+when clicking leave and logged in, update the number of tribe members
+
+
+*/
+
+import React from 'react';
+import * as rtl from '@testing-library/react';
```
nicksellen: I guess you imported it like that to distinguish the variant that takes the container as the first argument?

I quite like [`within`](https://testing-library.com/docs/dom-testing-library/api-helpers#within-and-getqueriesforelement-apis) as another approach:

```js
import { within } from '@testing-library/react';

// ... later

const button = within(tribe).getByText('Join');
// ... or ...
const { getByText } = within(tribe)
const button = getByText('Join');
```

Reads a bit clearer to me, when I first saw `rtl` in the code I was a bit confused, and had to check the imports to find out (it has the same initialization as "right to left").

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+
+const dummyUser = {
+  _id: 'user0',
+  id: 'user0',
+  memberIds: ['cccc', 'aaaa'],
+};
+
+api.read.mockImplementation(async () => dummyTribes);
+api.join.mockImplementation(async id => ({ tribe: dummyTribes.find(tribe => tribe._id === id) }));
+api.leave.mockImplementation(async id => ({ tribe: dummyTribes.find(tribe => tribe._id === id) }));
```
nicksellen: I have tended to only mock implementations at the exact scope they are required to try and make it clear what will expect it. So inside the test method if it is unique, or in a `beforeEach` in the closest `describe` block.

I am not set on that approach though, so more a comment for consideration.

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+      // the tribes from api should be displayed
+      const tribes = getAllByRole('listitem');
+      expect(tribes).toHaveLength(dummyTribes.length + 1); // there is additionally the Suggest Tribe
+
+      tribes.slice(0, -1).forEach((tribe, i) => {
+        expect(tribe).toHaveTextContent(dummyTribes[i].label);
+      });
+
+      // during the test the api should be called only once
+      expect(api.read).toHaveBeenCalledTimes(1);
```
nicksellen: I have been tending to check the parameters that are sent to the API call too, is there a reason you do not? I can imagine some bugs where the wrong thing is sent.

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+    });
+
+    it('the join button should be a link to tribe page', async () => {
+      const { getAllByRole } = render(<TribesPage onMembershipUpdated={() => {}} />);
+
+      await wait();
```
nicksellen: I tried to avoid "randomly" waiting, but tried to find elements I could wait for using `waitForElement`.

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+
+      // join button
+      const tribes = getAllByRole('listitem').slice(0, -1);
```
nicksellen: I wonder if it's better to get everything by the button role instead of the list items? Not sure.

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+      // the tribes from api should be displayed
+      const join = getAllByText('Join');
+      const leave = getAllByText('Joined');
+      expect(join).toHaveLength(1);
+      expect(leave).toHaveLength(2);
+    });
+
+    it('click join and send api request', async () => {
+      expect(api.join).toHaveBeenCalledTimes(0);
+      const { getAllByText } = render(<TribesPage user={dummyUser} onMembershipUpdated={() => {}} />);
```
nicksellen: You could create a mock function to check if onMembershipUpdated got called using `jest.fn()`, something like:

```js
const onMembershipUpdated = jest.fn();
const { getAllByText } = render(<TribesPage user={dummyUser} onMembershipUpdated={onMembershipUpdated} />);
// ...
expect(onMembershipUpdated).toHaveBeenCalled...
```

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+
+      await wait();
+
+      // click the join button
+      const join = getAllByText('Join');
+      expect(join).toHaveLength(1);
+      fireEvent.click(join[0]);
```
nicksellen: I _think_ if you use the non-_all_ version (`getByText`) it will automatically thrown an exception if there are more than 1, so can simplify this to:

```js
fireEvent.click(getByText('Join'));
```

... _or_, if combining with `waitForElement` something like:

```js
const join = await waitForElement(() => getByText('Join'));

fireEvent.click(join);
```

... which to me gets closer to what the user is doing ... (I can imagine the `await wait()` could be flakey, for example if another API call had to be added after the first one, it might not wait long enough).

Maybe [`findBy`](https://testing-library.com/docs/dom-testing-library/api-queries#findby) is even better though:

```js
const join = await findByText('Join');

fireEvent.click(join);

// .. or even inline

fireEvent.click(await findByText('Join'));
```

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+*/
+
+import React from 'react';
+import * as rtl from '@testing-library/react';
+import { render, fireEvent, wait } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import '@/config/client/i18n';
+
+import TribesPage from '@/modules/tribes/client/components/TribesPage.component';
+import * as api from '@/modules/tribes/client/api/tribes.api';
```
nicksellen: I was creating mock `api` objects as they are used in the components, e.g.:

```js
import * as tribes from '@/modules/tribes/client/api/tribes.api';
const api = { tribes };
```

Which I liked because:
* maybe some tests need to use two different APIs in some places
* consistent with how it looks in the code where it is used for real
* made it very clear when reading the code later which API I was using

Doesn't seem essential though, kind of stylistic. What think you?

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+  count: 5000,
+  slug: 'snails',
+  label: 'Snails',
+}, {
+  _id: 'cccc',
+  id: 'cccc',
+  count: 1500,
+  slug: 'teddybears',
+  label: 'Teddy Bears',
+}];
```
nicksellen: I think hard coded values are fine for now, but I kind of like a faker library that generates some random nuimbers/text, just in case the tests are accidentally hard coding some expectations based on static data.

modules/tribes/tests/client/components/TribesPage.component.test.js
```diff
@@ -0,0 +1,177 @@
...
+    it('user is member of some tribes and not member of others', async () => {
+      const { getAllByText } = render(<TribesPage user={dummyUser} onMembershipUpdated={() => {}} />);
+
+      await wait();
+
+      // the tribes from api should be displayed
+      const join = getAllByText('Join');
+      const leave = getAllByText('Joined');
+      expect(join).toHaveLength(1);
+      expect(leave).toHaveLength(2);
```
nicksellen: I would tend to put the queries inline, and then you'll also see the linter suggest using the `query*` version, e.g.:

```js
expect(queryAllByText('Join')).toHaveLength(1);
expect(queryAllByText('Joined')).toHaveLength(2);
```

See https://testing-library.com/docs/dom-testing-library/api-queries for docs on get vs query.

mrkvon:
This should to wait for #1167. Will need a little renaming. Can be even merged separately from tribes-react.

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: I like that, although I haven't used it before. Do you know a library you would recommend @nicksellen? Maybe the `faker` library we already have in our `devDependencies` could be used for this purpose?

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: I love the `within`. Replaced in 48216ee1f3691affb82eca49b6aaca06ed1271be.

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: Just bein' lazy :grin: 
Added the `toHaveBeenCalledWith` check in 48216ee1f3691affb82eca49b6aaca06ed1271be and e52079d55e2f93259c5a8115df108969cc2f1069.

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: replaced all `wait`s with `waitForElement` in e52079d55e2f93259c5a8115df108969cc2f1069

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: When the user is signed out, the buttons are just links, so `getAllByRole(button)` won't find anything. However we can do `getAllByText('Join', { selector: 'a' })` instead. The selector part is optional.

mrkvon: 98c404329bf72eef2a6276eb12484bcc098660f3

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: done in e52079d55e2f93259c5a8115df108969cc2f1069

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: Rewritten the test fundamentally. So I'll use what i've learned here later. Thank you!

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: It sounds fair to have this consistent. Changed in 48216ee1f3691affb82eca49b6aaca06ed1271be.

modules/tribes/tests/client/components/TribesPage.component.tests.js
```diff
@@ -0,0 +1,202 @@
+import React from 'react';
+import { within } from '@testing-library/react';
+import { render, fireEvent, waitForElement } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import '@/config/client/i18n';
+
+import TribesPage from '@/modules/tribes/client/components/TribesPage.component';
+import * as tribesApi from '@/modules/tribes/client/api/tribes.api';
```
mrkvon: FYI we called this `tribesApi` instead of `tribes` because the `eslint-plugin-import` was complaining when we used the `tribes` name later on.

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: Fixed in 48216ee1f3691affb82eca49b6aaca06ed1271be.

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: I'm trying to do this with the `faker` library, but I suggest we do it in different PR. It's a lot of added logic. WDYT?

modules/tribes/tests/client/components/TribesPage.component.test.js
mrkvon: I'll leave this as it is for now. It'll make a lot of sense to move to `beforeEach` when we create the fake data `beforeEach` test. These `mockImplementation`s seem to be bound to those dummyTribes. Does it make sense?

modules/tribes/tests/client/components/TribesPage.component.tests.js
```diff
@@ -0,0 +1,202 @@
...
+
+// helper function to determine a membership
+const isMember = (user, tribe) => user.memberIds.includes(tribe._id);
+
+// TextMatch function for finding an element with arbitrarily nested text
+// https://testing-library.com/docs/dom-testing-library/api-queries#textmatch
+const nestedTextMatch = text => (content, element) => element.textContent.includes(text);
+
+// helper that renders the TribesPage, waits for the tribes to be loaded from API and returns the rendered page
+const renderAndWaitForTribes = async ({ user }) => {
```
mrkvon: What do you think about extracting the render + tribe-waiting to `beforeEach` like this? Does it make the code drier, or harder to read?

mrkvon:
@nicksellen Thank you very much for the detailed review! I learned something from each comment. :heart: 

Based on the review i refactored the tests considerably. Another round of reviews would be beneficial.

modules/tribes/tests/client/components/TribesPage.component.test.js
simison: I think it's fine if you prefer to add some `data-identifier` to those elements (or `id` if you want some specific one) for the tests. Up to you!

modules/tribes/tests/client/components/TribesPage.component.test.js
simison: I on the other hand like when the test data is always the same & consistent on each run but has variety within tests to cover odd cases. 😁  That way test runs don't have random flakiness. Not a strong opinion tho!

modules/tribes/tests/client/components/TribesPage.component.tests.js
simison: Definitely DRYer. 👍

PR #1171, opened 2020-01-09
Migrate messages inbox to react

nicksellen:
#### Proposed Changes

* this switches the inbox view  (`/messages`) to be a react component

### TODO

- [x] i18n everything in sight
- [x] add back `syncUnreadMessagesCount` broadcast
- [x] add back analytics call
- [x] remove legacy controller

#### Testing Instructions

* visit inbox, perhaps have a lot of threads so you see the "More messages" button


modules/messages/tests/client/components/Inbox.component.tests.js
```diff
@@ -0,0 +1,84 @@
...
+    },
+  },
+];
+
+describe('<Inbox>', () => {
+
+  it('shows a nice message if there are no conversations', async () => {
+    api.fetchThreads.mockResolvedValue({ threads: [] });
+    const { findByText } = render(<Inbox user={me}/>);
+    expect(await findByText('No conversations yet.')).toBeInTheDocument();
```
nicksellen: This should probably be a `role="alert"` element.

modules/messages/tests/client/components/Inbox.component.tests.js
```diff
@@ -0,0 +1,84 @@
...
+      username: 'admin1',
+      displayName: 'Newell Jones',
+    },
+    userTo: {
+      _id: '5de7ec34f333ce7b23cdfc61',
+      username: '571eloyhintz',
+      displayName: 'Eloy Hintz',
+    },
+  },
+];
```
nicksellen: Would like to create some reusable generators for this kind of test data.

modules/messages/client/components/Inbox.component.js
```diff
@@ -0,0 +1,122 @@
...
+
+import * as api from '../api/messages.api';
+import Avatar from '@/modules/users/client/components/Avatar.component';
+import Activate from '@/modules/users/client/components/Activate';
+
+const userType = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  username: PropTypes.string.isRequired,
+  displayName: PropTypes.string.isRequired,
+});
```
nicksellen: We could start to put these propTypes into places where they can be reused across the app... maybe in 'modules/users/client/components/users.prop-types.js`?

modules/messages/tests/client/components/Inbox.component.tests.js
simison: There are a few at https://github.com/Trustroots/trustroots/blob/1bb2908072257cd737d66cbe13539d6ab1eab513/testutils/server/data.server.testutil.js

modules/messages/tests/client/components/Inbox.component.tests.js
simison: I have `NoContent` component in https://github.com/Trustroots/trustroots/pull/1162 that could be handy here.

nicksellen: Yup, lets try and get that merged soon! It doesn't have `role="alert"` though, might make sense to?

simison: Yep — feel free to add (and merge!)

modules/core/client/services/angular-compat.js
```diff
@@ -0,0 +1,13 @@
...
+  return angular.element(document).injector().get(name);
+}
+
+export function $broadcast(...args) {
+  return get('$rootScope').$broadcast(...args);
+}
+
+export function eventTrack(...args) {
+  return get('$analytics').eventTrack(...args);
+}
```
nicksellen: If there started to be a lot of stuff in here it could be refactored to have one file per "angular-thing" then used something like:

```js
import { eventTrack } from '@/modules/core/client/angular-compat/$analytics'
```
```

File: modules/messages/tests/client/components/Inbox.component.tests.js (hunk 1 above)
Comment (nicksellen) [reply modules/messages/tests/client/components/Inbox.component.tests.js]: #1162 has a quite a few comments/discussions on it still, so I'll leave it for now.

File: .travis.yml [hunk 5]
```diff
@@ -19,19 +19,12 @@ dist: trusty
...
     - graphicsmagick
 before_script:
   - mongo --version
   - mongod --version
 
-  # https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI
-  - 'export PATH=$PATH:/usr/lib/chromium-browser/'
-  - 'export DISPLAY=:99.0'
-  - 'sh -e /etc/init.d/xvfb start'
   - sleep 3 # give xvfb some time to start
```
simison: `sleep 3` can be removed now that you removed  `xvfb`

modules/core/client/components/TimeAgo.js
```diff
@@ -0,0 +1,40 @@
...
+const REFRESH_INTERVAL = 10000; // 10s
+
+/**
+ *  An automatically refreshing time ago component.
+ *
+ *  It refreshes every 10 seconds. If this seems a bit inefficient at some point
+ *  e.g. there are loads of components pointlessly refreshing, there are two things
+ *  we could do:
+ *    1. share the setInterval across all instances
+ *    2. switch to slower updates for further ago dates
```
simison: It could also be that `socket.io` type solution works better in some cases instead of polling. But I like how easy polling is to reason about.

modules/messages/client/api/messages.api.js
```diff
@@ -0,0 +1,20 @@
+import axios from 'axios';
+import parseLinkheader from 'parse-link-header';
+
+export async function fetchThreads(params = {}) {
+  const { data: threads, headers } = await axios.get('/api/messages', { params: { limit: 2, ...params } });
+  let nextParams;
+  if (headers.link) {
```
simison: This might eventually be good thing to abstract as a "middleware" type function we can re-use for other paginated endpoints. Good as it is now of course.

modules/core/client/components/TimeAgo.js
nicksellen: This is about updating "time from now" display things like from "a few seconds ago" so that a minute later it'll update to say "about a minute ago" (or whatever the moment time ago formatter says).

modules/messages/client/components/Inbox.component.js
simison: Valid! Could also be ok follow up PR.

modules/messages/client/components/Inbox.component.js
```diff
@@ -0,0 +1,134 @@
...
+      <section className="container-spacer">
+        <Activate/>
+      </section>
+    );
+  }
+  const { t } = useTranslation('messages');
+
+  const [nextParams, setNextParams] = useState(null);
+  const [isFetching, setIsFetching] = useState(false);
+  const [threads, setThreads] = useState([]);
```
simison: Not opposed to this, but is the idea we never use `this.state` anymore now that we have hooks? 😁 

I'm still kinda uneducated about React Hooks so seriously curious! Not asking to change anything here either. :-)

modules/messages/client/components/Inbox.component.js
```diff
@@ -0,0 +1,134 @@
...
+
+  const hasMore = Boolean(nextParams);
+
+  async function fetchThreads(next = false) {
+    setIsFetching(true);
+    try {
+      if (next) {
+        eventTrack('inbox-pagination', {
+          category: 'messages.inbox',
+          label: 'Inbox page ' + nextParams,
```
simison: Was `nextParams` Object or Integer?

modules/messages/client/components/Inbox.component.js
```diff
@@ -0,0 +1,134 @@
...
+    $broadcast('syncUnreadMessagesCount');
+    fetchThreads();
+  }, []);
+
+  return <section className="container-spacer">
+    {!isFetching && threads.length === 0 && (
+      <div className="content-empty">
+        <i className="icon-3x icon-messages-alt"/>
+        <h4 role="alert">{t('No conversations yet.')}</h4>
+      </div>
```
simison: We can replace this with a component from https://github.com/Trustroots/trustroots/pull/1162 later on.

modules/messages/client/components/Inbox.component.js
nicksellen: Yup, no `this.state` anymore, :fire: it all (well, perhaps there are things that only work with class components, and `this.state`, but I didn't come across anything).

Actually, later on the idea would be to use (or at least consider using) redux for some of the state management.

When you say _not opposed to this_, does that mean you have some caution/hesitation?

modules/messages/client/components/Inbox.component.js
```diff
@@ -0,0 +1,134 @@
...
+    )}
+  </section>;
+}
+
+Inbox.propTypes = {
+  user: userType.isRequired,
+};
+
+function InboxThread({ user, thread }) {
```
simison: What do you think about splitting this component into a separate file for clarity? I like one-component-per-file approach quite a bit since it encourages re-using.

Not a blocker tho.

simison:
Good to test what happens when user has been removed after messages are written. I think old Angular version didn't cope well with that so it's not a biggie if this doesn't either. 😁

modules/messages/client/components/Inbox.component.js
nicksellen: It's an object (with page and per page values iirc), I attempted to just replicate it as it was used before (to at least make the `label` consistent with existing values), although I didn't think about it too much.

How is that label used? Perhaps just showing the page number (and ignoring the per page value) would make more sense...

simison: The label just allows seeing how far in history people typically browse but I haven't been interested in that for long time so ok to drop the label.

modules/messages/client/components/Inbox.component.js
nicksellen: One of the things I like about using react is a component is as lightweight as a function, so easy to split things off into microcomponents, which might not make sense to have in seperate files if they are tiny, in this case maybe it justifies a seperate component. I'll split it out.

simison: Definitely case by case. 👍

modules/messages/client/components/Inbox.component.js
simison: Just haven't seen the Hooks used so much yet for everything stateful in a component (rather one off things), but I haven't seen hooks in action much yet at all. 

It's all good, thanks for elaborating!

modules/core/client/components/TimeAgo.js
simison: Ahhh, gotcha! 👍

nicksellen:
The other hanging question is what to do with the translations strings as we don't have a standard auto-extraction method yet, I think I could:
* manually/semi-automatically extract them, using existing format/structure
* ignore them for now, and revisit later (after merge) once we've got a clearer idea of the approach

simison:
I'd ignore them now — since we're not using them in production it's important that we add `t()` functions everywhere so that we're ready when we will have extraction and the rest of process in place.

modules/messages/client/components/Inbox.component.js
nicksellen: I separated it out now :)

modules/messages/client/components/Inbox.component.js
nicksellen: Yup! I'll wait until then... where we can revisit the `role` thing... I'll keep it as _alert_ for now.

modules/messages/client/components/Inbox.component.js
nicksellen: I changed it to just have the page number. Perhaps we can rip out lots of those analytics calls at some point, but I'll keep it in for now, seeing as it was already there.

modules/messages/client/api/messages.api.js
nicksellen: Yes, I agree! When the time comes.

modules/messages/client/components/Inbox.component.js
nicksellen: I did it already, as I needed it to be in a seperate file when I extracted the `InboxThread` component. I put that one in `/modules/users/client/users.prop-types` (as it's not a component... ). Will do for now at least.

nicksellen:
> Good to test what happens when user has been removed after messages are written. I think old Angular version didn't cope well with that so it's not a biggie if this doesn't either.

Yup, I'll leave that out for now, and I guess it's captured in https://github.com/Trustroots/trustroots/issues/596 ?

simison:
It might be that the small "are messages in spirit of trustroots" box has gone missing with this PR. Haven't tested yet before/after but wanted to log in case.

https://github.com/Trustroots/trustroots/blob/ba34ebf43f48fa6070cb4cffe983421edbeb6b80/modules/messages/client/views/thread.client.view.html#L200

I would expect it here (in red):

image


Props @noahsmindfuck for noticing.

nicksellen:
Only the inbox page is migrated so far (the list of threads) the thread page (the actual conversation) is still in angular. My current PR is switching that over, but it's not ready yet.

simison:
Ahh yeah, sleepy me. It must be gone earlier, then!

PR #1173, opened 2020-01-16
Implement messages threads in react

nicksellen:
#### Proposed Changes

* component for handling `/messages/<username>` in react

### TODO

- [x] make a better TODO list
- [x] find a solution for handling what was done in `thread-dimensions.client.directive.js` before
  - [x] layout that adjusts with reply box message height
  - [x] the scrolling/reloading kind of infinite scroll stuff
- [x] mark messages as read
- [x] make send message work
- [ ] call `eventTrack`
- [x] save draft message in cache
- [x] "fetch more" messages upward scroll thing
- [ ] add special case things
  - [x] non-public account
  - [x] when user profile is empty
  - [x] when no messages
  - [x] user is no longer a member
  - [ ] error and retry?
  - [x] quick reply
- [x] i18n everything
- [ ] move some of my inline style somewhere else
- [ ] consider moving some more less/css into the components
- [x] move general components into better places
  - [x] `TrEditor`
  - [x] `Monkeybox`
- [ ] find out why `Monkeybox` is called `Monkeybox`
- [x] remove old angular versions of the things
- [x] add some tests
- [x] documentation
  - [x] `InfiniteMessages` component

#### Testing Instructions

*

Fixes #


modules/messages/client/api/messages.api.js
```diff
@@ -15,8 +15,49 @@ export async function fetchThreads(params = {}) {
...
+  const { data: message } = await axios.post('/api/messages', {
+    userTo: userToId,
+    content,
+    read: false,
+  });
+  return message;
+}
+
+export async function markRead(messageIds) {
+  // the server response is a bit wrong, status 200, but no body or content-type, so we
```
simison, line 56: Can we change the response from server? :-)

modules/messages/client/components/QuickReply.js
```diff
@@ -0,0 +1,72 @@
...
+  const { t } = useTranslation('messages');
+
+  // It's not quite clear to me how to handle the i18n of
+  // these messages. They are a bit of a hack to start with.
+  // They rely on the "data-hosting" attribute being set on the
+  // p element, but it still has message content that will not
+  // be translated later, it's stuck at it's original content.
+  // I chose to write the message content in the language the
+  // user has set.
+  // Probably better would be not to include any message content,
```
simison, line 15: I'd be okay removing this feature for now too; we could analyze data a bit.

Initial motivation was to improve reply rates by making it easy to quickly reply with a click of a button (including from emails).

modules/messages/client/components/ThreadReply.js
```diff
@@ -0,0 +1,96 @@
...
+export default function ThreadReply({ onSend, cacheKey }) {
+  const { t } = useTranslation('messages');
+
+  const [editorKeyCounter, setEditorKeyCounter] = useState(0);
+  const [content, setContent] = useState(() => getDraft() || '');
+
+  function send(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    const value = content.replace(/<br><\/p>$/, '</p>'); // remove the last <br>
```
simison, line 17: We could switch this to server-side too but ok here as well.

modules/messages/client/components/QuickReply.js
nicksellen: I think it's OK to keep it in place, it's not _more_ hacky now than it was, and could be built on...  i.e. implementing "special messages" of the same type, and can still rewrite the old ones... but for another PR.

modules/messages/client/api/messages.api.js
nicksellen: Good point! Done.

modules/messages/client/components/ThreadReply.js
nicksellen: I did the opposite :) - moved it inside TrEditor (I think it should be a feature of medium editor, so it emits better html in the first place.

simison: Hehe, I'm not opposed to that idea either. I guess both are equally funky and not fixing the root cause anyway. :-)

modules/messages/client/components/QuickReply.js
simison: Sounds good! 👍

modules/users/client/components/Monkeybox.js
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <div className="monkeybox panel panel-default">
+      <div className="panel-body">
+        <Avatar user={user} size={64} />
+        <h3>
+          <a>{user.displayName}</a>
+        </h3>
+        <TribesInCommon user={user} otherUser={otherUser} />
+        {user.languages.length > 0 && (
+          <div className="monkeybox-section">
```
simison, line 60: For some reason when in my test `/api/users/[USERNAME]` response has `languages: []`, I'm still getting `Warning: Each child in a list should have a unique "key" prop.
` warning for this line and I don't understand why. 🤔  This section should never render...

simison: The section is actually rendering wrong profile — it shows up my own profile, while it should show the other user's profile:

image

modules/messages/client/components/Thread.component.js
```diff
@@ -0,0 +1,276 @@
...
+        setDoesNotExist(true);
+      } else {
+        throw error;
+      }
+    } finally {
+      setIsFetching(false);
+    }
+  }
+
+  async function sendMessage(content) {
```
simison, line 183: Not that it's like this in `master` either, but would be great if focus would return to text input after sending the message. :-)

simison:
quote format seems to be very sticky and hard to jump out from:
image

Unlike in `master`, where it's little too easy to exit quote. :-D Not sure if that's worth fixing now or letting it be for now.

modules/core/client/components/TrEditor.js
```diff
@@ -0,0 +1,202 @@
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import MediumEditor from 'react-medium-editor';
```
simison, line 3: Heh, updated last time 4 years ago.... https://www.npmjs.com/package/react-medium-editor

We might need to look for an alternative in a couple years. ;-)

nicksellen: Indeed, and it also uses bits of react that are deprecated and will be removed in the future.

Fortunately, it's only a relatively small shim between react and medium-editor, so not too much to reimplement ourselves if needed.

Also has had a few commits more recently https://github.com/wangzuo/react-medium-editor/commits/master (just no release).

They actually recommend to use https://draftjs.org/ instead, which could be worth investigating.

simison: Yepp, we can cross that bridge once the time comes.

modules/users/client/components/Monkeybox.js
nicksellen: Good catch for the wrong user! Fixed in https://github.com/Trustroots/trustroots/pull/1173/commits/70ee62cf94d3b378be9dca831185718c44437dad

I can't recreate the unique "key" issue... and it doesn't make sense as you pointed out. I'm wondering if something else caused that issue. Maybe a user can have multiple non-unique langauges in the seed data? I'll leave it out for now...

modules/messages/client/components/Thread.component.js
nicksellen: Implemented it in https://github.com/Trustroots/trustroots/pull/1173/commits/1d43720c0cb78b4556673405d573d1546027fd12

nicksellen:
> The box highlights the name as a link (with underline and everything) but it's not clickable:

Fixed in https://github.com/Trustroots/trustroots/pull/1173/commits/ac75e9f87ad6bdad8bace15ee1d2cfb44b0d5819

> When I have too short profile description and I've already written to someone, I don't see the inbox but I also don't see any explanation what's going on:

Fixed in https://github.com/Trustroots/trustroots/pull/1173/commits/4f92bd669df2e8cdd146c3765cd3b00c37db088c (in that it still allows you to write to people if you have an existing conversation, which I what I think the existing behaviour is, kind of from memory of looking at the code... but I think I only considered it for the quick reply case or something).

> quote format seems to be very sticky and hard to jump out from:

From my quick testing, it behaves the same as on https://dev2.trustroots.org/

(pressing backspaces gets you out, I think press enter twice is what I would expect to break me out)

> The above is an issue also when writing to someone with suspended user role.

I didn't consider this case, not sure what it _should_ do. I would leave it out for now... maybe I can leave it out to get it merged, and we fix after.

PR #1176, opened 2020-01-18
Fix/accidental assignment

nicksellen:
#### Proposed Changes

* there was a bug where I used assignment operator instead of equality
* I fixed the error (`=` -> `===`)
* I added a test that would fail with the same error we had in production
* I added an eslint rule ([no-multi-assign](https://eslint.org/docs/rules/no-multi-assign)) that prevents that kind of thing

#### Testing Instructions

* you should be able to run the tests and view messages :)


simison:
Thanks for such a thorough fix!

PR #1178, opened 2020-01-18
Shadowban

simison:
- Adds extra protection by filtering out suspended users.
- Adds a new "shadowban" role which is like suspended, but works slightly differently.

TODO:
- tests covering message thread/inbox changes
- tests covering notification changes
- tests covering profile description changes
- split into small easy-to-review PRs:
   - https://github.com/Trustroots/trustroots/pull/1180
   - https://github.com/Trustroots/trustroots/pull/1179
   - https://github.com/Trustroots/trustroots/pull/1181
   - https://github.com/Trustroots/trustroots/pull/1182
   - https://github.com/Trustroots/trustroots/pull/1183

### To test

- Create two users and log in different browser windows (e.g. use incognito). Confirm emails to both users (e.g. via http://localhost:1080 or switching `public` value to `true` in their profiles).

- Add `shadowban` to another user's `roles` array in the `users` DB

- Confirm that shadowban user can look at their own profile and other people's profiles

- Confirm that other users don't see shadow users profile

- Shadow users can look at other people's offers on the map. Likewise, regular users will see shadow users offers on the map. Hiding these can be done in follow-ups.

- Confirm that shadowban user _can_ send messages to others and can see those message threads

- Confirm that regular users cannot view messages from the shadowban user. Unfortunately, they still see the thread in their inbox and one "unread message". This is identical to how it works with "suspended" users, we'll have to work on that in follow-ups anyway.

    Screenshot 2020-01-18 at 17 05 13
    Screenshot 2020-01-18 at 17 26 33

- Confirm that shadowban user cannot remove themself, while a regular user can

   Screenshot 2020-01-18 at 17 11 42

- Add `suspended` role and confirm that you cannot login with that user

- Add `admin` role and confirm you can login and see  admin dash at `http://localhost:3000/admin`

- Write some phone numbers, emails, and URLs in the profile description of a regular user. Regular uers see those just fine, but shadowbanned user won't:
 
     
    Screenshot 2020-01-18 at 18 10 26

- I also modified notifications behavior for shadowban + suspended users but I have to write testing instructions for that. :-)

PR #1179, opened 2020-01-19
Profiles: hide profiles with 'shadowban' role

simison:

#### Proposed Changes

* Adds extra protection by filtering out suspended users.
* Adds a new "shadowban" role which is like suspended, but works slightly differently.

#### Testing Instructions

- Confirm that shadowban user can look at their own profile and other people's profiles

- Confirm that other users don't see shadow users profile

- Confirm that as a regular user, you still see regular user profiles

- Confirm that adding `suspended` role to user stops you from being able to login

### DIFF
--- a/modules/users/server/controllers/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users.authentication.server.controller.js
@@ -259,7 +259,7 @@ exports.signin = function (req, res, next) {
     }
 
     // Don't let suspended users sign in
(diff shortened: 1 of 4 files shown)


simison:
Rebased.

@rumwerfer if you're testing, you should reset the branch with:

```bash
git checkout update/shadow-ban-hide-profile
git reset --hard origin/update/shadow-ban-hide-profile
```

simison:
Thanks!

> (shadow-banned user is still visible on the map, though)

Yep, I'm planning on handling that in a separate PR.

PR #1180, opened 2020-01-19
User removal: don't let suspended users remove themself

simison:

#### Proposed Changes

* Adds a new "shadowban" role which is like suspended, but works slightly differently.
* Doesn't let shadowbanned users remove themself.

#### Testing Instructions

- Confirm that shadowban user cannot remove themself, while a regular user can

   Screenshot 2020-01-18 at 17 11 42


simison:
I've smoked tested this pretty thoroughly. 👍

rumwerfer:
working as expected :)

PR #1182, opened 2020-01-19
Do not send notifications to/from shadowbanned users

simison:
Part of https://github.com/Trustroots/trustroots/issues/1184

#### Proposed Changes

* Do not send notifications to shadowbanner users
* Do not send notifications about messages from shadowbanner users; this was pretty confusing as we don't show up their messages in people's inboxes either, but we'd still send the message notification.

#### Testing Instructions

- Send message from the banned user, and to the banned user

- Send also a message between two regular users

- Make sure not to read messages via web interface; notifications are sent only for messages that are unread.

- Using maildev (http://localhost:1080/), see that no notifications are sent to/from banned user

### DIFF
--- a/modules/messages/server/jobs/message-unread.server.job.js
+++ b/modules/messages/server/jobs/message-unread.server.job.js
@@ -228,7 +228,10 @@ function sendUnreadMessageReminders(reminder, callback) {
         // Remember to add these values also userNotFound object (see below)
         if (userIds.length > 0) {
           User.find(
-            { _id: { $in: userIds } },
+            {
+              _id: { $in: userIds },
+              roles: { $nin: ['suspended', 'shadowban'] },
+            },
             [
               // Fields to get for each user:
               'email',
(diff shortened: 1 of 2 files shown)


nicksellen (approved):
Looks good! I didn't check more than a quick code overview.

One bit I couldn't verify and there wasn't a test for is making sure the message still gets marked as notified as it seems it would trigger this code block, the comment says it will still mark the message, but I couldn't follow the code very easily:

https://github.com/Trustroots/trustroots/blob/734c346043deccaef4ade08be594e7b20f1ad58a/modules/messages/server/jobs/message-unread.server.job.js#L291-L300

The question is where the notifications get passed back to extract the message ids here: https://github.com/Trustroots/trustroots/blob/734c346043deccaef4ade08be594e7b20f1ad58a/modules/messages/server/jobs/message-unread.server.job.js#L375-L382

If it is all fine, maybe a test to check the message has been marked.

Also, the comment could be updated to say removed, suspended, or shadowbanned... or something reflecting reality.

PR #1183, opened 2020-01-19
Shadow ban: hide messages

simison:

#### Proposed Changes

* Don't show messages from shadowbanned user to regular user

Not in this PR and should be handled in follow-up PRs:
* Notifications still go through (https://github.com/Trustroots/trustroots/pull/1182)
* Message thread is visible in the inbox
* Unread message count might stay visible (pre-existing https://github.com/Trustroots/trustroots/issues/596) 

#### Testing Instructions

- Confirm that the shadow-banned user can send messages to a regular user and see those messages in their inbox and message thread

- Confirm that regular user sees the message in their inbox and gets notification (via http://localhost:1080/#/) (both aspects to be hidden in separate PR):
    Screenshot 2020-01-19 at 15 14 46

- Confirm that regular user cannot open the thread or reply to it:
   In this PR you'll see just loading spinner as the user does resolve but message thread doesn't, and we don't have handling for this on the frontend:
   
    Screenshot 2020-01-19 at 15 13 51

    If you cherry-pick commit from hiding profiles PR (https://github.com/Trustroots/trustroots/pull/1179), you can see how thread never even loads.
    ```
    git cherry-pick 6cab8940443b3242b86327a602f909799176389b
    ````
    Screenshot 2020-01-19 at 15 15 39

- If you open message thread _before_ the other user has `shadowban` role, you can try sending them a message after adding the shadowban role and without refreshing the page. You should see this error:
    image

- Should work even with various combinations of `public: false/true` state of users.

- Confirm that you can still send and see messages to/from other regular users

### DIFF
--- a/modules/messages/server/controllers/messages.server.controller.js
+++ b/modules/messages/server/controllers/messages.server.controller.js
@@ -413,32 +413,51 @@ exports.send = function (req, res) {
  * Thread of messages
  */
 exports.thread = function (req, res) {
-  res.json(req.messages || []);
+  // Sanitize messages
+  const messages = req.messages && req.messages.length ? sanitizeMessages(req.messages) : [];
+
+  res.json(messages);
 };
 
 /**
  * Thread middleware
  */
 exports.threadByUser = function (req, res, next, userId) {
+  if (!req.user) {
+    return res.status(403).send({
+      message: errorService.getErrorMessageByKey('forbidden'),
+    });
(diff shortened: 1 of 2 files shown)


rumwerfer:
All that works as expected.
(When A opens message thread with B before B gets shadow-banned, A sees another error though: "Member you are writing to does not exist.")

PR #1190, opened 2020-01-26
Generate fake test data

mrkvon:
#### Proposed Changes

* generate data in tests automatically

#### Testing Instructions

* `git switch fake-test-data`
* `npm run test:client`
* see that the tests pass, especially the TribesPage component tests

There is an [ongoing discussion](https://github.com/trustroots/trustroots/pull/1169#discussion_r364297251) whether to write fixed test data or generate them. I'm not opinionated and I don't mind merging this or closing it unmerged. :)

Follow up of #1169


modules/tribes/tests/client/components/TribesPage.component.tests.js
```diff
@@ -69,12 +40,20 @@ const renderAndWaitForTribes = async ({ user }) => {
 
 describe('TribesPage', () => {
 
+  const fake = generateFakeData({ users: 1, tribes: 5 });
```
mrkvon: FYI We need to run this outside `beforeEach` because we need the data outside of `it`.

Here:
https://github.com/Trustroots/trustroots/blob/bd09f9102da595d5e41853be743e79b1e427f440/modules/tribes/tests/client/components/TribesPage.component.tests.js#L126

modules/core/tests/client/helpers/generate-fake-data.js
```diff
@@ -0,0 +1,104 @@
+import faker from 'faker';
+import sample from 'lodash/sample';
+import sampleSize from 'lodash/sampleSize';
```
simison: Probably no difference in tests if you group `{ sample, sampleSize } from 'lodash'` or not since we don't need to worry about bundle size here. Up to you.

simison:
I'd expect future evolution here go towards:
- separating users & tribes to their own methods so that we can fake just users without tribes
- share these [with API route tests](https://github.com/Trustroots/trustroots/blob/53a825cbd21e3b3ad9d8e8a1151f3d7f46cb1e12/modules/tribes/tests/server/tribes.server.routes.tests.js)
- share with CLI fake content generator as you mention in code

modules/core/tests/client/helpers/generate-fake-data.js
mrkvon: Makes sense. :+1:

2b2dde6c1b41b6722f797191c9f97fa6b0069124

modules/core/tests/client/helpers/generate-fake-data.js
```diff
@@ -0,0 +1,103 @@
...
+    return this.displayName
+      .toLowerCase()
+      .replace(/'/g, '')
+      .replace(/\s/g, '');
+  },
+  memberIds: selectRandom(tribes, 0.4).map(tribe => tribe._id),
+});
+
+const generateTribes = count =>
+  [...Array(count).keys()].map(i => generateTribe(i));
```
mrkvon: This could be replaced with `_.range`.

mrkvon:
We already have `./testutils` for the fake data generators. Refactored to use those.

nicksellen:
I pressed the "update branch" button, and if it all passes will merge :)

mrkvon:
@nicksellen Thank you!

PR #1193, opened 2020-02-01
Changed individuals' contact to support email addr to join Slack chat.

Akronix:
Not sure if we are too concerned about web spiders crawling mailing address, in such case we should change support@trustroots.org by support (at) trustroots (dot) org or something alike.

#### Proposed Changes

* Change Guaka and Mikale contact by support@trustroots.org to join Slack chat.


simison:
Instead of adding the email address, you could add a link to https://www.trustroots.org/support which basically sends an email to that address.

PR #1194, opened 2020-02-01
Add members count to home landing page.

Akronix:
This is just a small exercise. We can discuss if it's valuable enough to be merged, but I'd like to do it just for the sake of learning more about the codebase.

#### Proposed Changes

* Add actual number of members to home landing page.

#### Testing Instructions

* Go to the home screen as a not authenticated user. Look for text: `Trustroots is over members strong! `


Akronix:
I got stuck with this error:
`Error: "[$injector:unpr] Unknown provider: statisticsDataProvider <- statisticsData <- HomeController`

modules/pages/client/config/pages.client.routes.js
```diff
@@ -169,7 +169,18 @@ function PagesRoutes($stateProvider) {
...
     });
   } else {
-    $stateProvider.state('home', { url: '/' });
+    $stateProvider.state('home', {
+      url: '/',
+      controller: 'HomeController',
+      controllerAs: 'stats',
+      resolve: {
+        // A string value resolves to a service
+        SettingsService: 'Home',
```
simison: This should be `UserCountService: 'UserCountService'` to work. (It's some Angular weirdness)

modules/pages/client/controllers/home.client.controller.js
```diff
@@ -36,6 +37,15 @@ function HomeController($stateParams, $window, TribesService, TribeService) {
   // Exposed to the view
   vm.boardHeight = $window.innerWidth <= 480 && $window.innerHeight < 700 ? 400 : $window.innerHeight - headerHeight + 14;
 
+  vm.statisticsData = statisticsData;
```
simison: No need to expose this if you're already exposing `vm.usersNo`, you can just default the value of `usersNo` to `40,000+` and use it directly at the template without using `ng-if`.

modules/pages/client/config/pages.client.routes.js
```diff
@@ -169,7 +169,18 @@ function PagesRoutes($stateProvider) {
       footerHidden: true,
     });
   } else {
-    $stateProvider.state('home', { url: '/' });
+    $stateProvider.state('home', {
+      url: '/',
+      controller: 'HomeController',
+      controllerAs: 'stats',
+      resolve: {
```
simison: I don't remember now but `resolve` at route level might actually not let you render the view before this data has been fetched — therefore it becomes blocking request for the landing page which slows it down. I'd prefer if you just do `UserCountService.get()` directly in the controller in that case.

Again I'm not 100 sure if this is really blocking the view.

simison:
Hmm. This query can be quite expensive to do on frontpage. It's not cached and now every hit on the `trustroots.org` would do `Users.count()` 😁 

It might or might not be an issue but my gut feeling tells not to add this to the frontpage.

Issue #1192, opened 2020-01-28
Add 'How to report a member' question to FAQ


noahsmindfuck:
There is a question missing in the FAQ:
Q:
>How do I report a member that violates the [rules](https://www.trustroots.org/rules)?

A:
>If you navigate to the profile of a user you will find a link that says "Report member" that will lead you to a report form. Also, at any point you can contact support@trustroots.org with your report. Please include a link to the profile or a username if you provide the report via email.

The "Report member" link is still not visible on small screens which means on all phones afaik. #919

noahsmindfuck:
A short how-to from Mikael in [Slack](https://trustroots.slack.com/archives/CHWDNEDTR/p1579873109000600)
>1. Add the question here so that it appears at the sidebar: https://github.com/Trustroots/trustroots/blob/3dfb08815cb17d2643c01ada2f1c986e8f33a65a/modules/pages/client/views/faq.client.view.html
2. Add the answer to one of the faq-*  files here: https://github.com/Trustroots/trustroots/tree/3dfb08815cb17d2643c01ada2f1c986e8f33a65a/modules/pages/client/views
Note that id has to be correct in both files for that question.

Pramuspl:
Shouldn't be closed? Seems to have been resolved.

simison:
True, thanks for the heads-up!

PR #1196, opened 2020-02-05
added "How to report a member" question to FAQ

orama254:
Greetings team,

I am submitting this pull request with reference to  issue #1192 

Fixes #
Added an additional question to the FAQ page with instructions from issue #1192

### DIFF
--- a/modules/pages/client/views/faq-general.client.view.html
+++ b/modules/pages/client/views/faq-general.client.view.html
@@ -122,5 +122,5 @@
 
 <div class="faq-question" id="how-do-i-report-a-member-that-violates-the-rules">
   <h3>How do I report a member that violates the rules?</h3>
-  If you navigate to the profile of a user you will find a link that says "Report member" that will lead you to a report form. Also, at any point you can contact support@trustroots.org with your report. Please include a link to the profile or a username if you provide the report via email.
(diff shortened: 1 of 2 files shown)


modules/pages/client/views/faq-general.client.view.html
```diff
@@ -119,3 +119,8 @@ <h3>How can I contact you?</h3>
     <a href="https://www.facebook.com/trustroots.org/">Facebook</a>
     and to directly contact individuals, see <a ui-sref="team">the team</a> page.
 </div>
+
+<div class="faq-question" id="how-do-i-report-a-member-that-violates-the-rules">
+  <h3>How do I report a member that violates the rules?</h3>
+  If you navigate to the profile of a user you will find a link that says "Report member" that will lead you to a report form. Also, at any point you can contact support@trustroots.org with your report. Please include a link to the profile or a username if you provide the report via email.
```
simison: > Also, at any point you can contact support@trustroots.org with your report. 

@noahsmindfuck @rumwerfer what do you think, would it be better to just direct folks to support form instead of asking them to send an email?

```html
<a ng-href="support">support</a>
```

noahsmindfuck: yeah,sure, it's enough

rumwerfer: oh yeah, that would be better indeed

PR #1225, opened 2020-02-14
Schedule Renovate run only on weekends

simison:
#### Proposed Changes

* Run Renovate only on weekends

https://docs.renovatebot.com/presets-schedule/#scheduleweekends

* Other limitations that are useful? Better if this runs only on Fridays, ready for merging on weekends and not taking CI capacity when everyone is already busy coding on weekends?

#### Testing Instructions

*  Renovate will validate this and show up in Github tests section
    image

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "assignees": ["nicksellen", "simison"],
   "extends": ["config:base"],
+  "schedule": "every weekend",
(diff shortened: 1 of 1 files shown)


simison:
Yeah I think it has a lot to start with, you're right.

I also dislike the automatic rebases for open PRs.

PR #1264, opened 2020-02-17
Add sentry

### LINKED ISSUE #1197: Set up JavaScript error tracking to production
- [x] Create TR account at https://sentry.io/pricing/ (@simison)
- [x] Share account with @nicksellen and anyone else interested  
- [ ] Add scripts to the app, remember to add domains to [CSP headers](https://github.com/Trustroots/trustroots/blob/a2e3ba06f3dd0abf1661b48ce6dd9da7a1eb4bfd/config/lib/express.js#L221) https://github.com/Trustroots/trustroots/pull/1264
- [ ] check sensitive information won't be sent to sentry (e.g. passwords)
- [ ] Add to `/docs`

nicksellen:
#### Proposed Changes

* add sentry to record errors to https://sentry.io
* record client errors, server errors, and worker errors

new config option default:

```js
  sentry: {
    enabled: false,
    options: {
      dsn: '',
    },
  },
```

### Locally

* checkout add/sentry branch, install, start, etc...
* create an account on sentry
* create a personal sentry project
* add sentry dsn to your `local.js` config (see above for config example)
* visit http://localhost:3000/offer/meet (which always causes an error)
* check your console for requests to sentry
* view the issue in your sentry issue lists

### Using dev2 site

Currently, for now, https://dev2.trustroots.org is configured with this branch, and with sentry reporting set up.

* create an account on sentry
* ask @nicksellen or @simison to invite you to the trustroots org
* visit https://dev2.trustroots.org/offer/meet (which always causes an error)
* check your console for requests to sentry
* view the issue in sentry https://sentry.io/organizations/trustroots-foundation/issues/ (you can mark all the existing issues as resolved to help see your one)

Fixes #1197

### DIFF
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -22,6 +22,19 @@ const git = require('git-rev');
 const path = require('path');
 const paginate = require('express-paginate');
 const uuid = require('uuid');
+const Sentry = require('@sentry/node');
+
+module.exports.initSentryRequestHandler = function(app) {
+  if (config.sentry.enabled) {
+    app.use(Sentry.Handlers.requestHandler());
+  }
+};
+
+module.exports.initSentryErrorHandler = function(app) {
+  if (config.sentry.enabled) {
+    app.use(Sentry.Handlers.errorHandler());
+  }
+};
 
 /**
  * Initialize local variables
@@ -360,6 +373,7 @@ module.exports.initHelmetHeaders = function(app) {
           'fcm.googleapis.com',
           'maitreapp.co', // Signup waiting list feature
           'www.facebook.com',
+          'https://sentry.io',
         ],
 
         // Allows control over Flash and other plugins.
@@ -518,6 +532,9 @@ module.exports.init = function(connection) {
   // Initialize express app
   const app = express();
 
+  // Initialize sentry request handler, must be first
+  this.initSentryRequestHandler(app);
+
   // Initialize local variables
   this.initLocalVariables(app);
 
@@ -547,6 +564,9 @@ module.exports.init = function(connection) {
 
   // Initialize modules server routes
   this.initModulesServerRoutes(app);
+
+  // Initialize sentry error handler, must be after routes, but before error handlers
+  this.initSentryErrorHandler(app);
 
   // Initialize error routes
   this.initErrorRoutes(app);
--- a/config/lib/express.js
+++ b/config/lib/express.js
@@ -48,6 +48,7 @@ module.exports.initLocalVariables = function(app) {
   app.locals.facebookPage = config.facebook.page;
   app.locals.googlePage = config.google.page;
   app.locals.googleAnalytics = config.googleAnalytics;
+  app.locals.sentry = config.sentry;
   app.locals.languages = languages;
   app.locals.env =
     ['development', 'test', 'production'].indexOf(process.env.NODE_ENV) > -1
--- a/modules/core/client/app/config.js
+++ b/modules/core/client/app/config.js
@@ -1,6 +1,7 @@
 /* eslint no-var: 0 */
 
 import angular from 'angular';
+import compact from 'lodash/compact';
 
 import bootstrapModules from '@/config/client/bootstrap';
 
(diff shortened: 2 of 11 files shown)


modules/core/client/app/config.js
```diff
@@ -37,14 +38,21 @@ import ngreact from 'ngreact';
 
 import 'angular-waypoints/dist/angular-waypoints.all';
 
+// eslint-disable-next-line angular/window-service
+const SENTRY_DSN = window.SENTRY_DSN;
+
+if (SENTRY_DSN) {
+  require('@/config/client/sentry').init(SENTRY_DSN);
```
simison: Should we have this on top of the file before other imports?

config/lib/express.js
```diff
@@ -360,6 +374,7 @@ module.exports.initHelmetHeaders = function(app) {
           'fcm.googleapis.com',
           'maitreapp.co', // Signup waiting list feature
           'www.facebook.com',
+          'https://sentry.io',
```
simison: I think this should be just `sentry.io` without protocol.

modules/core/server/views/layout.server.view.html
```diff
@@ -17,7 +17,12 @@
     <meta name="fragment" content="!" />
     {% for cssFile in cssFiles %}
     <link rel="stylesheet" href="/{{cssFile}}?c={{appSettings.commit}}" />
-    {% endfor %} {% if env === 'development' %} {# in dev load js in head, as it
+    {% endfor %} {# Sentry DSN - make sure this is above scripts #} {% if
```
simison: I was about to comment that I'd just split these `endfor`+comment+`if` to their own lines for easier readability but realized the whole file is like this. 😁

modules/core/server/views/layout.server.view.html
```diff
@@ -17,7 +17,12 @@
...
+    sentry.enabled === true %}
+    <script>
```
simison: This will need `nonce="{{ nonce }}"` to work past our security stuff. See other `<script>` tags in this file.

See this in console without it:
image

modules/core/server/views/layout.server.view.html
```diff
@@ -17,7 +17,12 @@
...
+      window.SENTRY_DSN = '{{ sentry.options.dsn }}';
```
simison: We could also pass this via Webpack config — I'm not tied to the way or the other tho so fine as is.

nicksellen: I actually did it like that originally, but then realised we may not have the sentry configuration available at the time the webpack build is done (esp. in this new approach where the build is done in CI).

I think it would be confusing if some of the stuff in `config/` would update immediately and some only after a webpack rebuild.

Given there is a mechanism to pass config from server to client I think it's nice to expand on this (... and could be made a more _proper_ and _general_ mechanism in a future enhancement!)

modules/core/server/views/layout.server.view.html
nicksellen: Ah, yes! I did the manually on dev2.trustroots.org after noticing that, but didn't update it properly...

nicksellen: Fixed in https://github.com/Trustroots/trustroots/pull/1264/commits/81e8ff1c47eb2854468fc3d52b27a8d1b41b139d

modules/core/server/views/layout.server.view.html
nicksellen: Yes, I think the prettier did it like that...

config/lib/express.js
nicksellen: I always include protocol too, I'll keep it like this for now, should consider changing them all to be consistent, but at least having one different provokes the question until it's resolved ;)

modules/core/client/app/config.js
nicksellen: I think dynamic imports should be after static imports, logic is that the static imports can never be conditionally imported or not, so not subject to the logic of the file, and to me the logic begins where the static imports end...

... but I think the counterpoint you are maybe thinking about, is what if there is an error during the import process? ... they are all library imports though so should not be a problem.

I would just keep it like this for now... to get it merged.

PR #1266, opened 2020-03-02
Automated deployments to dev site

nicksellen:
#### Proposed Changes

* creates `trustroots.tar.gz` containing a fully built setup
* can be used for deploying to a server (e.g. via https://github.com/Trustroots/trustroots-debops/)
* gets uploaded to DigitalOcean spaces s3 compatible storage (downloadable from https://trustroots-builds.fra1.digitaloceanspaces.com/trustroots.tar.gz)
* only keeps latest build (as fixed filename `trustroots.tar.gz`) - this is mostly so it does not need any logic for pruning old versions...

It's import to note that the build is architecture/environment-specific (think of those node deps that need compiling) - I switched travis to use ubuntu bionic - which corresponds with debian buster (which is what dev2.trustroots.org is running) - and hopefully is always compatible! ... otherwise we can either switch dev2 to use ubuntu (not in favour), or use docker to build the tarball using a buster image (gives more flexible, not tied to what travis supports).

... and it works! https://dev2.trustroots.org/ was deployed automatically via travis.

#### TODO

- [ ] ~fix broken server tests https://travis-ci.org/Trustroots/trustroots/jobs/657267561~ (was probably just temporary error)
- [ ] ~consider naming it `trustroots-<gitref>.tar.gz` (e.g. `trustroots-master.tar.gz` to make it clearer what it is)~ nah, it's fine for now
- [x] run dev2.trustroots.org deployment inside travis
- [ ] ~consider restricting permissions of deploy user (currently has full access to dev2.trustroots.org which is kind of ok as it's just a throwaway server...)~ fine for now...
- [ ] ~consider plan for "production" releases (but maybe we just ignore this for now...)~ delay until we need to think about that...
- [x] see if I can modify the "encrypted_*" environment variables in https://travis-ci.org/Trustroots/trustroots/settings to be only available to master branch (they were automatically added by `travis encrypt-file`) - alternatively I can just regenerate my own ones manually... (I could get the CLI to show the values, then delete/re-add them in the travis UI as restricted to master branch)

#### Testing Instructions

* hard to test...
* you can look at a successful deployment travis job https://travis-ci.org/Trustroots/trustroots/jobs/657355778
* ... I think it needs to merged, then refined there if there are still issues ... especially as it's limited to master branch now

Fixes #


nicksellen:
The first auto-deployed job just completed successfully! ---> https://travis-ci.org/Trustroots/trustroots/jobs/657355778

simison:
Super lightweight initial pass and this is pretty exciting!

`deploy` section gets run on Travis only on merge, correct?

Starting some `/docs/deploy` would be good to explain the setup, either in this PR or follow-up.

I'll do another pass later on.

.travis.yml
```diff
@@ -1,36 +1,63 @@
...
+    secure: KZ/M6Nnn3raktHWjYs0ddoRS/1ojfQpBXKiqroJMh1k2qy07j/GM5j/CKZH+aJGB5AU9QxKahF+zrQs2Gh5Uo3Uoo9gpVGUxqTb2Ok57zUBw3lywTbYCliisIpSA+NPLCzEI/C82Db603wk8yiD8TJyZCTPWFUyhGNUVKSy4Y/w=
+  bucket: trustroots-builds
+  endpoint: https://fra1.digitaloceanspaces.com
+  glob: "trustroots.tar.gz*"
+  acl: public_read
+  verbose: true
+  # https://docs.travis-ci.com/user/deployment-v2/conditional
+  on:
+    branch: master
+    condition: "$DEPLOY == 'true'" # only run for the build job
```
nicksellen: > deploy section gets run on Travis only on merge, correct?

Yup, this is the bit that controls it:
- `master` branch only (to get it working in a branch/PR I had it running for all branches before)
- only if the $DEPLOY variable is set to true, so it only runs in the build matrix job...

.travis.yml
```diff
@@ -1,36 +1,63 @@
...
-      env: RUN="npx concurrently --kill-others-on-fail 'npm:lint' 'npm:build'"
+  - name: Test client
+    node_js: 12
+    env: RUN="npm run test:client"
+  - name: Test server
+    node_js: 12
+    env: RUN="npm run test:server"
+  - name: Lint and build
+    node_js: 12
+    env: RUN="npx concurrently --kill-others-on-fail 'npm:lint' 'npm:build:tarball'" DEPLOY="true"
```
simison: Do we need `NODE_ENV=production` here since we're building the bundle for deploying?

.travis.yml
```diff
@@ -1,36 +1,63 @@
...
-    - name: "Lint and try building"
-      node_js: 12
-      env: RUN="npx concurrently --kill-others-on-fail 'npm:lint' 'npm:build'"
+  - name: Test client
+    node_js: 12
+    env: RUN="npm run test:client"
+  - name: Test server
+    node_js: 12
+    env: RUN="npm run test:server"
+  - name: Lint and build
```
simison: Now that build+deploy might take longer, we could split the build to another container to increase parallelism. Not in this PR and only if it makes things faster of course. Just noting down.

.travis.yml
nicksellen: `NODE_ENV=production` is set in the "build" script in package.json.

And checking https://dev2.trustroots.org/assets/main.js it looks pretty minified.

.travis.yml
nicksellen: The deploy part uses the output of the build part, so they can't be parallelised.

In sequence it has to install npm deps, build the webpack stuff, bundle all up in a tarball, put the tarball on DigitalOcean Spaces, then run the deploy playbook.

The lint could be split out, but I doubt that would make much difference.

simison: Sounds good! 👍

PR #1267, opened 2020-03-06
Update dependencies

nicksellen:
Based on renovate versions, but combined into one update. I just went through all the passing ones, and updated to those versions. I left out firebase/firebase-admin because the version leap is huge and I'm not convinced they are covered by tests.


simison:
Did you confirm that influx and firebase stuff still works? Might not be covered by tests to extend it needs.

nicksellen:
I didn't upgrade firebase. I didn't confirm influx... but yes should check :)

simison:
I think the TR Docker setup contains also Influx so that's prolly the easiest way to go about it.

simison:
> I didn't upgrade firebase.

Did you mean to say didn't do major version bump on it?

https://github.com/Trustroots/trustroots/pull/1267/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R109

nicksellen:
Ah yeah quite right, I did the minor upgrade, which was an earlier renovate PR.

This is also why I'd quite like to get https://github.com/Trustroots/trustroots/pull/1266 set up, so we can try things out in a sort-of real environment (it doesn't currently have influxdb/firebase setup for it, but it can be extended for that).

PR #1286, opened 2020-03-07
Support listening on a file descriptor

nicksellen:
#### Proposed Changes

* extra optional configuration option (`fd`) to support listening on a file descriptor
* if set, replaces use of `host` / `port` options
* if not set, no difference to before
* useful for using with systemd socket activation (set `fd: 3`, see https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html), which can be used for zero down time deploys

#### Testing Instructions

* ... can verify it still works under normal conditions
* probably too fiddly to explain how to actually use the new option here... configuration will be in https://github.com/trustroots/trustroots-debops at some point

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -48,6 +48,7 @@ module.exports = {
   uploadDir: './public/uploads-profile',
   port: 3000,
   host: 'localhost',
+  fd: null, // listen on a file descriptor (instead of host/port)
   https: false,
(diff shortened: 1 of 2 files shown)


nicksellen:
Gunna merge now, so I can keep working on this. But feel free to still add comments/etc, and I can incorporate them in future PRs.

PR #1287, opened 2020-03-07
Fix push messaging config

nicksellen:
When adding some fcm config for dev2.trustroots.org I noticed it didn't work. Then I noticed it didn't work on production either.

tr-push-error

In the switch to use webpack I changed it to get the FCM_SENDER_ID through webpack... _but_ webpack build time is not a good place to rely on having configuration available (evidently).

#### Proposed Changes

The `/config/sw.js` endpoint already exists and sets a global FCM_SENDER_ID variable and was used before, so I just reverted to using this method.

I tried it locally and could observe the error with the old way, and this change fixing it such that I could receive push messages.

#### Testing Instructions

* add some fcm config to your local setup
* try enabling it!
* you should receive a push notification (to tell you that you have enabled push notifications)

(also just to note, changes to the service worker are _not_ automatically rebuilt, you need to run `npm run webpack:service-worker` if you change something).

### DIFF
--- a/config/webpack/entries/pushMessagingServiceWorker.js
+++ b/config/webpack/entries/pushMessagingServiceWorker.js
@@ -1,4 +1,4 @@
-/* global FCM_SENDER_ID */
+/* global FCM_SENDER_ID,importScripts */
 
 /**
  *  This will not rebuild on change, so if you're doing development you might want to rebuild it manually:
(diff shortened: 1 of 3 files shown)


nicksellen:
I'll merge this in now as it's currently broken, so can't get much worse ;)

But very open to post-merge reviews, as we can change.  Also please complain at me if you don't like my eager merges...

nicksellen:
It works on https://dev2.trustroots.org now :)

(which also addresses the point in https://github.com/Trustroots/trustroots/pull/1267#issuecomment-595724302)

simison:
No worries, good call to merge already. I can look tonight separately.

PR #1288, opened 2020-03-08
[deprecated] Automate translation extraction with puppeteer

mrkvon:
- start the app (e.g. with `npm start`)
- run `node bin/browse.js` and see that the finnish translation file `public/locales/fi/volunteering.json` gets created.

It happens with [puppeteer](https://www.npmjs.com/package/puppeteer) which runs headless chrome browser.

#### Proposed Changes

*

#### Testing Instructions

*

Will be rendered obsolete by #1291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,9 @@ matrix:
     env: RUN="npm run test:client"
   - name: Test server
     node_js: 12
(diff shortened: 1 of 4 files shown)


mrkvon:
This PR is very likely not going to be used, because #1291 extraction is more robust and simpler.

PR #1289, opened 2020-03-08
Convert translations from json to po and vice versa

mrkvon:
#### Rationale

There are [tools](https://poedit.net/) [available](https://www.transifex.com/) for translating the gettext (.po) translations. Not so many for [i18next](https://locize.com/) format.

Therefore, let's keep both the `json` and `po` versions of the translations. This PR will help keep them in sync. Developers can generate `json` files and translators can translate `po` files.

#### Proposed Changes

* a script to convert translation files in `public/locales/{locale}/` from json to po and vice versa
* pre-commit hook to do the conversion, when translation files are changed.

There may be more elegant solutions than this. I'm interested to hear about them, please!

_If you can think of a better name or location for the tests, please share._

#### TODO

- [x] write the conversion scripts
- [x] make a pre-commit hook which will automatically convert json <-> po and keep them in sync
- [x] refactor, comment
- [x] format the output json file (has 4 spaces for tab instead of 2)

#### Testing Instructions

0. `git switch convert-translations-po`
0. json to po
    - create or edit files in `public/locales/{locale}/{translation}.json`
    - `git add . && git commit -m "test"`
    - see that their `.po` counterparts were generated/updated
0. po to json
    - create or edit files in `public/locales/{locale}/{translation}.po`
    - `git add . && git commit -m "test"`
    - see that their `.json` counterparts were generated/updated
0. both directions
    - create or edit various `json` and `po` files in `public/locales`
    - `git add . && git commit -m "test"`
    - see that their counterparts were generated/updated
0. both directions with conflict
    - create or edit both `json` and `po` with the same name and locale, e.g. `public/locales/en/volunteering.{json|po}`
    - `git add . && git commit -m "test"`
    - see that the pre-hook will complain about conflict between json and po file
0. use script directly
    - run `npm run convert-translations public/locales/**/*.json`
    - see that everything got converted to po.


package.json
```diff
@@ -262,6 +264,10 @@
     }
   },
   "lint-staged": {
+    "public/locales/**/*.{json,po}": [
+      "npm run convert-translations",
+      "git add public/locales/**/*.json public/locales/**/*.po"
```
simison: This `git add` might not be required since `lint-staged` already does it for us — unless the specific path requires it?

bin/i18n/convert.js
```diff
@@ -0,0 +1,74 @@
...
+function changeExtension(file, extension) {
+  const parsed = path.parse(file);
+  delete parsed.base;
+  parsed.ext = extension;
+  return path.format(parsed);
+}
+
+function getLocale(file) {
+  const directory = path.parse(file).dir;
+  return directory.split(path.sep).slice(-1)[0];
```
simison: `directory.split(path.sep).pop()` should do the same.

bin/i18n/convert.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+}
+
+async function convertFile(file) {
+  const extension = path.parse(file).ext;
+
+  const extensions = ['.json', '.po'];
+
+  if (!extensions.includes(extension))
```
simison: Hm, odd, I expected Prettier to add `{}` here.

bin/i18n/convert.js
mrkvon: Changed in 04dd8f8cf4264c91b06ad72a7700f35ff46309cc

bin/i18n/convert.js
mrkvon: Yes, interesting. Anyways, added it manually in 04dd8f8cf4264c91b06ad72a7700f35ff46309cc

package.json
mrkvon: I agree with this observation, I also hoped the converted files would get staged automatically. However, they are left behind without this.

simison:
Rebased.

simison:
What do you think about placing `modules/i18n/tests/server/convert-translation.bin.tests.js` under `bin/i18n/test/*` so that it's easier to find and connect what it's testing?

package.json
```diff
@@ -263,6 +265,10 @@
     }
   },
   "lint-staged": {
+    "public/locales/**/*.{json,po}": [
```
simison: Should we hold-off from adding this into `lint-staged` yet while figure out the whole process?

simison:
`fs-extra` needs to be added in dependencies; looks like it was a transient dependency earlier. We'll need eslint for these. :-)

mrkvon:
> fs-extra needs to be added in dependencies; looks like it was a transient dependency earlier.

What happened was the following: the fs-extra was removed in #1312. When this was rebased, we started missing fs-extra here.
Now it's fixed as part of the last rebase.

package.json
mrkvon: Sure, that makes sense. Removed in 9c21c306f73376d811fe1f52e3579ef6d3c3360f

mrkvon:
> What do you think about placing modules/i18n/tests/server/convert-translation.bin.tests.js under bin/i18n/test/* so that it's easier to find and connect what it's testing?

I see the tests don't get executed when placed there. Do you think it's worth changing the gulpfile?

It seems that changing [gulpfile#L116](https://github.com/Trustroots/trustroots/blob/master/gulpfile.js#L116) and [gulpfile#L149](https://github.com/Trustroots/trustroots/blob/master/gulpfile.js#L149) would do the trick:

```
['modules/*/tests/server/**/*.js', 'bin/**/tests/**/*.js']
```
@simison I'd appreciate your opinion whether to do this, and some guidance when it comes to changing gulpfile. I haven't grokked the gulpfile, yet. :)

simison:
@mrkvon yeah, I think it makes sense. 👍 Let the Gulpfile serve us and not us serve the Gulpfile. 😅  Paths are regular [globs](https://en.wikipedia.org/wiki/Glob_%28programming%29) so your change looks good.

package.json
mrkvon: We removed the related lint-staged config in 0b8f8609efc765854a48b60d1e5d76bdaddabef0 as suggested in [a review](https://github.com/Trustroots/trustroots/pull/1289#discussion_r396008019).

mrkvon:
> https://github.com/Trustroots/trustroots/pull/1289#issuecomment-606058175

The tests moved to `/bin/i18n/tests/...` in 7ff522c02dc66324b5c97dfca4eb29f03c4c9dae

mrkvon:
_Update:_

[Weblate](https://hosted.weblate.org/projects/trustroots/) seems to become our main translation tool. It supports `i18next`, therefore the conversion between `json` and `po` seems unnecessary. Maybe nice to have, but a low priority.

nicksellen (commented):
Just trying to remove this "review request" from my list... (update: success, I didn't have an option just to cancel the review request)

mrkvon:
It seems this isn't useful since we use Weblate. Closing.

PR #1291, opened 2020-03-11
Use babel-plugin-i18next-extract to extract translations

mrkvon:
#### Proposed Changes

Configuration to extract translations with [babel-plugin-i18next-extract](https://github.com/gilbsgilbs/babel-plugin-i18next-extract)

This will make #1288 #1056 outdated.

~To work properly, we'll need to migrate all `withTranslation` to `useTranslation`. Otherwise the namespace is not recognized. #1292~ _done_

#### Usage

```sh
npm run build
```

#### Testing Instructions

- `git switch i18next-extract`
- `npm run build`
- new files in `./public/locales/{locale}` should appear. All strings should be sorted to their namespaces, so [translation.json](https://github.com/Trustroots/trustroots/blob/master/public/locales/en/translation.json) (i.e. fallback namespace) should not get updated.
- (optional) add a new locale to `./config/shared/locales.json`, run the build and see that a folder for the new locale with translation files gets generated. (`./public/locales/{newLocale}`)

#### Follow up
- ~close #1288~ _done_
- remove the endpoint for saving missing translations added in #1056 (#1312)
- extract the translations

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -25,7 +25,7 @@ module.exports = {
   ]),
   env: {
     production: {
-      plugins: compact([
+      plugins: [
         [
           // extraction of the i18next translation strings
           'i18next-extract',
@@ -38,7 +38,7 @@ module.exports = {
             discardOldKeys: true,
           },
         ],
-      ]),
+      ],
     },
   },
 };
--- a/babel.config.js
+++ b/babel.config.js
@@ -22,16 +22,23 @@ module.exports = {
     '@babel/plugin-proposal-optional-chaining',
     'angularjs-annotate',
     isDevelopment && 'react-refresh/babel',
-    [
-      'i18next-extract',
-      {
-        nsSeparator: false,
-        locales: locales.map(locale => locale.code),
-        keySeparator: false,
-        outputPath: 'public/locales/{{locale}}/{{ns}}.json',
-        keyAsDefaultValue: ['en'],
-        discardOldKeys: true,
-      },
-    ],
   ]),
+  env: {
+    production: {
+      plugins: compact([
+        [
+          // extraction of the i18next translation strings
+          'i18next-extract',
+          {
+            nsSeparator: false,
+            locales: locales.map(locale => locale.code),
+            keySeparator: false,
+            outputPath: 'public/locales/{{locale}}/{{ns}}.json',
+            keyAsDefaultValue: ['en'],
+            discardOldKeys: true,
+          },
+        ],
+      ]),
+    },
+  },
 };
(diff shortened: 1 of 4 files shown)


mrkvon:
@nicksellen You may have played with this or worked on this, have you? If yes, I'll be interested in hearing your experience and/or collaborating on this.

Also, @nicksellen, is there a better way to run this than `npm run build`? WDYT?

babel.config.js
```diff
@@ -20,5 +22,16 @@ module.exports = {
...
     isDevelopment && 'react-refresh/babel',
+    [
+      'i18next-extract',
+      {
+        nsSeparator: false,
+        locales: locales.map(locale => locale.code),
+        keySeparator: false,
+        outputPath: 'public/locales/{{locale}}/{{ns}}.json',
+        keyAsDefaultValue: ['en'],
+        discardOldKeys: true,
```
mrkvon: This option trashes all unused translations. Do we want it?

mrkvon:
I'm marking this as ready for review, although I'm not sure where to place and how to execute the extraction properly.

Currently it works with `npm run build`.

nicksellen:
This also runs in generates the file in development too (e.g. after doing `npm start`) as the plugin config isn't turned on/off for different envs. You could use the `env` config option to just enable it for production, see https://babeljs.io/docs/en/6.26.3/babelrc#env-option.

I think we need to wait till the wider translation workflow is worked out to know exactly when/how it should run, but I think this is good to merge before that anyway.

mrkvon:
@nicksellen The config is fixed as you suggested, so the translations get extracted only in production (e.g. `npm run build`) and not when `npm start` is run.

@nicksellen @simison This is ready for review, comments, approval and merge. :wink:

babel.config.js
```diff
@@ -21,4 +23,22 @@ module.exports = {
     'angularjs-annotate',
     isDevelopment && 'react-refresh/babel',
   ]),
+  env: {
+    production: {
+      plugins: compact([
```
simison: `compact` can be removed here — it can be used to remove plugins from the list conditionally but there are not conditions here.

mrkvon: :+1: removed in 965bf935c914ff98b99a97be97115ddd2ab20e82

_note: Seems like we could move all the environment conditions into the [env-option](https://babeljs.io/docs/en/6.26.3/babelrc#env-option) as Nick suggested [above](https://github.com/Trustroots/trustroots/pull/1291#issuecomment-599250830). Not in this PR._

mrkvon:
@simison Thank you!

PR #1312, opened 2020-03-19
Revert #1056 - Remove endpoint for runtime extraction of missing translations

mrkvon:
#### Proposed Changes

Revert all changes introduced in #1056, except extracted translations and some i18n config. Why? Because when #1291 gets merged, we don't need this anymore.

#### Testing Instructions

* `git switch revert-1056-remove-i18next-backend`
* see that the app still works

Reverts #1056 
Follow up of #1291


mrkvon:
@simison Thank you for the review!

RIP #1056
It had a short and unfulfilled life... :wilted_flower:

simison:
I value learnings from it quite a bit. :-) thanks for exploring it!

PR #1329, opened 2020-03-21
Migrate Search map to React

simison:
Migrate the main map search to React.

#### Proposed Changes

* Uses `react-map-gl` for Map components React library to render (already used elsewhere for other maps)
* Instead of reusing base `Map` component, extract its components and re-use to build more custom map component for search
* For now, pass data across Angular/React via props (observe changes in prop values and use callback functions). I didn't want to start adding shared data layer just yet because it could end up complicated across Angular and React. Right next up we should start migrating rest of the search UI to React and it's also a good chance to introduce data layer and dispact/listen actions via it, instead of using props and callbacks.
* Has 3 different Mapbox layers and one OSM layer for development purposes only.
* Convert API response to a valid GeoJson response thus making it more straightforward to use "natively" with mapping libraries 

Screenshot 2020-12-21 at 16 28 25

Screenshot 2020-12-21 at 16 27 14

Screenshot 2020-12-21 at 16 27 56

Screenshot 2021-01-09 at 18 23 40

#### Testing

Test `/search` thoroughly:

* Configure Mapbox token. Config & token can be copied from `config/env/local.sample.js` and place in `config/env/local.js`

* Search for locations
* Click "see in big map" from profiles to open `/search` at correct location
* Click on location names from profiles to open search at right location
* Change map layers
* Try Zooming and browsing the map with buttons, mouse scroll/drag, touch device if possible
* Refresh the page and observe map remember location and map layer
* Open offers, note how already opened dots stay faded
* Click clusters to zoom into them
* Ensure "meet" offers work
* Ensure different filters work
* Ensure mobile experience works
* Try search map with user who didn't confirm their email yet (freshly created user). No API queries should fire.


modules/search/client/less/map.less
```diff
@@ -76,3 +70,4 @@
     }
   }
 }
+*/
```
simison, line 73: TODO: clean these styles out

modules/core/client/components/Map/MapStyleControl.js
```diff
@@ -1,80 +1,126 @@
...
+                key={MAP_STYLE_OSM.name}
+                onClick={() => {
+                  this.close();
+                  setMapstyle(MAP_STYLE_OSM);
+                }}
+                type="button"
+              >
+                <MapIcon />
+                {MAP_STYLE_OSM.name}
+              </button>
```
hlibbabii, line 111: Possible candidate for a separate component?

hlibbabii:
I'll add the link to the slack thread so that we don't lose it: https://trustroots.slack.com/archives/C0A3Q15SS/p1609599737108400

simison:
Some selected bits from the [above Slack convo](https://github.com/Trustroots/trustroots/pull/1329#issuecomment-766823832):


**Faded-effect on dots**

- "I only noticed the difference between faded and not after you told me :) do you think it's well visible for others?"

- "I like how AirBnb map works", also animations are neat:
    image
    https://user-images.githubusercontent.com/87168/106379577-9da50480-63b5-11eb-86e8-50cd5bee4dc3.mov

- probably for airbnb it looks better because the background is always opaque, only the font changes opacity

- badges are also bigger, and the shadow helps to lift the badges from the map a bit better

- I hit a bit the limitations (in skill, not possibilities) with the json way of defining styles for dots but it’s something I wanna spend more time on at some point

- Also looks like the opacity thing doesn't work on mobile

**Dot visibility on higher zoom levels**

- Probably it's just me but for me from aesthetic point of view, the yellow dots look too small compared to the clusters at this zoom level:
    image

- not just you, it does disappear a bit there! At that zoom level clicks aren’t too important anyway, more to indicate that “there’s more here, zoom closer”, so I wasn’t too concerned about making them clickable. (just my assumption about the zoom level, could be wrong of course)

- yeah, i don't think we should be concerned about clicking, just about the way it looks

**OSM Bug**
I found one strange thing:
- Choose the OSM style
- Reload the page
- The offers won't load
- Then if you try clicking on "OSM"(OSM style is already chosen), it will crash:
    image


**Click-collection-to-zoom funkiness**

https://user-images.githubusercontent.com/87168/106379656-22901e00-63b6-11eb-9ec7-62099cfc113f.mov

https://user-images.githubusercontent.com/87168/106379663-2b80ef80-63b6-11eb-80b0-fde493d5f8ab.mov

https://user-images.githubusercontent.com/87168/106379666-2c198600-63b6-11eb-8f6b-281ee5d92e0b.mov

modules/core/client/components/Map/MapStyleControl.js
simison: Yup, I wasn't entirely sure about that since it's not really useful independently and it doesn't really affect line count — but I split it for now.

Thinking if to just remove OSM layer for now and that would remove the need for separate component here, too.

https://github.com/Trustroots/trustroots/pull/1329/commits/d3497344fe6ae55230765523dc4c51f9a72da33b

simison:
@hlibbabii I worked on the "click cluster => zoom" behaviour. It's a bit snappier and more accurate now (https://github.com/Trustroots/trustroots/pull/1329/commits/b36d933d2ba5ae3aeffee5d7bfffa6b7d2764229, https://github.com/Trustroots/trustroots/pull/1329/commits/fcd86844eadcbf62c75965979a4dc23ec3d726e0).

I didn't touch opacity and dot sizes; I agree it needs iteration but I think it's best left for another PR because I don't exactly know yet how it should look like. It's not perfect, but not broken either. :-)

hlibbabii:
> @hlibbabii I worked on the "click cluster => zoom" behaviour. It's a bit snappier and more accurate now ([b36d933](https://github.com/Trustroots/trustroots/commit/b36d933d2ba5ae3aeffee5d7bfffa6b7d2764229), [fcd8684](https://github.com/Trustroots/trustroots/commit/fcd86844eadcbf62c75965979a4dc23ec3d726e0)).
> 
> I didn't touch opacity and dot sizes; I agree it needs iteration but I think it's best left for another PR because I don't exactly know yet how it should look like. It's not perfect, but not broken either. :-)

@simison I tested the last changes, looks good

simison:
Thanks @hlibbabii ! I'll do another round of testing myself and deploy to production for further testing if it all works fine. It's a good base for iterating further.

PR #1332, opened 2020-03-22
Offer map: resize on window resize

simison:
When adding the new base `Map` component in https://github.com/Trustroots/trustroots/pull/1156, I intentionally left window resize-refreshing for a follow-up PR. Here it is. :-)

#### Proposed Changes

* Add window resize event handler to Map base component. Debounce event to once per second to avoid performance issues.

#### Testing Instructions

* Either at "meet" page (after adding one meet offer) or at a profile page (after adding a hosting offer)...
* Observe how zooming and panning works
* Drag browser window to different size so that map container size changes and observe map canvas accommodate new width
* Zoom/pan around, then resize again and observe zoom and location be preserved

Before the change:
image

After the change:
image


modules/core/client/components/Map/index.js
```diff
@@ -30,6 +31,21 @@ export default function Map(props) {
     zoom,
   });
 
+  useEffect(() => {
+    const debouncedHandleResize = debounce(function handleResize() {
```
simison: [`debounce`](https://lodash.com/docs/4.17.15#debounce) is to ensure we don't refresh constantly while refreshing, thus avoiding performance issue.

modules/core/client/components/Map/index.js
```diff
@@ -30,6 +31,21 @@ export default function Map(props) {
...
+      setViewport({ latitude, longitude, zoom });
+    }, 1000);
+
+    window.addEventListener('resize', debouncedHandleResize);
+
+    // Cleanup function
+    return () => {
+      window.removeEventListener('resize', debouncedHandleResize);
+    };
+  });
```
mrkvon: Would it be enough to set this event listener only once?

```js
}, []);
```

modules/core/client/components/Map/index.js
mrkvon: Could we even afford [`throttle`](https://lodash.com/docs/#throttle) and shorter time?
I think 1 second is quite long.

modules/core/client/components/Map/index.js
```diff
@@ -1,7 +1,8 @@
 // External dependencies
+import { debounce } from 'lodash';
```
mrkvon: The _[whatever does the build - webpack?]_ probably optimizes the bundle size, however `import debounce from 'lodash/debounce'` would be the safe option for client side.

nicksellen: I agree: `lodash/debounce` is best bet.

Otherwise need [https://github.com/lodash/babel-plugin-lodash](babel-plugin-lodash) to rewrite it for us...

nicksellen:
I noticed something funny. After one initial rendering with the non-resizing behave (like from your screenshots), subsequent resizing of the window seemed to be fine, so I tried _just_ doing one initial setting of the viewport, and then after it seems fine. See https://gist.github.com/nicksellen/05b1e241c6de49f07f439df244da725b for my whole file (I also removed passing in `{...props}` to `ReactMapGL` as otherwise it duplicates some of the stuff passed in as `{...viewport}`.

This seems a bit odd to me, so I suspect it's a symptom of something else not being quite configured right... digging a little more ...

simison:
Thanks for digging and thorough reviews, both! I'm not gonna be able to jump in to address this soon so I'd appreciate if either one of you can take it over for time being. :-) Thanks!

nicksellen:
I reimplemented as per my comment :)

modules/core/client/components/Map/index.js
simison: Yup! I'm just used to that being available from other codebases I work with. 😁 Should we just add the plugin tho? No big deal if you prefer not.

modules/core/client/components/Map/index.js
```diff
@@ -38,9 +38,10 @@ export default function Map(props) {
       mapStyle={MAP_STYLE_DEFAULT}
       onViewportChange={setViewport}
       touchRotate={false}
-      width="100%"
-      {...props}
```
mrkvon, line 42: By removing `props`, don't we lose the ability to reuse the `Map` component in a flexible way?
e.g. we may want to set the `Map`'s width and height, or control the state from outside (I'd like to do both of these for refactoring offer-host-edit).

Wouldn't something like the following work? (I haven't figured out details.)

```jsx
{...defaults}
{...props} /* provided props overwrite defaults */
{...viewport} /* zoom, latitude and longitude overwrite the props */
/* only zoom, latitude and longitude get updated */
onViewportChange={({ zoom, latitude, longitude }) => setViewport({ zoom, latitude, longitude })}
```

But I suspect this won't allow me to tap into the internal zoom, latitude and longitude state anyways, since the `viewport` overwrites whatever we provide. Probably fit for a different PR.

nicksellen: Ah I see, yes, we can reintroduce props, but important thing is to only pass down the props that we don't use directly... I can implement that now :)

nicksellen: The `onViewportChange` callback has loads of extra stuff it sets beyond just zoom, latitude, longitude, so I quite liked keeping it intact.

See https://github.com/Trustroots/trustroots/pull/1332/commits/8936d4095e1dd44d90d222c884a765ac43ecf6e6 for passing down extra props. Although I guess we'd get a lint error if we don't declare what we want to pass down in the prop types...

PR #1333, opened 2020-03-23
Map: add style control

simison:
#### Proposed Changes

* Add a control to change between satellite/streets map layers in base `Map` component

Screenshot 2020-04-12 at 22 51 30
Screenshot 2020-04-12 at 22 51 26
Screenshot 2020-04-12 at 22 51 21

#### Testing Instructions

* Add `showMapStyles={true}` prop in `modules/offers/client/components/OfferLocation.component.js` to test it.
* Otherwise the component isn't in use visibly anywhere
* This component will be useful for the main map canvas and I just wanted to Pr this separately to keep migration more focused.

### DIFF
--- a/modules/core/client/components/Map/MapIcon.js
+++ b/modules/core/client/components/Map/MapIcon.js
@@ -1,0 +1,34 @@
+// External dependencies
+import PropTypes from 'prop-types';
+import React from 'react';
+
+/**
+ * Map icon
+ * @link https://docs.mapbox.com/playground/static/
+ * @link https://docs.mapbox.com/api/maps/#static-images
+ */
+export default function MapIcon({ mapboxStyle, mapboxToken }) {
+  const image =
+    mapboxStyle && mapboxToken
+      ? `https://api.mapbox.com/styles/v1/mapbox/${mapboxStyle
+          .split('/')
+          .pop()}/static/14.1663,55.6438,5.22,0/128x128@2x?attribution=false&access_token=${mapboxToken}`
+      : // Default image when API isn't available
+        'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNc8mTSfwAHaQMbL4UQfQAAAABJRU5ErkJggg==';
+
+  return (
+    <img
+      className="img-circle"
+      alt=""
+      aria-hidden="true"
+      width="24"
+      height="24"
+      src={image}
+    />
+  );
+}
+
+MapIcon.propTypes = {
+  mapboxStyle: PropTypes.string,
+  mapboxToken: PropTypes.string,
+};
--- a/modules/core/client/components/Map/MapStyleControl.js
+++ b/modules/core/client/components/Map/MapStyleControl.js
@@ -1,0 +1,74 @@
+// External dependencies
+import { useTranslation } from 'react-i18next';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+// Internal dependencies
+import {
+  MAPBOX_TOKEN,
+  MAP_STYLE_MAPBOX_STREETS,
+  MAP_STYLE_MAPBOX_SATELLITE,
+  MAP_STYLE_MAPBOX_OUTDOORS,
+  MAP_STYLE_OSM,
+} from './constants';
+
+/**
+ * The main LanguageSwitch component.
+ * @param {'dropdown'|'select'} presentation - should we show dropdown (header), or select (account) selector?
+ * @param {Boolean} [saveToAPI=false] - should we save selected language to API?
+ */
+export default function MapStyleControl({ mapStyle, setMapstyle }) {
+  const { t } = useTranslation('core');
+
+  return (
+    <div
+      className="btn-group-vertical btn-group-xs"
+      role="group"
+      aria-label={t('Map style')}
+    >
+      <button
+        className={classnames('btn btn-default', {
+          'btn-primary': mapStyle === MAP_STYLE_MAPBOX_STREETS,
+        })}
+        disabled={!MAPBOX_TOKEN}
+        onClick={() => setMapstyle(MAP_STYLE_MAPBOX_STREETS)}
+      >
+        {t('Streets')}
+      </button>
+      <button
+        className={classnames('btn btn-default', {
+          'btn-primary': mapStyle === MAP_STYLE_MAPBOX_SATELLITE,
+        })}
+        disabled={!MAPBOX_TOKEN}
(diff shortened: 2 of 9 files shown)


modules/core/client/components/Map/MapIcon.js
```diff
@@ -0,0 +1,34 @@
...
+ * @link https://docs.mapbox.com/api/maps/#static-images
+ */
+export default function MapIcon({ mapboxStyle, mapboxToken }) {
+  const image =
+    mapboxStyle && mapboxToken
+      ? `https://api.mapbox.com/styles/v1/mapbox/${mapboxStyle
+          .split('/')
+          .pop()}/static/14.1663,55.6438,5.22,0/128x128@2x?attribution=false&access_token=${mapboxToken}`
+      : // Default image when API isn't available
+        'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNc8mTSfwAHaQMbL4UQfQAAAABJRU5ErkJggg==';
```
simison, line 17: This could actually just print out nothing in this case but I wanted to keep UI as consistent as possible even when Mapbox isn't configured.

modules/core/client/components/Map/MapStyleControl.js
```diff
@@ -0,0 +1,80 @@
...
+import MapIcon from './MapIcon';
+
+export default function MapStyleControl({ mapStyle, setMapstyle }) {
+  const { t } = useTranslation('core');
+  const mapboxToken = getMapboxToken();
+
+  const mapboxStyleNames = {};
+  mapboxStyleNames[MAP_STYLE_MAPBOX_STREETS] = t('Streets');
+  mapboxStyleNames[MAP_STYLE_MAPBOX_SATELLITE] = t('Satellite');
+  mapboxStyleNames[MAP_STYLE_MAPBOX_OUTDOORS] = t('Outdoors');
```
simison, line 24: Better ideas how to map labels to map styles welcome 😁

modules/core/client/components/Map/MapStyleControl.js
```diff
@@ -0,0 +1,80 @@
...
+          </MenuItem>
+        ))}
+        {process.env.NODE_ENV !== 'production' && (
+          <MenuItem
+            active={selectedStyle === MAP_STYLE_OSM.name}
+            key={MAP_STYLE_OSM.name}
+            onClick={() => setMapstyle(MAP_STYLE_OSM)}
+          >
+            <MapIcon />
+            {MAP_STYLE_OSM.name}
```
simison, line 69: Here I'm pulling the name directly from OSM json definition, but really it could just be statically here too.

modules/core/client/components/Map/MapStyleControl.js
```diff
@@ -0,0 +1,80 @@
...
+  const { t } = useTranslation('core');
+  const mapboxToken = getMapboxToken();
+
+  const mapboxStyleNames = {};
+  mapboxStyleNames[MAP_STYLE_MAPBOX_STREETS] = t('Streets');
+  mapboxStyleNames[MAP_STYLE_MAPBOX_SATELLITE] = t('Satellite');
+  mapboxStyleNames[MAP_STYLE_MAPBOX_OUTDOORS] = t('Outdoors');
+
+  // If it's an object, it'll have a name. Otherwise it's Mapbox URL in string presentation.
+  const selectedStyle = mapStyle?.name || mapStyle;
```
simison, line 27: This was important so that for OSM I wouldn't be comparing to `[object object]`.

OSM is a Json object, while other styles are just strings with `mapbox://blabla`.

We might in future have some other json objects too  so I felt worthwhile be little prepared.

I also like to keep OSM layer around as a backup we can quickly switch on even in production.

modules/core/client/components/Map/constants.js
```diff
@@ -0,0 +1,12 @@
+import osmStyle from './osm.json';
+
+export const getMapboxToken = () => window.settings?.mapbox?.publicKey;
```
simison, line 3: `const MAPBOX_TOKEN = window.settings?.mapbox?.publicKey;` gets back as undefined when imported. 🤷

I don't want to include in bundles either via Webpack. Not sure if this file is best for `getMapboxToken()` tho or should I move it to `utils.js`?

modules/core/client/components/Map/osm.json
```diff
@@ -1,6 +1,6 @@
 {
   "version": 8,
-  "name": "OpenStreetMap",
+  "name": "OSM",
```
simison, line 3: Shorter just fit better in the button without going to multiple lines. 😁 it's for internal use only anyway.

nicksellen:
Looks nice!

I had some issues with the mapbox config though: without anything special set I could see only the *OSM* option, the others are visible, but disabled, and no icons:

mapstyles

... with the default mapbox config from `config/env/local.sample.js` I get:

mapboxtoken

So, I guess invalid token, but also it doesn't show the mapbox logo in the corner.

Also, I wonder about the position, it looks a little awkward to me floating away from the corner (even when it has the mapbox logo), maybe top right would look neater?, or tucked next to the zoom controls?

I guess I'll wait till I have a working mapbox config to go through the rest.

simison:
@nicksellen I added functioning key in https://github.com/Trustroots/trustroots/pull/1333/commits/1897003905276631b4863a08e545597c7a394222 — feel free to test with that. Otherwise I'll go ahead and merge in a few days time. :-)

simison:
>Also, I wonder about the position, it looks a little awkward to me floating away from the corner (even when it has the mapbox logo), maybe top right would look neater?, or tucked next to the zoom controls?

I agree — I didn't concentrate too much on that this time around since we likely won't need the style switch in these small maps. The big map has different design constraints so we can return to that question once that progresses little further.

modules/core/client/components/Map/MapStyleControl.js
```diff
@@ -0,0 +1,80 @@
...
+            <MapIcon
+              mapboxStyle={
+                selectedStyle !== MAP_STYLE_OSM.name ? selectedStyle : false
+              }
+              mapboxToken={mapboxToken}
+            />
+            {mapStyle?.name || mapboxStyleNames[mapStyle]}
+          </>
+        }
+        id="map-style-picker"
```
simison, line 45: This will cause issues with multiple maps on one page since we'll have non-unique IDs, but it's a shortcoming of this version of ReactBootstrap. I'd leave for now and address in follow-ups. It doesn't break, just sub-optimal.

modules/core/client/components/Map/constants.js
nicksellen: I think its fine for now. There are some other ideas floating around for a more well defined settings structure.

PR #1343, opened 2020-04-01
Highlight unread messages #1311

### LINKED ISSUE #1311: Highlight unread messages
Looking at the inbox at https://www.trustroots.org/messages, all messages look the same, no matter if they are read or unread.

Unread messages should be highlighted (e.g. bold & black instead of grey).

aldenfbrown:
## Proposed Changes

*  Added conditional that applies styling via a preexisting className

#### Testing Instructions

* Go to the inbox verify the unread messages appear bolded

Fixes #1311

### DIFF
--- a/modules/messages/client/components/InboxThread.js
+++ b/modules/messages/client/components/InboxThread.js
@@ -47,7 +47,7 @@ export default function InboxThread({ user, thread }) {
 InboxThread.propTypes = {
   user: userType.isRequired,
   thread: PropTypes.shape({
-    read: PropTypes.string.isRequired,
+    read: PropTypes.boolean.isRequired,
     updated: PropTypes.string.isRequired,
     userFrom: userType.isRequired,
     userTo: userType.isRequired,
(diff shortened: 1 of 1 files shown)


rumwerfer:
Works as expected.

modules/messages/client/components/InboxThread.js
```diff
@@ -10,8 +10,14 @@ export default function InboxThread({ user, thread }) {
   const { t } = useTranslation('messages');
   const otherUser = findOtherUser(user, thread);
   const haveReplied = thread.userFrom._id === user._id;
+  const { read } = thread;
+
   return (
-    <li className="list-group-item threadlist-thread">
+    <li
+      className={`list-group-item threadlist-thread ${
+        !read ? 'threadlist-thread-unread' : ''
```
simison, line 18: We can use `classnames` package here to simplify; we use it elsewhere in codebase too.

simison:
Can you have a look at the failure in CI?

modules/messages/client/components/InboxThread.js
```diff
@@ -41,6 +47,7 @@ export default function InboxThread({ user, thread }) {
 InboxThread.propTypes = {
   user: userType.isRequired,
   thread: PropTypes.shape({
+    read: PropTypes.boolean.isRequired,
```
simison, line 50: I think it's `bool`, not `boolean`

aldenfbrown: Ahh that's right silly me I have been in TypeScript land too long. Hang tight.

aldenfbrown: Ok now we are good. Thanks!

PR #1364, opened 2020-04-05
Add Tooltip component

mrkvon:
#### Proposed Changes

* create `Tooltip`, a reusable component that adds tooltip to the wrapped component

```jsx
<Tooltip tooltip="text"><WrappedComponent></Tooltip>
```

##### Usage

* Read `modules/core/client/components/README.md`

#### Testing Instructions

* `git switch with-tooltip`
* test tooltip in tribes `JoinButton`:
  - log in
  - go to `/tribes`
  - see that the tooltip under Joined button works
    image

* test tooltip in offers
  - log in
  - go to my profile
  - see that the tooltip next to Accommodation edit button works
    image

Preparation for #1350


simison:
Thanks for thinking about this! 

I find this API a little odd but I guess that's bound to happen with modal like components. :-)

I would expect something more component like where we'd have `<Tooltip>` either as wrapper for something or we'd pass `React.createRef()` for it to attach it to target DOM.

Could we alternatively use what we already have available in `react-bootstrap`? https://react-bootstrap-v3.netlify.com/components/tooltips/

mrkvon:
@simison 
Actually, we're using the react-bootstrap `Tooltip`. But since it has a rather flexible but complicated API, I decided to make this wrapper. 
However, I like the <Tooltip><Something /></Tooltip> syntax more than HOC, so I'll refactor. I don't know how to create the tooltip attached via `ref`, so I'll leave that for later or for somebody who knows.
Thanks for the feedback!

mrkvon:
Gave it another try with a different API: `<Tooltip tooltip="text" placement="top" {...otherProps}><Element /></Tooltip>`

One way or the other, it's really just a wrapper around the react-bootstrap's [`Tooltip`](https://react-bootstrap-v3.netlify.com/components/tooltips/). It implements a common use case, i.e. _hover and see a tooltip_.

Please have a look and share your opinions! :bowing_man:

public/locales/en/offers.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Accommodation": "Accomodation",
+  "Accommodation": "Accommodation",
```
mrkvon, line 2: This is an unrelated fix. Please bear with me! :smiley_cat:

modules/offers/client/components/OffersPresentational.js
```diff
@@ -54,48 +49,49 @@ export class OffersPresentational extends Component {
...
     return (
       <ButtonGroup className="pull-right dropdown-menu-offers">
-        <DropdownButton
-          pullRight
-          className={`btn-offer-hosting, btn-offer-hosting-${offer.status}`}
-          bsSize="small"
-          bsStyle="success"
-          title={this.hostingStatusLabel(offer.status)}
-          id={'dropdown-offers-button'}
-          overlay={tooltip}
```
mrkvon, line 75: this api simply didn't work before

mrkvon:
Hiding whitespaces can be useful for reviewing this one:

image

or `git diff master -w` if you're reviewing locally

mrkvon:
@nicksellen, @simison Thank you for your reviews and feedback!

PR #1365, opened 2020-04-05
Refactor Navigation component

mrkvon:
#### Proposed Changes

- move `Navigation` component from `references` to `core`
- add navigation for small screens
- block buttons based on provided (validation) errors
- add tooltip with error message to blocked buttons
- use in `CreateReference` component

#### Testing

- `git switch refactor-navigation`
- make sure you have references enabled: `{ featureFlags: { reference: true } }` in `/config/env/development.js`
- go to `/profile/{some-username}/references/new`
- see that navigation works on a big screen
  image
- see that navigation works on a small screen
  image
- hover disabled buttons and see validation errors in tooltip
- see whether `modules/core/client/components/README.md` makes sense

~Depends on #1364, which should be merged first~
Preparation for #1350

### DIFF
--- a/modules/core/client/components/StepNavigation.js
+++ b/modules/core/client/components/StepNavigation.js
@@ -71,11 +71,11 @@ const SubmitButton = ({ small, ...props }) => {
 SubmitButton.propTypes = { small: PropTypes.bool };
 
 /**
- * Navigation is a react component.
+ * StepNavigation is a react component.
  * It can contain three different buttons: Back, Next, Submit.
  * Each of them has a related property onBack, onNext, onSubmit
  */
-export default function Navigation({
+export default function StepNavigation({
   disabled,
   tab,
   tabs,
@@ -84,10 +84,14 @@ export default function Navigation({
   onNext,
   onSubmit,
 }) {
-  const errorTab = errors.findIndex(errors_ => errors_.length > 0);
+  const errorTab = errors.findIndex(tabErrors => tabErrors.length > 0);
   const tabDone = errorTab === -1 ? tabs : errorTab - 1;
-  // get the lowest error valid for the current tab
-  const error = errors.slice(0, tab + 1).flat()[0] ?? '';
+  // get the lowest non-empty error valid for the current tab
+  const error =
+    errors
+      .slice(0, tab + 1)
+      .flat()
+      .find(error => error.length > 0) ?? '';
 
   /**
    * We'll reuse the buttons and tooltip defined below
@@ -160,7 +164,7 @@ export default function Navigation({
   );
 }
 
-Navigation.propTypes = {
+StepNavigation.propTypes = {
   onBack: PropTypes.func.isRequired,
   onNext: PropTypes.func.isRequired,
   onSubmit: PropTypes.func.isRequired,
--- a/modules/core/client/components/StepNavigation.js
+++ b/modules/core/client/components/StepNavigation.js
@@ -2,8 +2,24 @@ import React, { cloneElement } from 'react';
 import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
 import classnames from 'classnames';
-
 import Tooltip from './Tooltip';
+
+/**
+ * StepNavigation can contain three different buttons: Back, Next, Submit we can go through.
+ * Each of them has a related property onBack, onNext, onSubmit.
+ * See it at `/offer/host` or `/profile/{username}/references/new`.
+ * This component doesn't keep its own state.
+ *
+ * @param {number} currentStep - index of current step
+ *                               first step doesn't have Back button
+ *                               last step has Finish/Submit button instead of Next
+ * @param {number} numberOfSteps - amount of steps in total
+ * @param {boolean} [disabled] - disable Next button
+ * @param {string} [disabledReason] - message to show in tooltip of disabled button
+ * @param {function} onBack - method executed on clicking Back
+ * @param {function} onNext - method executed on clicking Next
+ * @param {function} onSubmit - method executed on clicking Submit
+ */
 
 const BackButton = ({ small, ...props }) => {
   const { t } = useTranslation('core');
@@ -70,29 +86,15 @@ const SubmitButton = ({ small, ...props }) => {
 };
 SubmitButton.propTypes = { small: PropTypes.bool };
 
-/**
- * StepNavigation is a react component.
- * It can contain three different buttons: Back, Next, Submit.
- * Each of them has a related property onBack, onNext, onSubmit
- */
 export default function StepNavigation({
-  disabled,
-  tab,
-  tabs,
-  errors,
+  currentStep,
+  numberOfSteps,
+  disabled = false,
+  disabledReason = '',
   onBack,
   onNext,
   onSubmit,
 }) {
-  const errorTab = errors.findIndex(tabErrors => tabErrors.length > 0);
-  const tabDone = errorTab === -1 ? tabs : errorTab - 1;
-  // get the lowest non-empty error valid for the current tab
-  const error =
-    errors
-      .slice(0, tab + 1)
-      .flat()
-      .find(error => error.length > 0) ?? '';
-
   /**
    * We'll reuse the buttons and tooltip defined below
    * either with different children or with additional props.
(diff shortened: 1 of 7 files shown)


modules/core/client/components/README.md
```diff
@@ -1,5 +1,36 @@
...
+### Usage
+
+```jsx
+import Navigation from `@/modules/core/client/components/Navigation`;
+
+// ...
+<Navigation
+  tab={currentTabNumber}
+  tabs={totalTabNumber}
+  errors={arrayOfArraysOfErrorsForEachTab}
```
simison, line 23: Seems little confusing. Can we simplify this?

mrkvon: > Seems little confusing. Can we simplify this?

@simison, do you mean simplify the variable name or API?

On side note: This is a typical scenario where I :heart: Typescript. Not that I'm ready to start migrating Trustroots to Typescript any time soon, or ever. :laughing:

modules/core/client/components/README.md
```diff
@@ -1,5 +1,36 @@
 # Reusable components
 
+## Navigation
```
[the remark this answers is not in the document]
mrkvon: Good point! What was fine in `references` module suddenly becomes way too generic in `core`. I'll go with `StepNavigation`, because when I hear `Wizard`, I imagine Gandalf or Harry Potter.

modules/references/client/components/CreateReference.component.js
```diff
@@ -133,11 +150,11 @@ export default function CreateReference({ userFrom, userTo }) {
           {tabs[1]}
         </Tab>
       </Tabs>
-      {/* <!-- Navigation for big screens -->*/}
+      {/* <!-- Navigation -->*/}
```
simison, line 153: Fine to remove `<!--` since this is a JS comment now.

Actually, fine to remove altogether since it doesn't tell much anymore. :-)

modules/core/client/components/README.md
simison: API. :-) Not a blocker, just I felt like doing some brain-gymnastics when trying to grok how to use this component and "Ask @mrkvon if you need explanation." comment was a flag towards some complexity too. 😁 

Not saying these are better but just wanted to share for inspiration:
- https://wpcalypso.wordpress.com/devdocs/design/wizard
- https://wpcalypso.wordpress.com/devdocs/design/wizard-progress-bar

modules/core/client/components/README.md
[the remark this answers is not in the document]
simison: <h1>🧙‍♂️</h1>

modules/core/client/components/README.md
mrkvon: > API

Fair enough. Thank you for the examples! I get some inspiration how to at least name the props more clearly...

> Ask @mrkvon if you need explanation.

Lol that really is a flag... :grin: If one is too lazy or tired to write the API docs, one should consider changing the API...

Currently I'm thinking about:

- `currentStep` instead of `tab`
- `numberOfSteps` instead of `tabs`
- leave `onBack`, `onNext`, `onSubmit` as it is

I still don't know what to do about `disabled` and `errors`. They're both meant to block the `Next` or `Finish` buttons.
- `errors` - when `numberOfSteps={3}`, `[[], ['error', 'error2'], []]` would disable progress from second and third tab, and show text `error` in the tooltip above the disabled button. Meant primarily for form validation.
- `disabled` serves to disable `Finish` button e.g. when a form gets saved to API. Alternatively it could be replaced with `errors=[[], [], ['saving']]`, so it could be completely removed.

I can imagine removing the `disabled` and renaming the `errors` to something like `progressErrors`, but I'm not sure about a simpler data structure.
Well, now that I think about it, maybe we can pass just the single progress error applicable to the current step.
e.g. instead of both `errors` and `disabled` we can have only one of the following:
```js
progressError: PropTypes.string // not required
```
or
```js
disabled: PropTypes.oneOf(PropTypes.string, PropTypes.boolean) // not required
```


Do you have ideas about the above or how to further simplify the API; or change it completely?

modules/core/client/components/README.md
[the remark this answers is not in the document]
mrkvon: @simison, I haven't realized `Wizard` is a common term. I've done the renaming to `StepNavigation` in a25fa4fc396a02ee0d61e3c918a748331df6affd, but if you strongly prefer `WizardNavigation`, let me know. It doesn't matter to me.

modules/references/client/components/CreateReference.component.js
mrkvon: :ok_hand: a15884f55282bf4570726f50d957a9712b8d53d8

modules/core/client/components/README.md
simison: Would using `disabled` props (for all and/or each button separately) simplify it? That way consumer could just `disabledNext={ isError }` etc.

mrkvon: > Would using disabled props (for all and/or each button separately) simplify it?

Yes, I think so. It sounds like a reasonable direction to simplify this.
Just `disabled` would do it, since we only disable forward buttons, ever.
However we need to pass the message for the tooltip, too. So I wonder whether to name it `disabled` (which i'd expect to be boolean, not string) or something arbitrary like `progressError`. Or pass both.
WDYT?

simison: I've started to think we should use less tooltips, since they're hidden info and if the message is important, it should just outright be visible.

`disabledReason` seems like good optional prop otherwise?

modules/core/client/components/README.md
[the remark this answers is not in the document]
simison: `Step*` sounds great and clear 👌

modules/core/client/components/README.md
mrkvon: @simison That makes sense. I'll be off the screen now and will implement later. Thank you for the guidance.

mrkvon: Changed in 711113df9d86af557ddb5dd164d9777f2370ff1e

modules/core/client/components/StepNavigation.js
```diff
@@ -0,0 +1,172 @@
...
+  disabledReason = '',
+  onBack,
+  onNext,
+  onSubmit,
+}) {
+  /**
+   * We'll reuse the buttons and tooltip defined below
+   * either with different children or with additional props.
+   * We use React.cloneElement(element, props, children) for that purpose:
+   * https://reactjs.org/docs/react-api.html#cloneelement
```
nicksellen, line 102: Is the `cloneElement` really needed? Caused me to look twice, wondering what it is for.

My guess would be the intention is to improve performance by not re-creating buttons all the time, but I don't think this would be very heavy, as the button components are not complex.

Additionally, the "source" buttons (to be cloned) are being created on each render anyway (as it's a function component), so I think it is not saving much.

Or perhaps there is a reason I'm missing?

modules/core/client/utils/validation.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+/**
+ * provided a dictionary of rules and a dictionary of values,
+ * return a dictionary of errors for each value
+ *
+ * @param {Object.<string, rule[]>} ruleDict - dictionary of array of rules
+ * @param {Object.<string, any>} valueDict - dictionary of values
+ * @returns {Object.<string, any[]>} - dictionary of array of errors
+ */
+export const validate = (ruleDict, valueDict) => {
```
nicksellen, line 15: I can imagine we might have much more extensive use for declarative clientside validation at some point, and it might be worth using a third party library for it.

Considered in that scope, then this seems fine, otherwise I might propose a structure that lets us define the rules seperate to the validation function, something like:

```js
const validate = createValidator(ruleDict);
const errors = validate(valueDict);
```

modules/core/client/components/StepNavigation.js
mrkvon: The only reason was that the code looked less repetitive to me. With `cloneElement`, I was able to pass the props and children to buttons and Tooltip once, and then only tweak a little bit.
The alternative was to assign the props to variables and pass them to buttons and tooltip each time, which felt more repetitive, but maybe it's clearer?
Or maybe there is an alternative, like creating the navigation only once and styling it with media queries.
Does this make sense?

mrkvon:
@nicksellen Thank you for the review!

This PR depends on #1364, so that one needs to be merged first.

modules/core/client/components/StepNavigation.js
mrkvon: changed in b8238c24c77407e89f24ab2aff5d25a9c4996c3b

it's clearer without cloneElement

modules/core/client/utils/validation.js
mrkvon: 2eb41973dc95f1b9dc6fd600f797f8f9b6d6669f
Made `createValidator` function. I wouldn't like to decide about the final version of client-side validation now. Let's leave it undecided until we're more sure what we need.

mrkvon:
@nicksellen and @simison, thank you for your reviews, both here and #1364.

Issue #1368, opened 2020-04-07
Provide namespace to Trans by passing t function


mrkvon:
When the issue gilbsgilbs/babel-plugin-i18next-extract#153 gets resolved, let's change all occurrences of `<Trans ns="namespace">` to `<Trans t={t}>`.

Currently, providing `t` to `Trans` works in react-i18next. However, with `babel-plugin-i18next-extract` the strings are extracted to a wrong (default) namespace, i.e. to `public/locales/en/translation.json` instead of `public/locales/en/namespace.json`.

PR #1369, opened 2020-04-07
Name i18next namespaces by modules

mrkvon:
#### Proposed Changes

To have a consistent naming of i18next namespaces, let's use a namespace according to the module in which the component in question lives.
E.g. `modules/users/client/components/Avatar.component.js` should have `useTranslation('users')` instead of something arbitrary.

Additionally, we've also renamed the translation files according to the new convention, and extracted translations by running `npm run build`. (These two commits should be merged separately, not squashed.)

#### Testing Instructions

* Maybe see that the app and translations still work as expected. I can't think of anything else to do.
* If you don't like that translation extraction is a part of this PR, let me know in review, I'll revert the second commit (ddde13a6577fd6f600b9ff87092a8a08583767e4).
* When reviewing, you can probably ignore `public/locales`. Or just focus on ddd2bfde4b8dc6fb18d8b3e9583ca6ca454b3fbd only.

#### Curious

Shall weblate pick the changes in translation files from `master` automatically? :crossed_fingers: (Please _do not_ pull to weblate manually so we can debug the process.)


docs/i18n.md
```diff
@@ -24,25 +26,28 @@ Don't use class components, use [hooks](https://reactjs.org/docs/hooks-overview.
...
 export function MyComponent() {
-  const { t } = useTranslation('myNamespace');
+  const { t } = useTranslation('currentModule');
 
   return (
     <>
       <div>{t('I am a text to translate!')}</div>
       <div>
-        <Trans>
+        <Trans ns="currentModule" t={t}>
```
nicksellen, line 50: We only need `t={t}` and not `ns="currentModule"` I think, from reading https://react.i18next.com/latest/trans-component`

modules/contacts/client/components/ContactPresentational.js
```diff
@@ -25,7 +25,7 @@ export default function ContactPresentational({
         <div>
           <i className="icon-fw icon-building text-muted"></i>
           <small>
-            <Trans ns="contact" locationLiving={locationLiving}>
+            <Trans ns="contacts" locationLiving={locationLiving}>
```
nicksellen, line 28: Passing in `t` seems neater than re-specifying the namespace.

mrkvon: I see. I agree with that.
There is what i think is a bug in babel-plugin-i18next-extract. It doesn't pick up the proper namespace from the passed `t` (at least for me). The translation strings from `Trans` without explicit `ns` end up in `translation.json` instead of their proper `namespace.json`. I've opened gilbsgilbs/babel-plugin-i18next-extract#153 and #1368.
Should we document the reasons for using `ns` instead of `t` more clearly? Add TODOs everywhere? Add `t={t}`?
Alternatively, it's possible that I'm overlooking something and the plugin works as it should.

mrkvon: Added `t={t}` and TODO about removing `ns` to all `Trans` components. (squashed into the first commit)
I think that's all we can do with this issue now.

docs/i18n.md
mrkvon: Added both `ns` and `t` to every component, with TODO about removing `ns` when the bug in babel-plugin-i18next-extract library is removed. #1368 
See [below](https://github.com/Trustroots/trustroots/pull/1369#discussion_r406729506) for further discussion.

PR #1371, opened 2020-04-10
Translations update from Weblate

weblate:
Translations update from [Weblate](https://hosted.weblate.org/projects/trustroots/translation/) for Trustroots/translation.

PR #1372, opened 2020-04-11
Add new languages

mrkvon:
#### Proposed Changes

* add a few new languages to `./config/shared/locales.js`
* extract translations

I use https://jsfiddle.net/sm9wgLze to decide on language names and whether they are uppercase or lowercase.

#### Testing Instructions

- please let me know about any languages you know of that people may wish to translate.
- to review, you can ignore everything in `public/locales`. Those are extracted by machine.


config/shared/locales.json
```diff
@@ -1,6 +1,12 @@
 [
   { "code": "en", "label": "English" },
-  { "code": "cs", "label": "česky" },
+  { "code": "de", "label": "Deutsch" },
+  { "code": "es", "label": "español, castellano" },
```
mrkvon, line 4: Which one to keep? Or both? We're trying to have the language name in its own language.

@Akronix what do you think?

mrkvon:
@nicksellen, thank you for the review. I'll be interested to discuss this and to find out what exactly you mean.

I don't know how this piece should work, yet. I do find it frustrating that with current settings adding languages via weblate UI is not integrated and we need to do it through that config file.

For example, see #1371 - Bokmål was added via weblate interface, but doesn't help with adding it to app.

I'd also like to see this part improved. Indeed I don't know now.

config/shared/locales.json
simison: To avoid disputes and diving into language politics (usually impossible to resolve and there are _many_ as you'll surely learn in coming months), I just usually every project refer to whatever Wikipedia uses for their language menus.

They use just español so I'd go with that here too.

config/shared/locales.json
```diff
@@ -1,6 +1,12 @@
...
+  { "code": "fr", "label": "français" },
+  { "code": "it", "label": "Italiano" },
+  { "code": "nb", "label": "Norsk bokmål" },
+  { "code": "pl", "label": "polski" },
   { "code": "fi", "label": "suomi" },
```
simison, line 9: I think we should remove Finnish until there is someone to actually translate it. I'm not gonna have time for it. :-)

config/shared/locales.json
Akronix: Personally, I'd prefer "Castellano". But @simison has an interesting point and it's better to have a well-founded rule than relying on personal preferences. So I'd go for Español, also I feel it's more familiar for international people.

simison: I know the feeling, I also have preferences for certain ways to express some languages. That said, I've found it easier to be fair and consistent by relying on external well defined list that is relatively "neutral". UN list of languages is another good one.

PR #1375, opened 2020-04-12
Delete yarn.lock

simison:
Removing `yarn.lock` — we also have vanilla `package-lock.json` in the repo. 😁 

@nicksellen I see you added this in https://github.com/Trustroots/trustroots/pull/1173  — was that accidental?


nicksellen:
Doh, seems I did. Funny though too because as around the same time as that merged, I fixed up this other PR that had a yarn.lock in it :) https://github.com/Trustroots/trustroots/pull/1045/commits/9ea1a39c8cfef877beb1374e6e930d4384fa2de6

PR #1384, opened 2020-04-14
References work (hackweek) - team 2

nicksellen:
#### Proposed Changes

* added feedbackPublic field...
* maybe some other stuff...

#### Testing Instructions

*

Fixes #

### DIFF
--- a/modules/references/client/components/create-reference/Feedback.js
+++ b/modules/references/client/components/create-reference/Feedback.js
@@ -1,0 +1,36 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+export default function Feedback({ feedback, onChangeFeedback }) {
+  const { t } = useTranslation('references');
+
+  return (
+    <div className="panel panel-default">
+      <div className="panel-heading">
+        <h4>optionally write some feedback for them</h4>
+      </div>
+      <div className="panel-body">
+        <label htmlFor="feedback-message" className="control-label">
+          {t('Public feedback')}
+        </label>
+        <textarea
+          className="form-control input-lg"
+          rows="7"
+          id="feedback-message"
+          onChange={event => onChangeFeedback(event.target.value)}
+          value={feedback}
+        ></textarea>
+        <span className="help-block">
+          {t('some nice help text')}
+          <br />
+        </span>
+      </div>
+    </div>
+  );
+}
+
+Feedback.propTypes = {
+  feedback: PropTypes.string.isRequired,
+  onChangeFeedback: PropTypes.func.isRequired,
+};
--- a/modules/references/server/controllers/reference.server.controller.js
+++ b/modules/references/server/controllers/reference.server.controller.js
@@ -2,6 +2,7 @@ const mongoose = require('mongoose');
 const _ = require('lodash');
 const path = require('path');
 const util = require('util');
+const config = require(path.resolve('./config/config'));
 const errorService = require(path.resolve(
   './modules/core/server/services/error.server.service',
 ));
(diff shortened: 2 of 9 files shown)


simison:
In this PR there might be some CSS/HTML that you can re-use for free-text part: https://github.com/Trustroots/trustroots/pull/649

There's also HTML/UI work for +1ing people for different tribes. 😁

nicksellen:
Thanks! Also intending to pinch some stuff from https://github.com/Trustroots/trustroots/pull/918 where relevant.

nicksellen:
I split the first bit of work we did off into it's own branch, so it can be reviewed and merged without waiting for the rest of the stuff we're doing ---> https://github.com/Trustroots/trustroots/pull/1385

nicksellen:
Now added initial work on 3rd tab to write the (optional) feedback, without proper labels/hint/title or anything though. A few questions came up for me:
- should it actually be on a 3rd tab
- should it have a checkbox you have to select before getting the text field (to make it a positive action to decide to write some feedback)
- should it be called feedback in the UI
- should you be able to write a feedback whichever options you selected in the previous steps (e.g. we could make it only possible to write feedback if you selected "yes" to the previous question)

simison:
>  should it actually be on a 3rd tab

I think that would make sense so that folks can concentrate on one action at once. Ideally we'd carry on selections done until then as a summary but we can improve this later on. Separate steps also make it easier to build good UI for mobile as each step is more contained.


 > Should it have a checkbox you have to select before getting the text field (to make it a positive action to decide to write some feedback)

This could make sense, yeah. We can keep simple for starters and add that once we get to test the whole flow and if it feels like it needs it?

 > Should it be called feedback in the UI

Would be important to think about copy since it'll direct the nature of this content quite a bit. We can always add extra guidance but even that should be very concise and to the point. I wouldn't hold merging pr tho and change copy separately. You might wanna keep translation functions out (or renamed with `as`) until we have final form.

 > should you be able to write a feedback whichever options you selected in the previous steps (e.g. we could make it only possible to write feedback if you selected "yes" to the previous question)

My gutfeeling tells it's really most insightful for other than "yes" but by leaving it open we can observe how people actually start using it. What made you think it could be available only when answering "yes"?

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -72,8 +73,23 @@ function validateCreate(req) {
     details.userTo = 'userId expected';
   }
 
+  if (_.has(req, ['body', 'feedbackPublic'])) {
+    const feedbackPublic = req.body.feedbackPublic;
```
simison, line 77: What kind of sanitation we should add here?

modules/references/client/components/create-reference/Feedback.js
```diff
@@ -0,0 +1,36 @@
...
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+export default function Feedback({ feedback, onChangeFeedback }) {
+  const { t } = useTranslation('references');
+
+  return (
+    <div className="panel panel-default">
+      <div className="panel-heading">
+        <h4>optionally write some feedback for them</h4>
```
simison, line 11: Needs `t()`

modules/references/server/controllers/reference.server.controller.js
nicksellen: See https://github.com/Trustroots/trustroots/pull/1385/files#r409637173

nicksellen: (I split the server change off into a seperate branch/PR)

nicksellen:
I think this would do best to be recreated with just the UI part, in a fresh branch/PR, now https://github.com/Trustroots/trustroots/pull/1385 is merged.

nicksellen:
Closed in favour of https://github.com/Trustroots/trustroots/pull/1413 - which is a cleaned up, tests added, branched from fresh master version of the same work.

PR #1385, opened 2020-04-14
References public feedback

nicksellen:
#### Proposed Changes

* add `feedbackPublic` field for references
* make `recommend` field private

#### Testing Instructions

Fixes #


modules/references/tests/server/reference-create.server.routes.tests.js
```diff
@@ -72,6 +74,7 @@ describe('Create a reference', () => {
                 hostedThem: true,
               },
               recommend: 'yes',
```
edvardmk, line 76: should this still be included?

nicksellen: Yes, as the user will still submit that info, it's just that we won't return it.

modules/references/tests/server/reference-create.server.routes.tests.js
```diff
@@ -605,6 +608,28 @@ describe('Create a reference', () => {
         });
       });
 
+      it.only('[too long public feedback] 400', async () => {
```
mrkvon, line 611: remove `.only`

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -72,8 +73,23 @@ function validateCreate(req) {
     details.userTo = 'userId expected';
   }
 
+  if (_.has(req, ['body', 'feedbackPublic'])) {
+    const { feedbackPublic } = req.body;
```
nicksellen, line 77: Comment from @simison in https://github.com/Trustroots/trustroots/pull/1384#discussion_r409512920

> What kind of sanitation we should add here?

I borrowed the client side part from the report message, which is plaintext and uses `textService.plainText`, so seems a good option here too.

nicksellen:
Thanks for review!

PR #1386, opened 2020-04-14
Update react bootstrap

rumwerfer:
#### Proposed Changes

* `npm install --save react-bootstrap@latest` : 0.33.1 -> 1.0.0
* migrate files using react-bootstrap according to https://react-bootstrap.netlify.com/migrating/#summary-of-breaking-changes-from-v0320 

#### Testing Instructions

*

### DIFF
--- a/modules/core/client/components/LanguageSwitchPresentational.js
+++ b/modules/core/client/components/LanguageSwitchPresentational.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Dropdown, MenuItem } from 'react-bootstrap';
+import { Dropdown } from 'react-bootstrap';
 import locales from '@/config/shared/locales';
 
 /**
@@ -31,13 +31,13 @@ export function LanguageSwitchDropdown({
       </Dropdown.Toggle>
       <Dropdown.Menu>
         {locales.map(({ code, label }) => (
-          <MenuItem
+          <Dropdown.Item
             key={code}
             onClick={() => onChangeLanguage(code)}
             active={code === currentLanguageCode}
           >
             {label}
-          </MenuItem>
+          </Dropdown.Item>
         ))}
       </Dropdown.Menu>
     </Dropdown>
--- a/modules/offers/client/components/OffersPresentational.js
+++ b/modules/offers/client/components/OffersPresentational.js
@@ -1,5 +1,10 @@
 // External dependencies
-import { Button, DropdownButton, MenuItem, ButtonGroup } from 'react-bootstrap';
+import {
+  Button,
+  DropdownButton,
+  DropdownItem,
+  ButtonGroup,
+} from 'react-bootstrap';
 import { withTranslation } from 'react-i18next';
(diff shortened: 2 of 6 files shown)


rumwerfer:
Apparently, I missed some parts. I only went through files in /modules though.
I don't quite understand the output from Travis; would be glad if someone could go through it with me.

nicksellen:
I think the bootstrap upgrade is going to be a _lot_ more involved, as in addition to the breaking changes for `react-bootstrap` itself, it also says to go through the bootstrap migration guide itself: https://getbootstrap.com/docs/4.3/migration/

That includes various stuff, e.g. _Switched from Less to Sass for our source CSS files._ - the way react-bootstrap works is that they don't provide the CSS themselves, they rely on us providing, and we currently do that using the bootstrap less files, but we also use lots of the bootstrap less functions in our own custom CSS (less), so there could be some major changes when upgrading that too.

The style is coming from the `bootstrap` dependency in `package.json`.

The errors in CI should be the same as local errors when running the tests, which is true for me (I run the tests with `npx jest`), and there is quite a lot of broken stuff there, a few things:
- the test matchers find elements by their aria role, and it looks like some things have changed from menus to buttons, so role `menuitem` -> `button` should fix those.
- some invalid props, `bsSize` and `pullRight` that it doesn't know about, I guess they changed?

I think it would be useful to upgrade at some point, but for the given purpose ("neutralizing" the green/red buttons, there will be a way using the existing versions :) - it might be the docs for the older version are still online).

rumwerfer:
Thank you for the detailed explanation. It actually says very clear, "PLEASE FIRST READ THE UPSTREAM BOOSTRAP MIGRATION GUIDE", but I unfortunately didn't start reading before the link Abel gave me. I don't think I'm the right one to do the bootstrap migration then.

simison:
Yep, this is going to be a pretty involved chunk of work and we need to think a bit how to approach it.

I regret little bit not taking the opportunity with React migration to just switch something else and phase out Bootstrap v3 together with old Angular files. At the same time, current approach (using React-Bootstrap) has the benefit of making React-migration easier since it won't involve any re-designing, and old CSS can mostly be re-used. Switching to something "incompatible" with Bootstrap also makes it easier to run the two systems in parallel easier, since they wouldn't collide too much.

My personal preference would be to phase out Bootstrap entirely and switch to a more comprehensive framework that also translates nicely to native mobile, has strong accessibility built-in, covers more components, has a stronger type system, has theming built-in, etc. By more comprehensive I also mean that I've been missing having _a design-system_ that guides how to build sections, not just having a bunch of components available. That would make us depend less on having polished designs available since everything built according to the system would look more or less good and balanced with less effort (flows/mechanics aside, some design still needed of course).

The gold standard for all that is [Material](https://material.io/) so I'd suggest we start at some point plan how to include it in the build pipeline, how to theme it for Trustroots and how to eventually switch everything over. Taking it as a chance to redesign some things would be great; e.g. I'd love to re-design navigation/information architecture in Trustroots if someone would be up for coding it. :-)

PR #1389, opened 2020-04-14
References private experience - hackweek team 1

Akronix:
#### Proposed Changes

* Change question to “did you feel comfortable hosting this person/being guest/meeting them?”
* Change colors for the answer to the above question (Yes/No) from green/red to neutral


modules/references/client/less/recommend.less
```diff
@@ -0,0 +1,3 @@
+.recommend-reply-button {
+  margin: 0 1em 0 1em !important;
```
simison, line 2: Let's avoid `important` and ass CSS specificity if there's a problem applying this otherwise. So e.g. add `.btn.recommend-reply-button`.

`0 1em` can be applied together to shorthand (two values Vs four).

Akronix: Yeah, I was also annoyed by the `!important` statement.
Thank you for the tips.

simison: CSS can be tricky. :-)

simison:
I'd always appreciate screenshots in description for easier feedback, but it's ok for me to later run this too. :-) just often commenting on phone on these.

Akronix:
@rumwerfer might you double check this? (Download branch and run it locally)
Specially, I'm interested in this point I didn't have the time to check deeply:
> * Check expected behaviour and database storing
>   * How is stored in db when you check more than one option of how do you know them

Akronix:
image
^ Changed Yes / No question. Missing text of what we expect to be answered and clarify this data won't be public (next step).

image
^ This text should be changed clarified (next step).

simison:
> Changed Yes / No question. Missing text of what we expect to be answered and clarify this data won't be public (next step).

Looks great! I wonder if there's any umabuigity about wording "feel comfortable" for non native English speakers? Copy changes seem indeed fine to follow up in separate pr.

> This text should be changed clarified

What should be clarified about it? (Don't mean to say it shouldn't)

Akronix:
> > This text should be changed clarified
> 
> What should be clarified about it? (Don't mean to say it shouldn't)

Hum so the text says "Your reference will become public", but that's not actually the case if you just gave a yes/no reference. So I think this should be clarified or better written.

simison:
I think _something_ should go public; otherwise it's just confusing if nothing goes?

Akronix:
I wasn't able to fix those tests. I'll give it another try in some other moment. A little bit of help or point where I can read how to fix it would be appreciated.

rumwerfer:
> Check expected behaviour

Apart from the "will become public" sentence, I now realized two small things I didn't think of earlier, unfortunately:
* I perceive it as a bit inelegant that we have the same sentence "How do you know ..." both as tab name and as heading. I can't come up with a good word for the tab rn.

Screenshot_2020-04-19_21-41-39

* Also, now that I look at it again, it feels weird for me that the tab above has the same color as the main button to proceed, as @simison also said somewhere else.

>Check database storing: How is stored in db when you check more than one option of how do you know them

Not sure if that's what you meant, but there are 3 separate values for the _interactions_.

Screenshot_2020-04-19_22-07-51

For this part, we wouldn't need the _public_ value anymore... but I guess the private experience and the public reference are tied together later?

modules/references/client/components/CreateReference.component.js
```diff
@@ -150,10 +156,10 @@ export default function CreateReference({ userFrom, userTo }) {
         onSelect={() => {}}
         id="create-reference-tabs"
       >
-        <Tab eventKey={0} title={t('How do you know them')} disabled>
+        <Tab eventKey={0} title={t(`How do you know ${userToName}`)} disabled>
```
simison, line 159: Let's leave the username out from here since it makes the tab overly lengthy and doesn't add any value; it's already clear from the page itself.

image

Furthermore, the username is secondary identification while their "display name" is the primary identification used thorough the UI otherwise.

modules/references/client/components/create-reference/Experience.js
```diff
@@ -17,69 +18,80 @@ export default function Recommend({
...
 
   const recommendQuestions = {
-    hostedMe: t('Would you recommend others to stay with them?'),
-    hostedThem: t('Would you recommend others to host them?'),
-    met: t('Would you recommend others to meet them?'),
+    hostedMe: t(
+      `Did you feel comfortable being hosted by ${referencedUserName}?`,
+    ),
+    hostedThem: t(`Did you feel comfortable hosting ${referencedUserName}?`),
+    met: t(`Did you feel comfortable meeting ${referencedUserName}?`),
```
simison, line 25: I suggest we remove the username for reasons described in https://github.com/Trustroots/trustroots/pull/1389/files#r411222341

modules/references/client/components/create-reference/Experience.js
```diff
@@ -17,69 +18,80 @@ export default function Recommend({
...
             onChangeReportMessage={onChangeReportMessage}
             report={report}
             reportMessage={reportMessage}
           />
         )}
+        <span className="help-block">
+          {
+            "This answer won't be shared with them and is collected for safety reasons and aggregated analysis only. "
+          }
+          {/* <a>Read more</a> */}
```
simison, line 84: Let's just remove the commented-out code.

modules/references/client/components/create-reference/Experience.js
```diff
@@ -17,69 +18,80 @@ export default function Recommend({
...
     </div>
   );
 }
 
-Recommend.propTypes = {
+Experience.propTypes = {
   primaryInteraction: PropTypes.string.isRequired,
-  recommend: PropTypes.string,
+  referencedUserName: PropTypes.string.isRequired,
+  comfortable: PropTypes.string,
```
simison, line 94: I think this prop maps to something in backend and in database model, so renaming only the clientside prop will lead to confusion.

modules/references/client/components/create-reference/Interaction.js
```diff
@@ -55,4 +61,5 @@ export default function Interaction({ interactions, onChange }) {
 Interaction.propTypes = {
   onChange: PropTypes.func.isRequired,
   interactions: PropTypes.object.isRequired,
+  referencedUserName: PropTypes.string.isRequired,
```
simison, line 64: Nit; `Username` vs `UserName`

modules/references/client/components/CreateReference.component.js
simison: We might need to shorten this for mobile anyway, in which case short "Interaction" might work best.

simison:
> > the text says "Your reference will become public", but that's not actually the case if you just gave a yes/no reference. So I think this should be clarified or better written.

> I think something should go public; otherwise it's just confusing if nothing goes?

It occurred to me that "interaction" (host/guest/met) status will anyawy be required and go public. I still think we should avoid situation where two people leave references to each other and other side wouldn't get _anything_ publicly.

modules/references/client/components/CreateReference.component.js
nicksellen, line 159: Also, to note we can't use es6 string interpolation for i18n, e.g.:

```js
t(`How do you know ${userToName}`)
```

Like this, the translation system receives something like "How do you know Nick", which won't match up with a key.

Using i18n interpolation (see [docs](https://www.i18next.com/translation-function/interpolation#basic)) gives us this:

```js
t('How do you know {{name}}', { name: userToName })
```

Now, the translation system (the `t` function) gets passed "How do you know {{name}}", which it can lookup.

(now, some part of it I don't quite understand, as the extraction process did extract the key with a variable `"How do you know $username"`, but not in the format of the other named variables `{{likeThis}}`)

modules/references/client/components/create-reference/Experience.js
nicksellen: Would be nice to find a good word for the field, it's still `recommend` in the backend and other places. I don't have any good suggestions.

I think we could just keep it as `recommend` for now though, but if we an come up with a good alternative, it needs to be changed _everywhere_ (e.g. just below too `onChangeRecommend`).

simison: Agree in preference to keep it consistent for now. :-)

PR #1412, opened 2020-04-19
Migrates pages module

edvardmk:
#### Proposed Changes

- creates react components for pages:
-- /donate
-- /donate/help
-- /donate/policy
-- /faq
-- /faq/foundation
-- /faq/technlogy
-- /faq/tribes
-- /foundation
-- /guide
-- /home
-- /media
-- /navigation
-- /privacy
-- /rules
-- /team

- angular views for these pages
- changes pages.client.routes.js

#### Testing Instructions

- npm install
- npm start
- the above listed pages

Fixes #


modules/pages/client/components/Faq.component.js
```diff
@@ -0,0 +1,356 @@
...
+          <div className="col-xs-12 col-sm-4 col-md-4">
+            <div
+              id="faq-sidebar"
+              className={classnames('sidebar', {
+                'sidebar-sticky':
+                  faq.waypoints.flags.on && faq.allowStickySidebar,
+              })}
+              // zum-waypoint="faq.waypoints" // TODO
+              // down="flags.on" // TODO
+              // up="flags.off" // TODO
```
edvardmk, line 52: How should I implement these?

modules/pages/client/components/Faq.component.js
```diff
@@ -0,0 +1,356 @@
...
+
+  // TODO connect to the API and remove mock data
+  const faq = {
+    allowStickySidebar: false,
+    waypoints: {
+      flags: {
+        on: true,
+      },
+    },
+  };
```
edvardmk, line 18: Where should I get this from?

modules/pages/client/components/Faq.component.js
simison: I think it's fine if we dont. It's not very important feature.

edvardmk: Alright, I took it out

modules/pages/client/components/Faq.component.js
edvardmk: and took this one out, too

modules/pages/client/components/Faq.component.js
edvardmk: Thanks for review!

modules/pages/client/components/Foundation.component.js
```diff
@@ -0,0 +1,151 @@
...
+              <li>
+                <a href="https://beta.companieshouse.gov.uk/company/09489825">
+                  {t('Details')}
+                </a>
+              </li>
+              <li>
+                <a href="https://ideas.trustroots.org/wordpress/wp-content/uploads/2015/03/Trustroots-Articles-2015.pdf">
+                  {t('Foundation’s Articles')}
+                </a>
+                <small className="text-muted">(pdf)</small>
```
mrkvon, line 51: a space missing

modules/pages/client/components/Foundation.component.js
```diff
@@ -0,0 +1,151 @@
...
+                <a href="/team">{t('Board')}</a>
+              </li>
+              <li>
+                <a href="/foundation">{t('FAQ')}</a>
+              </li>
+              <li>
+                <a href="https://ideas.trustroots.org/2015/03/10/announcing-trustroots-foundation/">
+                  {t('Announcement')}
+                </a>
+                <small className="text-muted">{t('(March, 2015)')}</small>
```
mrkvon, line 63: a space missing

modules/pages/client/components/Foundation.component.js
```diff
@@ -0,0 +1,151 @@
...
+import '@/config/client/i18n';
+import Board from '@/modules/core/client/components/Board.js';
+
+export default function Foundation() {
+  const { t } = useTranslation('pages');
+
+  // mock data for testing
+  const app = {
+    user: true,
+  };
```
mrkvon, line 13: Should this mock data be here, or is it temporary?
I see the button `Join Trustroots` is missing on `/foundation` page when user is not logged in. I suppose it's because this value is fixed. Is it meant to work similar like passing `user` prop to `Home` component?

edvardmk:
Thanks a lot for this valuable input @mrkvon ! I was looking out for all these things but missed this bunch in the end. How should I proceed? Make the changes and force push the branch again?

The only thing I don't know how to solve is to make the tests work and what to do about the $broadcast in the Boards component you suggested.

mrkvon:
@edvardmk 
> Make the changes and force push the branch again?

Yes, when you push changes, this pull request will be updated. You probably don't need `--force` when you add a new commit. Only if you change the already pushed commits. (e.g. after `git rebase`, `git commit --amend` or squashing commits).

> make the tests work and what to do about the $broadcast in the Boards component

I'll have a look when I continue the review (evening or tomorrow) and see if i can help. There are many changes in this PR, so it will take me a while to review. Or maybe somebody else will jump in so we have it reviewed quicker. :slightly_smiling_face:

mrkvon:
Also, notice the changes in #1405 to fix conflicts with master. (`git rebase master` or `git merge master`). In case you need, I can help with that.

modules/pages/client/components/FaqGeneral.component.js
```diff
@@ -0,0 +1,279 @@
...
+          <a href="http://team.trustroots.org/" rel="noopener">
+            developing Trustroots
+          </a>{' '}
+          . <br />
+        </Trans>
+        <br />
+        <em>
+          {t('See also')}
+          <a href="/faq/technology#f-droid">
+            {t('Could you offer mobile app at F-Droid store?')}
```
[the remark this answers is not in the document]
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/FaqGeneral.component.js
```diff
@@ -0,0 +1,279 @@
...
+
+      <div className="faq-question" id="why-is-there-facebook-connection">
+        <h3>{t('Why is there Facebook connection?')}</h3>
+        {t(
+          'So that people who have their data, photos and connections on Facebook can release their information to Trustroots. Using Facebook from Trustroots is optional.',
+        )}
+        <Trans t={t} ns="pages">
+          <a href="https://ideas.trustroots.org/2014/12/29/life-outside-big-blue-box/">
+            Read our blog post
+          </a>{' '}
```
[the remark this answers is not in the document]
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/FaqFoundation.component.js
```diff
@@ -0,0 +1,175 @@
...
+          'The guarantee of LBG under section 60 requires a clause where in case of Trustroots Foundation ceasing to exist, assets could be moved only to another non-profit entity where transferring assets is limited in the same way. CIC status would make this even more specific, but isn’t required for this.',
+        )}
+        <br />
+        <br />
+        {t(
+          'Additionally our by-laws (articles of association) specifically prohibit selling the users database and specify that the company code will be licensed under an',
+        )}
+        <a href="https://github.com/trustroots/trustroots/#license">
+          {t('open source license')}
+        </a>
```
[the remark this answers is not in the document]
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/FaqTechnology.component.js
```diff
@@ -0,0 +1,107 @@
...
+          <a href="https://github.com/Trustroots/trustroots/projects/1">
+            our roadmap
+          </a>{' '}
+          for our current todolist. Feel free to{' '}
+          <a ui-sref="support">tell us</a> about your ideas! The current version
+          of Trustroots is the most limited software you can imagine that still
+          works for hospitality. We simply haven&apos;t had time to build most
+          of the features yet. Please be patient and be mindful about
+          <a href="http://www.productstrategymeanssayingno.com/">
+            feature creep
```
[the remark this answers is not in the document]
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/FaqTechnology.component.js
```diff
@@ -0,0 +1,107 @@
...
+          Great! Read about{' '}
+          <a href="http://team.trustroots.org/">development</a> and get in touch
+          with us! We are looking for JavaScript developers (back- and frontend)
+          and sysops.
+        </Trans>
+      </div>
+
+      <div className="faq-question" id="are-you-planning-to-do-x-feature">
+        <h3>{t('Are you planning to do X feature?')}</h3>
+        <Trans t={t} ns="module">
```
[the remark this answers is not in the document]
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/Foundation.component.js
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/Foundation.component.js
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/Home.component.js
```diff
@@ -0,0 +1,461 @@
...
+      const tribeIsLoaded = tribes.some(t => t.slug === tribeRoute);
+
+      if (tribeRoute && !tribeIsLoaded) {
+        const extraTribe = await api.tribes.get(tribeRoute);
+
+        if (extraTribe && extraTribe._id) {
+          tribes.unshift(extraTribe);
+        }
+      }
+      setTribes(tribes);
```
edvardmk, line 67: I implemented the behaviour of my understanding of the angular home controller, but it behaves differently compared to production. Is that a bug in production or should this be different? @simison

modules/pages/client/components/Foundation.component.js
edvardmk: thanks, resolved in b3ef25cf

modules/pages/client/components/FaqTechnology.component.js
```diff
@@ -0,0 +1,107 @@
...
+          See
+          <a href="https://github.com/Trustroots/trustroots/projects/1">
+            our roadmap
```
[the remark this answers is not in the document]
edvardmk: thanks, resolved in b3ef25cf

nicksellen:
> Also the client tests are failing. I suspect it's related to adding $broadcast to Board component. Maybe this function will need mocking in tests. That's a guess. I haven't really investigated it.

Indeed, if you just want to make the test work, then you can add this to the test file:

```js
jest.mock('@/modules/core/client/services/angular-compat');
```

If you also want to assert that the `$broadcast` function was actually called as expected, you can do something like this:

```js
import { $broadcast } from '@/modules/core/client/services/angular-compat';

jest.mock('@/modules/core/client/services/angular-compat');

// then, later inside a test block
expect($broadcast).toHaveBeenCalledWith({ whatever: 'it gets', called: 'with' });
```

See https://github.com/Trustroots/trustroots/blob/6bdf0f9ca4243e6a4c134782aba53177e091c489/modules/messages/tests/client/components/Inbox.component.tests.js#L15 for an example of this using the `eventTrack` angular-compat function.

modules/core/client/components/Board.js
```diff
@@ -24,6 +25,7 @@ function selectName(names) {
  */
 export default function Board({
   names = 'bokeh',
+  style = null,
   children,
   onDisplayPhoto = () => {},
   onHidePhoto = () => {},
```
nicksellen, line 31: Can we entirely remove this two props now? As we `$broadcast` directly.

Although, maybe it's worth addressing the question that by using `$broadcast` directly in the component changes existing patterns of only triggering "actions" by passing in `on*` functions as props. Although we are already triggering API calls from components, but as we haven't yet added the architecture for a proper data layer (see https://github.com/Trustroots/trustroots/issues/1416), it's in a hybrid state.

Whichever data layer we move towards, I think it will end up working _something_ like how redux with react hooks would work (see https://react-redux.js.org/api/hooks#usedispatch) - which basically, dispatches an "action/event" from the component - and the `$broadcast()` call would probably be replaced by this.

The logic here to include the `$broadcast` call in the component was on my suggestion, to avoid "prop drilldown" where we have to start passing `onDisplayPhoto`/`onHidePhoto` to _all_ the static pages (now there are many in react), and this function was ultimately connected to the old angular controller methods that we also want to get rid of (ideally the react page-level components do not need to keep the old angular controller).

But in this intermediate state it might seem like this is making the components too entangled in other parts of the app. So let's progress https://github.com/Trustroots/trustroots/issues/1416 !

modules/core/client/components/BoardCredits.js
```diff
@@ -0,0 +1,76 @@
...
+  const [photoCredits, setPhotoCredits] = useState(initialPhotoCredits); // TODO: get initial value from app.photoCredits
+
+  useEffect(() => {
+    return () => {
+      $on('photoCreditsUpdated', (scope, photo) => {
+        setPhotoCredits({ ...photoCredits, ...photo });
+      });
+
+      $on('photoCreditsRemoved', (scope, photo) => {
+        setPhotoCredits({ ...photoCredits, ...photo });
```
nicksellen, line 30: This needs to _remove_ the photo. Similar to the existing handler for the core part https://github.com/Trustroots/trustroots/blob/6bdf0f9ca4243e6a4c134782aba53177e091c489/modules/core/client/controllers/app.client.controller.js#L311-L316

Maybe a nice modern approach (using https://lodash.com/docs/4.17.15#omit) would be:

```js
// at top...
import omit from 'lodash/omit';

// in $on callback handler
setPhotoCredits(omit(photoCredits, Object.keys(photo));
```

I didn't try it, but I think that should work :)

modules/pages/client/components/Donate.component.js
```diff
@@ -0,0 +1,160 @@
...
+              </a>{' '}
+              and foremost they do a great job. Sign up via{' '}
+              <a href="https://www.namecheap.com/?aff=90636">this link</a> and
+              we&apos;ll get little bit in return.
+            </Trans>
+            <br />
+            <h4>DigitalOcean</h4>
+            <Trans t={t} ns="pages">
+              Our servers are hosted at reliable and easy to use DigitalOcean.
+              Sign up via{' '}
```
nicksellen, line 122: I'll be interested to see how this turn out in the i18n json files... I also feel like there could be another solution, but not sure what it is:
- `&nbsp;` is maybe just as ugly
- disabling prettier/eslint rules for these `Trans` blocks
- maybe some CSS solution to not collapse white-space?

modules/pages/client/components/Faq.component.js
```diff
@@ -0,0 +1,346 @@
...
+          <div className="col-xs-12 col-sm-4 col-md-4">
+            <div
+              id="faq-sidebar"
+              className={classnames('sidebar', {
+                'sidebar-sticky': false,
+                // TODO implement waypoint and sticky sidebar feature by setting sidebar-sticky to: faq.waypoints.flags.on && faq.allowStickySidebar,
+              })}
+              // TODO zum-waypoint="faq.waypoints"
+              // TODO down="flags.on"
+              // TODO up="flags.off"
```
nicksellen, line 42: You can remove the commented out stuff too. I think if you/we/someone decides sticky behaviour would be desirable, it would get re-implemented in a modern way.

On that, when doing a quick search for a react sticky library, I found this [1], which told me that CSS has a `position: sticky` option now, it's not usable in all browsers, but that would be OK for us.

The description of what it does on https://developer.mozilla.org/en-US/docs/Web/CSS/position hurt my brain:

> The element is positioned according to the normal flow of the document, and then offset relative to its nearest scrolling ancestor and containing block (nearest block-level ancestor), including table-related elements, based on the values of top, right, bottom, and left. 

We don't need to add it, but thought it was worth mentioning :) Some people like CSS puzzles...

[1] https://www.npmjs.com/package/react-sticky#css

modules/pages/client/components/Donate.component.js
```diff
@@ -0,0 +1,160 @@
...
+                  <a href="/donate/policy">donor policy</a>. Trustroots
+                  Foundation is located in the United Kingdom. We do not sell or
+                  trade your information to anyone.
+                </Trans>
+              </small>
+            </p>
+          </div>
+          <a
+            className="btn btn-xs btn-primary pull-right"
+            href="https://github.com/Trustroots/trustroots/edit/master/modules/pages/client/views/donate.client.view.html"
```
nicksellen, line 145: The edit links are not valid any more as they are pointing to angular templates.

With i18n, editing is a bit more complex, so arguably we could remove all the `Edit this page` links? Whaddya you say @simison?

modules/pages/client/components/Faq.component.js
simison: Ohh, I was just reading about sticky CSS another day and should've remembered to mention it here. Totally not necessary, but it'd be great if it would work. Ok for follow-ups too.

modules/pages/client/components/Donate.component.js
simison: Good to go away indeed :thumbsup:

nicksellen:
Something seems a bit up with the photo credits:
- I visit http://localhost:3000/faq/tribes
- I see a nice photo, and a photo credit at the bottom, all good
- I click on "Technology" (for example, but I don't think matters which one)
- I see a nice new photo, but there is no photo credit any more
- I do a full page reload of http://localhost:3000/faq/technology
- there is a photo + credit again

So, it's something around when the pages change after first load does not work. Maybe try tracing through what happens and what is fired (I use `console.log()` statements for this kind of stuff, but explore fancier debugger options if you want).

One thing I did notice is that the `useEffect` block in the board `Board` needs to declare `names` as a dependency [1], I think this was already like this in master maybe. Not sure if this is related to the issue.

[1] quick recap of `useEffect` "dependencies":

```js
useEffect(() => {
}, []) // <--- this array is the "dependencies" (controlling when it will re-run the block)

// so if we use a prop

function MyComponent({ someProp }) {
  useEffect(() => {
     // does something with the prop...
     fetchThings(someProp);
  }, [someProp]); // <--- declaring the dependency on "someProp"
}

```

(I think https://reactjs.org/docs/hooks-overview.html is a nice overview of how hooks work, not currently loading for me though... 520 error...)

modules/pages/client/components/Team.component.js
```diff
@@ -0,0 +1,497 @@
...
+            </div>
+          </div>
+        </div>
+        <div className="col-xs-12 col-sm-4">
+          {/* Kasper */}
+          <div className="media">
+            <a className="pull-left" href="https://guaka.org/">
+              <img
+                className="media-object img-circle"
+                src="//gravatar.com/avatar/edccc1f234833b3b5c266c18c2db139d?s=200"
```
nicksellen, line 207: Kaspers photo is broken, same on production, but maybe we can find a fresh one :)

nicksellen:
Great work! This is a significant effort, and nearly ready I think! I think a lot of learning went into this.

I check every page and didn't see anything wrong (apart from photo credits behaviour described above). I didn't click all the links but assume they are fine.

I'm not quite sure what the `/navigation` page is. I didn't look into it, have to eat now... looked a bit like a thing that is included into another thing, do we need it in react yet? (Or rather, is it used?). I probably couldn't have answer that myself in the time it took to write this paragraph.

simison:
> I'm not quite sure what the /navigation page is.

It's linked from mobile header, 3 lines link:

Screenshot_20200421-220128_Firefox

I think it'll need to be changed to "slide out" menu but it's useful to turn it into react already now.

modules/pages/client/components/Donate.component.js
edvardmk: For me, this is an HTML problem needing an HTML solution, cause we would need to insert an optional space or line break in case that an inline element follows the Trans block. I find that {' '} is the best I know since
- it breaks when necessary
- it creates a space when needed
- it only creates a space after the last character and not between the whole paragraph and the next element

I might miss out on something here but from what I understand, it appears the best solution to me that comes to my mind, given that it works out with the i18n json files.

modules/pages/client/components/Faq.component.js
edvardmk: Alright, I took that out and left a note about implementing 'position: sticky' in the future

modules/pages/client/components/Donate.component.js
edvardmk: Alright, took that out, too.

modules/pages/client/components/Team.component.js
nicksellen: Here is a photo of him https://www.crunchbase.com/person/kasper-souren :)

Or this nice one from 15 years ago https://sw.m.wikipedia.org/wiki/Picha:Kasper_Souren_Berlin_Beach_Camp_september_2005.jpg

nicksellen:
After chatting with @edvardmk we decided it might be better if I fixed up the remaining issues, they were:

1. the broadcast calls were being made far too much, due to `useEffect` dependency using array value, fixed in https://github.com/Trustroots/trustroots/pull/1412/commits/3aa531cd4e651ee1f436acb8498d89f6c6f71d65
2. `vm.photoCreditsCount` was ended up as `0` even when there were some, causing the photos credits not to be displayed (this is actually a bug on existing site, not just on this branch as I had thought in https://github.com/Trustroots/trustroots/pull/1412#issuecomment-617348660) fixed in https://github.com/Trustroots/trustroots/pull/1412/commits/03f7aa31764cfc78390d91c114039a817ecefe4e
3. the subscribe/unsubscribe to `$on` events with `useEffect` wasn't quite right, fixed in https://github.com/Trustroots/trustroots/pull/1412/commits/0b16cb7204504d287388109be4e5a4e3517e22c5
4. test failures because we  use angular-compat now, which needs to be mocked with jest, as described in https://github.com/Trustroots/trustroots/pull/1412#issuecomment-617321269, fixed in https://github.com/Trustroots/trustroots/pull/1412/commits/dae7fff8ce2add08870d862a664f9d369538a57e (also added a basic test in https://github.com/Trustroots/trustroots/pull/1412/commits/fd0cef8bc82a15363a7ef058c950efdc584baa40)

The commits themselves have reasonable detailed messages, and I also removed a few unneeded comments/TODO/files...

modules/pages/client/components/Team.component.js
nicksellen: It's broken in prod too, so can leave it for the unknown future :)

PR #1413, opened 2020-04-19
Add public feedback UI

nicksellen:
#### Proposed Changes

* add third tab to the create reference flow to add optional public feedback

tr-feedback

#### Testing Instructions

* `git switch add/feedback-public-ui`
* `npm start`
* find a user on the map
* view their profile
* click "create a reference" near the top
* go through the 3 steps
* add some public feedback text
* see that it submits successfully
* (it won't be displayed on their profile, that is being done in a future PR)
* ... try the whole thing again, but _don't_ leave a public reference
* it should still work (as it's optional)

I would particularly like feedback on the text on the page, it should maybe be clearer that it is optional, and could have some better guidance for the user. As I understand it, @Akronix was particularly keen to have it that the user is writing content that gives a more complete picture of the user, to complement their profile, so not really writing about the visit/stay/meet itself, but about the person.

Some optimizations/tweaks I could add if people are interested:
- have a checkbox to first tick, to confirm they want to write something, before showing the text box
- ... maybe that's it

There are a few thoughts from @simison on the old PR: https://github.com/Trustroots/trustroots/pull/1384#issuecomment-613518304

Part of https://github.com/Trustroots/trustroots/issues/718


Akronix:
Personally, I'm OK with the change.
Although I'd like it more with a "Finish" button in the previous "Yes/No" step and next to it another "Give public feedback" or sth along those lines. So it feels more optional.
Maybe good to leave it like this at the beginning so we also get more public feedback from users.

modules/references/client/components/create-reference/Feedback.js
```diff
@@ -0,0 +1,42 @@
...
+        </label>
+        <textarea
+          className="form-control input-lg"
+          rows="7"
+          id="feedback-message"
+          onChange={event => onChangeFeedback(event.target.value)}
+          value={feedback}
+        ></textarea>
+        <span className="help-block">
+          {t('Were they were they great at cooking? or singing?')}
```
Akronix, line 31: I'm OK with the text on the page, but I don't consider myself an expert on this.
Just correct this repeated excerpt:
"Were they were they"

Akronix:
Now I see it there is not maximum length limit, right? Should be added me think as discussed here: https://codi.kanthaus.online/trustroots-references

simison:
> I'd like it more with a "Finish" button in the previous "Yes/No" step and next to it another "Give public feedback" or sth along those lines. So it feels more optional.

I didn't understand this, could you rephrase? :-)

nicksellen:
> Now I see it there is not maximum length limit, right? Should be added me think as discussed here: https://codi.kanthaus.online/trustroots-references

Ah, yeah. There is on the server, but not the client. It would just give an incomprehensable error to the user.

Pretty please can I postpone it to a references round 2 PR? We can create an issue to collect the refinements?

modules/references/client/components/create-reference/Feedback.js
nicksellen: Fixed in https://github.com/Trustroots/trustroots/pull/1413/commits/513bc81b096bf7d88c9d5975ddf30e7a4cfddec1

Akronix:
This is OK for me.
Replying to @simison this is what I mean:
image

Akronix:
image
We were discussing @MatthijsMali and I and we came up with this text that makes sense for a good flow when writing the public feedback.
> The reason for doing this: First let the user know what to do. Then if they are curious they can find other information (what do they do with this information).

simison:
> image

Thanks for the example! I appreciate you might've not meant this to be exactly how we should do it, but I'll comment on it anyway since I think this can be helpful. I think this UI example would be problematic for a couple of reasons, I'll outline why I think so here but we work on it in follow-up PRs and hash out the details.

The goal is really just to communicate that the final step is optional, right? :-)

- Two "equal" (color, placement) call-for-actions are more confusing than need to be; offering a split way via CTA buttons doesn't necessarily help if the user is just looking for a way to continue; they now need to make a decision based on button label and nothing else. If we want to avoid actually showing them the final step, it could be a toggle in the content of this "recommendation" step. I don't think that's optimal either because:
- User doesn't know what the next optional step is (or that there even is one). Either one of these buttons could equally indicate "publish my recommendation". User doesn't yet know that "feedback" is a separate free-text thing, so it's harder for them to make a decision to skip it.
- I think it's most clear if we just let the user go to the final feedback step to see what it is about before they need to decide if they wanna skip it. I also think we should push them more towards not needing necessarily to fill the freetext, but that's a separate design problem.

Lemme know what you think! We can also continue on a separate issue. :-)


> image
>
> We were discussing MatthijsMali and I and we came up with this text that makes sense for a good flow when writing the public feedback.
>
> The reason for doing this: First let the user know what to do. Then if they are curious they can find other information (what do they do with this information).

Valid point! I'd argue that since both guides are actually quite important to how they might write the description, and the bit about this being public being actually quite critical, I'd move them both above the field and instead of using small+grey text, have it in regular text size + color. Since they're important, we should optimize for users really reading them.

Screenshot 2020-04-20 at 12 45 14

What do you think?

modules/references/client/components/create-reference/Feedback.js
```diff
@@ -0,0 +1,42 @@
...
+        </label>
+        <textarea
+          className="form-control input-lg"
+          rows="7"
+          id="feedback-message"
+          onChange={event => onChangeFeedback(event.target.value)}
+          value={feedback}
+        ></textarea>
+        <span className="help-block">
+          {t('Were they great at cooking? singing? dancing?')}
```
mrkvon, line 31: Non-blocking:

I think this guideline would encourage people to judge.
I would prefer a culture in which people say `I felt such and such, experienced that and that, needed that and that.` instead of `This person was such and such.`. This point of view is shaped heavily by my understanding of [Nonviolent Communication](https://en.wikipedia.org/wiki/Nonviolent_Communication).
This should not block merging. If there is a desire to talk about this, let's do it independently of this PR; and iterate on the guideline only after such discussion.
I just really wanted to express my point of view on this, because it matters to me.

modules/references/client/components/create-reference/Feedback.js
```diff
@@ -0,0 +1,42 @@
...
+          {t('Would you like to describe something about them?')}
+        </h4>
+      </div>
+      <div
+        className="panel-body"
+        role="group"
+        aria-labelledby="feedback-public-question"
+      >
+        <label htmlFor="feedback-message" className="control-label">
+          {t('Public feedback')}
```
mrkvon, line 21: Non-blocking:

It's not clear to me that one can/should skip this step. Perhaps adding `(optional)` here or somewhere would help communicate our intention: `"Speak only if you have something to say"`.
WDYT?

simison: "(Optional)" seems like a great quick way to make it clearer 👍

nicksellen:
OK, I reworked the text a bit, sort-of incorporating a few ideas from everyone (maybe it loses all clarity in the process!)

* very _slightly_ more asking people to speak from their personal experience @mrkvon - although I found it hard to align with the idea I got from @Akronix, which was about getting them to write extra descriptive content that functions like an extended profile (so more about _them_ than the _experience_)
* added a bit more extended text from @MatthijsMali and @Akronix 
* moved the "help" text into the main body in normal size as @simison mentioned

It looks like this now:

publicfeedback

Unless there is a blocking request, I would propose we merge this as-is, then @Akronix can rebase, and potentially work on further changes in that branch (as it deals with various aspects of language), _or_ a totally seperate PR are in the future.

PR #1415, opened 2020-04-20
fix port for maildev

sitarane:
#### Proposed Changes

* Fix Maildev port from wrong to right

#### Testing Instructions

* Open localhost:1080. Do you see Maildev?

Fixes #

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ maildev:
   container_name: maildev
   restart: always
   ports:
-    - "1080:1080"
+    - "1080:80"
 
 ## Database container
 mongodb:


nicksellen:
The image we're [1] using it actually out of date and tells us to use a newer image [2].

The newer image listens on port 1080 (from the usage example: `docker run -p 1080:1080 -p 1025:1025 maildev/maildev`), so just switching to the new image and reverting the port change should work.

It might be even nicer to explicitly configure the port to what we want, it supports env variables, e.g. `MAILDEV_WEB_PORT`, and then if the underlying image/app changes it's default, we are not impacted, but that could be optional.

[1] https://hub.docker.com/r/djfarrelly/maildev 
[2] https://hub.docker.com/r/maildev/maildev

PR #1431, opened 2020-04-26
Docker compose update

sitarane:
#### Proposed Changes

S'all in the commits...

* Remove maildev container: maildev already runs in the main container with Trustroots app
* Remove nginx service
* Upgrade docker-compose syntax from ? to 3.7 (current)
* Tweak a couple of commands here and there
* Updates to docs
* Copy both `package.json` and `package-lock.json` for `node_modules` cache layer busting
* Remove redundant `imagemagick`
* Remove InfluxDB container: this might be nice to add back in follow-ups, but for now the one we used was deprecated.
* Expose more ports

#### Testing Instructions

* Ensure you have `host: "0.0.0.0"` in your `config/env/local/js` file
* If you had previous instances, you might wanna `docker-compose build` again
* `docker-compose up`
* works?


docker-compose.yml
```diff
@@ -1,79 +1,82 @@
...
+    #  that they won't get overwritten by app directory
+    #  Read more: http://stackoverflow.com/a/32785014/1984644
+    networks:
+      - trustroots
+    volumes:
+      - ./:/trustroots
+      - node-modules:/trustroots/node_modules
+      - public:/trustroots/public
+      - uploads:/trustroots/uploads
+    command: npm run start
```
nicksellen, line 29: This now means it runs `maildev` inside the trustroots container _aswell_ as the one in the seperate container. The `start:docker` command is basically the same, except that it doesn't run `maildev`.

I'm pretty agnostic on whether it's better to run it in it's own container, or inside the trustroots container, but it should only run in one.

**if running in trustroots container**

- remove the `maildev` container
- move the mapped port to the trustroots container
- make sure the local.docker.js config stuff reflects the change too (I had some more thoughts on this I'll write seperately, as it could be for another PR)
- remove "start:docker" script definition from package.json

**if running standalone**

- either switch back to `start:docker` command...
- ...or could implement an environment variable to disable starting maildev (e.g. something like `"dashboard:mail": "[ -n \"$DISABLE_MAILDEV\" ] || maildev",`, although that's a bit ugly imo...)

docker-compose.yml
```diff
@@ -1,79 +1,82 @@
...
+    depends_on:
+      - mongodb
+      - maildev
+      - influxdb
+    ports:
+      - "3000:3000"
+      - "5858:5858"
+      - "35729:35729"
```
nicksellen, line 37: This can go away, it was for livereload which we don't use any more (webpack dev server connects via websocket connection on the same port for the main service, 3000).

nicksellen:
Oh, also re: [this slack thread](https://trustroots.slack.com/archives/C0A3Q15SS/p1588578399007500) my docker setup didn't work out the box, because it didn't install the correct npm deps... again, I don't think we need a solution in this PR, possibly needs more general discussion in slack about the best approach.

docker-compose.yml
sitarane: Well we gotta keep 3000 so we can access it from browser

docker-compose.yml
sitarane: I have done that.

> trustroots@0.3.3 install-dependencies /trustroots
> node ./bin/install-deps.js

internal/modules/cjs/loader.js:960
  throw err;
  ^

Error: Cannot find module '/trustroots/bin/install-deps.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:957:15)
    at Function.Module._load (internal/modules/cjs/loader.js:840:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
    at internal/main/run_main_module.js:18:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

I don't feel very qualified to fuck around with node (or whatever this all is...)

sitarane: Done. We'll know if it's good once the one below is resolved. I couldn't get to that step

sitarane:
Aw shit! I can't see the complaint about the "npm start"

I was going to add that shouldn't the "install dep" script conflict with npm start? I had a look in it and it seems to check for node-modules and if not there install deps, which, not that I'd know, is what npm start does.

Now what did I do wrong that the comment isn't here anymore?

docker-compose.yml
nicksellen: Yup, I just meant port `35729` :)

simison:
Rebased

docker-compose.yml
simison: I removed the separate container for Maildev from Docker compose. It's heavy handed to run it in its own container if it can simply run together with the app — and that's where its version is already maintained. Yay for no duplication. :-)

docker-compose.yml
```diff
@@ -1,79 +1,49 @@
...
-## http://djfarrelly.github.io/MailDev/
-maildev:
-  image: djfarrelly/maildev
-  container_name: maildev
-  restart: always
-  ports:
-    - '1080:1080'
+networks:
+  trustroots-network:
+    driver: bridge
```
simison, line 5: Todo, need to read a bit more about this `driver: bridge`

docker-compose.yml
```diff
@@ -1,79 +1,49 @@
...
-    - ./scripts/docker/nginx/trustroots.dev:/etc/nginx/vhost.d/trustroots.dev
-    - ./scripts/docker/nginx/trustroots.dev_location:/etc/nginx/vhost.d/trustroots.dev_location
-  ports:
-    - '3080:80'
+  ## Database container
+  mongodb:
+    image: mongo:4.2
+    container_name: db_1
+    volumes:
+      - dbdata:/data/db
```
simison, line 43: Might be we don't need volume here.

docker-compose.yml
```diff
@@ -1,79 +1,49 @@
...
+    restart: unless-stopped
+    command: '--smallfiles --logpath=/dev/null'
```
simison, line 45: TODO: check if this is really needed anymore.

docker-compose.yml
```diff
@@ -1,79 +1,49 @@
...
-mongodb:
-  image: mongo:3.7
-  container_name: db_1
-  restart: always
-  command: '--smallfiles --logpath=/dev/null'
-  ports:
-    - '27017:27017'
+services:
+  trustroots:
+    image: node:12
```
simison, line 9: Need to adjust `Dockerfile` for this.

Issue #1381, opened 2020-04-13
Remove unused profiles after x years


rumwerfer:
Profiles that were not used for years are cluttering the map. Also, it doesn't make sense for us to store their data in the database forever.

After not logging in for x time (@simison's suggestion: 3 years), an email should be sent out saying "Do you still want to use your Trustroots profile? If so, log in. If not, we will delete your profile soon."
There should be a reasonable amount of time (2 weeks?, maybe even a second email?) before the profile is actually deleted.

simison:
To be clear, we do hide these from map by default (you can see the toggle in filters).

It's the fact of keepin un-needed personal data around that's not much fun.

People sometimes try sign-up and forget they already have an account, too. :-D

Here is a nice podcast episode on topic: https://rework.fm/delete-your-account/

rumwerfer:
Oh! I knew about the filter, but wasn't aware it was activated by default.

simison:
I _think_ it is. :-) if it's not, that's a problem to fix. :-) should show up in incognito window.

rumwerfer:
> should show up in incognito window.
it doesn't

Lemoert:
I like the idea of cleaning out old profiles, but perhaps first a "Have you forgotten about us?" email after two years and then a "Hello, we'll delete your account." email about a year later and deletion three-four weeks after that. Would having two emails be trickier to implement than one? Also, should this be combined with notifications seeing as a lot of people do not check their emails often?

simison:
@Lemoert yep, email/notification sequence should definetely be part of it.

Multiple emails in the sequence is no problemo.

Another thing we might wanna do is disable the account first before deleting a month later, so that there is still "last minute" recovery possibility.

PR #1434, opened 2020-04-29
Search filter: enable "online in the past 6 months" by default

simison:
As @rumwerfer noted in https://github.com/Trustroots/trustroots/issues/1381#issuecomment-615365707, "online within the last 6 months" toggle in map search wasn't on by default.

This makes 6 months default and requires people to toggle it off to get users from the past 24 months. Anyone not logged longer than that won't be visible on the map no matter how the toggle is set.

#### Proposed Changes

* Enable "online in the past 6 months" by default

#### Testing Instructions

* Open map search and filters panel
* You can either open in an incognito window to ensure you don't have filter settings in local storage saved, or you can manually delete "trustroots.search.filters" entry as seen in 2nd screenshot. Go to the "Application" tab and to "Local Storage".

Screenshot 2020-04-29 at 09 14 30
Screenshot 2020-04-29 at 09 14 50

By default, the toggle is on. You can also ensure from networks tab that correct "seen" value is sent (24 when the toggle is disabled, 6 months when enabled):

image

### DIFF
--- a/modules/search/client/services/filters.client.service.js
+++ b/modules/search/client/services/filters.client.service.js
@@ -11,7 +11,7 @@ function FiltersService($log, Authentication, locker) {
     types: ['host', 'meet'],
     languages: [],
     seen: {
-      months: 24,
+      months: 6,
     },
   };
 


sitarane:
Hey, I don't see how this:

Screenshot from 2020-04-29 23-35-11

Achieves what we want here.

Pass test on Firefox and Chrome, but I'm not sure where to look for the code changes.

nicksellen:
I agree with @sitarane - it looks like you forgot to commit a change?

simison:
No that's all we need. :-) The existing toggle just switches between 6 and 24 months.

Default filters are returned when no filters are saved in localStorage:
https://github.com/Trustroots/trustroots/blob/c39d01e8dff77eb089d8a4b008775fe496c162cf/modules/search/client/services/filters.client.service.js#L23-L27

Here's the toggle which is now on by default since it sees 6 months vs 24 months as value:

https://github.com/Trustroots/trustroots/blob/c39d01e8dff77eb089d8a4b008775fe496c162cf/modules/search/client/views/search-sidebar-filters.client.view.html#L38-L45
https://github.com/Trustroots/trustroots/blob/c39d01e8dff77eb089d8a4b008775fe496c162cf/modules/search/client/controllers/search.client.controller.js#L150-L161

Without the change the filter defaults to 24 months.

Unless I'm missing something?

nicksellen:
OK, I see now :) Actually, I don't see how I didn't see before. Oh well, brains are a funny thing.

simison:
It's a little complicated bit of code just nature of Angular.js services, hope we can simplify it with React. :-)

PR #1447, opened 2020-05-05
Mobile app: improved detection

simison:
Another attempt at fixing https://github.com/Trustroots/trustroots/pull/1139#issuecomment-618893929 and unblock iOS release: https://github.com/Trustroots/trustroots-expo-mobile/issues/10

#### Proposed Changes

* Instead of waiting for native mobile app wrapper send us a signal and _then_ setting `window.isNativeMobileApp`, set it already on server-side rendered template if there's `?app` query argument.

I updated the mobile app load this way in https://github.com/Trustroots/trustroots-expo-mobile/pull/15/files#diff-958f6f56cd42c904603d226cd36a57b6R78 a while back. There might be some apps out there that don't support this but I'm not overly concerned about those; they'll keep functioning anyway.

Motivation is to make easier to modify UI for mobile app wrapper, more acutely to hide "app store badges" for iOS release ([Apple requirement](https://github.com/Trustroots/trustroots-expo-mobile/issues/10#issuecomment-528238274)).

#### Testing Instructions

* Open page via `http://localhost:3000?app`
* Ensure as logged-out user that frontpage doesn't have app store badges
* Ensure as logged-in usre that http://localhost:3000/navigation doesn't have app store badges
* Ensure that otherwise via `http://localhost:3000` those are visible

Screenshot 2020-05-05 at 23 38 46
Screenshot 2020-05-05 at 23 38 25
Screenshot 2020-05-05 at 23 38 12
Screenshot 2020-05-05 at 23 38 05


config/lib/express.js
```diff
@@ -102,6 +102,12 @@ module.exports.initLocalVariables = function(app) {
     // https://expressjs.com/en/api.html#req.path
     res.locals.canonicalUrl = res.locals.hostPort + req.path;
 
+    // Native mobile app wrapper loads the site with `?app` URL query argument.
+    res.locals.isNativeMobileApp = Object.prototype.hasOwnProperty.call(
```
nicksellen, line 106: We have lodash `has` here presumably? Would look nicer/neater :)

simison: Ahh yes, you're right. I was too sleepy to think. :-)

PR #1465, opened 2020-05-18
Fix Map screenshot not based on real data

### LINKED ISSUE #1367: Map screenshot not based on real data
Very minor issue: I just introduced a friend to TR, and he said: "Well, that's weird. You guys have 
100 members in Luxembourg and noone in Berlin?" – Took me a while to understand he was referring to the map screenshot on the startpage (not logged in):

screenshot from startpage

I guess this is still from the early times of Trustroots when the real map was still empty. But now it might make more sense to have the map resemble the real locations of our members. Okay if I replace it with this one?

screenshot of actual map

gabrielsr:
#### Proposed Changes

* Change the image with the map on the home page with an updated one already provided by @rumwerfer. 

#### Testing Instructions

* In the old version of the home you see the map with 13 in London region.
* In the new version you see 41. 

Fixes #1367


simison:
Thanks! I think these are too big; currently we have online 1MB 2x png (which is also way too big for something like this) and this one is 2.5mb.

Can we squeeze these smaller? Note that they're not very big at the homepage.

gabrielsr:
I agree that this is too big! I tried to use Gimp to optimize it while keeping the format and resolution and was not able to get a smaller size, though =/

simison:
Closing in favor of https://github.com/Trustroots/trustroots/pull/1931

PR #1466, opened 2020-05-18
Stats: record spoken languages daily

simison:
Collects spoken languages from profiles as counts to Influx DB.

Kudos @nicksellen for writing the query:
```
db.users.aggregate([{$unwind: "$languages" }, { $group: { _id: '$languages', count: { $sum: 1 }  } }, { $sort: { count: -1 } }, { $limit: 40 }]).map(item => ([item._id, item.count].join(' '))).join('\n')
```

Which today on live DB gives:
```
eng 30895
spa 9448
fre 8876
ger 7660
ita 2742
rus 2715
por 2489
dut 1723
pol 1616
ara 1081
tur 1074
cze 858
hin 852
swe 792
chi 739
cat 535
rum 505
dan 486
hun 478
ukr 465
```

#### Proposed Changes

* Adds collecting top 40 languages in daily stats run

#### Testing Instructions

* Add some languages to the database: you can manually modify `users` collection and add codes to `languages` entry for profiles, or alternatively save some languages to profiles via UI.
* Run the app `npm start`
* Run the worker dash: `npm run dashboard:jobs` and open `http://localhost:1081/` to observe jobs.
* Pick "daily stats" and hit rerun:

    image

* You should observe in console the job run with query:
    
    image
    
    You could add some console.logging to the job in [`modules/statistics/server/jobs/daily-statistics.server.job.js`](https://github.com/Trustroots/trustroots/pull/1466/files#diff-0752896853e5ef2c8805dc0e8d6e8c1aR236) to see what's going on as I've done.

### DIFF
--- a/modules/statistics/server/jobs/daily-statistics.server.job.js
+++ b/modules/statistics/server/jobs/daily-statistics.server.job.js
@@ -230,6 +230,36 @@ module.exports = function (job, agendaDone) {
           done,
         );
       },
+
+      // Get number of languages spoken
+      function (done) {
+        statistics.getUserLanguagesCount(40, function (err, counts) {
+          if (err) {
+            log(
+              'error',
+              'Daily statistics: failed fetching languages spoken counts.',
+              err,
+            );
+            return done();
+          }
+
+          // Write numbers to stats
+          counts.map(({ _id, count }) => {
+            writeDailyStat(
+              {
+                namespace: 'language',
+                values: {
+                  count,
+                },
+                tags: {
+                  language: _id,
+                },
+              },
+              done,
+            );
+          });
+        });
+      },
     ],
     function (err) {
       if (err) {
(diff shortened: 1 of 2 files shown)


modules/statistics/server/jobs/daily-statistics.server.job.js
```diff
@@ -230,6 +230,41 @@ module.exports = function (job, agendaDone) {
...
+            log(
+              'error',
+              'Daily statistics: failed fetching spoken languages counts.',
+              err,
+            );
+            return done();
+          }
+
+          // Write numbers to stats
+          async.eachOfSeries(
```
simison, line 247: We run through each language and send stats to InfluxDB separately for each.

simison:
I'd like to switch this to promises format so that it would be much easier to read and work with. But that's for another time and PR! :-)

modules/statistics/server/jobs/daily-statistics.server.job.js
nicksellen: Seems a bit inefficient as you can send many stats at once to influxdb, maybe we don't have that ability exposed right now though.

Not blocking.

simison:
> Seems a bit inefficient as you can send many stats at once to influxdb, maybe we don't have that ability exposed right now though.

Yeah, that'd be great follow-up.

PR #1476, opened 2020-05-22
Fixed Comments

NobleTyler:
#### Proposed Changes
Fixed spelling and reformatted non ES2015 style comments. As well as removed redundant import comment.
*

#### Testing Instructions
No tests required as it was just comment changes
*

Fixes #
Documentation


nicksellen:
Thanks! It failed the linting check in CI, you should be able to see the report pressing the "Details" button (which takes you to https://travis-ci.org/github/Trustroots/trustroots/jobs/690183840).

You can run the linting locally with `npm run lint`, or try autofixing with `npm run lint -- --fix` :)

Also, you can probably get rid of the line with `eslint-disable-next-line angular/window-service ` in it, as there is no use of window anymore anyway, and actually also the line above (`When testing, `window.env` is undefined, thus default to 'test'`) as `window.env` is not longer used, and the testing infrastructure has changed since then, so maybe not even true any more so removing that line seems reasonable, and just keeping the first 2 lines of the comment but just formatted to please the linter :)

Otherwise looks good!

Issue #1467, opened 2020-05-19
A mechanism for users to block abusive users


simison:
**Is your feature request related to a problem? Please describe.**
Users would need to be able to block each other right away, without writing to and waiting for support to remove offending users.

- [x] Add API. Done in https://github.com/Trustroots/trustroots/pull/1490
  - [x]  Administrative user should be able to list a user nonetheless
  - [ ] Hide messages and events in both ways
  - [ ] Anything else that needs to be hidden from blocked users? 
- [x] Add UI for blocking users in website https://github.com/Trustroots/trustroots/pull/2121
  - [x] In the message thread
  - [x] In profile
- [x] Add UI for un-blocking users in website https://github.com/Trustroots/trustroots/pull/2121
  - [x] In the message thread
  - [x] In profile
- [ ] Add UI to see blocked users as a list at your account
- [ ] Add UI in native mobile app (https://github.com/Trustroots/trustroots-mobile)

**Describe the solution you'd like**
Add "block member" link in users' profiles:
<img src="https://user-images.githubusercontent.com/87168/82382858-6a903000-9a35-11ea-81d6-87a18fa9fbf4.png" alt="" width="300"/>

Clicking the button can open regular JS [`window.confirm()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) question,  doesn't need anything fancier:

<img src="https://user-images.githubusercontent.com/87168/82383125-dd011000-9a35-11ea-890f-6dd47be99eaf.png" alt="" width="500"/>

(with different text ;-) )

Confirming this, it will add blocked user's ID in authenticated user's profile.

When A wants to block B, A will end up with profile data looking like:
```
user: A,
blocked: [ B ]
```

This should work similar to `shadowban` role ([see code](https://github.com/Trustroots/trustroots/search?q=shadowban&unscoped_q=shadowban), or issue https://github.com/Trustroots/trustroots/issues/1184) in a way that blocker user just gets hidden for the other user quietly.

**Describe alternatives you've considered**
—

**Additional context**

This will help us comply with Apple's app review requirement: https://github.com/Trustroots/trustroots-expo-mobile/issues/10#issuecomment-631104166 ...and it's a nice safety feature regardless.

gabrielsr:
Working on an api for it.

loberto:
could this be a simple check box in the report dialog box?

simison:
Good idea!

[...]

PR #1490, opened 2020-05-25
api for blocking users

gabrielsr:
The backend part of #1467, including new routes

Workin in progress. 

- [x] Added basic test cases  - partial impl and not passing.
- [x] Add new routes.
- [x] Add 'bocked' to the model
- [x] Impl Update profile with blocked users
- [x] Impl Filter blocking user of blocked users queries
- [x] Impl Filter blocked users of blocking user queries
- [x] passing: see not blocked
- [x] passing: not seen by blocked
- [x] passing: not see blocked
- [x] passing: list blocked
- [x] passing: unblock
- [ ] Administrative user should be able to list a user nonetheless
- [ ] Hide messages and events in both ways (anything more?) 

#### Proposed Changes

* Include ROUTES:
GET: '/api/blocked-users'
POST, DELETE: '/api/blocked-users/:username': 

* Altered 
GET: '/api/users/:username'
GET: '/api/users?search=?str'

#### Testing Instructions

* check new test cases
* exercise the API using two api clients with different auths

Fixes #

- [ #1467 ]  - API.


modules/users/tests/server/user-block.server.routes.tests.js
```diff
@@ -0,0 +1,279 @@
...
+ * User routes tests
+ */
+describe('User block - user', function () {
+  before(function (done) {
+    // Get application
+    app = express.init(mongoose.connection);
+    agent = request.agent(app);
+
+    done();
+  });
```
simison, line 29: I know this is how most of the app does this, but it can also be simplified nowadays to this:

https://github.com/Trustroots/trustroots/blob/aaa8b9ed0269c8e4ea157bcc6097db18b7eb5719/modules/admin/tests/server/admin.audit-log.server.routes.tests.js#L8-L10

Not a blocker for merging.

modules/users/tests/server/user-block.server.routes.tests.js
```diff
@@ -0,0 +1,279 @@
...
+    };
+    const carolUser = new User(carolProfile);
+    carol = {
+      credentials: carolCredentials,
+      user: carolUser,
+      profile: carolProfile,
+    };
+
+    // Save alice and bob to test db
+    Promise.all([aliceUser.save(), bobUser.save(), carolUser.save()])
```
simison, line 99: I know this is how most of the app does this, but it can also be simplified nowadays to this:

https://github.com/Trustroots/trustroots/blob/aaa8b9ed0269c8e4ea157bcc6097db18b7eb5719/modules/admin/tests/server/admin.audit-log.server.routes.tests.js#L12-L19

Not a blocker for merging.

gabrielsr: Nice! Much more simple.

modules/users/tests/server/user-block.server.routes.tests.js
gabrielsr: Nice! Much more simple.

modules/users/server/routes/users.server.routes.js
```diff
@@ -26,6 +27,17 @@ module.exports = function (app) {
     .all(usersPolicy.isAllowed)
     .post(userProfile.avatarUploadField, userProfile.avatarUpload);
 
+  app
+    .route('/api/users/blocked-users')
+    .all(usersPolicy.isAllowed)
+    .get(userBlock.getBlockedUsers);
```
simison, line 33: What would be scenario where we need to fetch a list of IDs via API?

modules/users/tests/server/user-block.server.routes.tests.js
```diff
@@ -0,0 +1,279 @@
...
+            return done();
+          });
+      });
+  });
+
+  it('should not see a user if users blocked by her', function (done) {
+    /* alice login */
+    agent
+      .post('/api/auth/signin')
+      .send(alice.credentials)
```
simison, line 135: Some examples how to simplify this nowadays:

https://github.com/Trustroots/trustroots/blob/aaa8b9ed0269c8e4ea157bcc6097db18b7eb5719/modules/admin/tests/server/admin.users.server.routes.tests.js#L115-L122

https://github.com/Trustroots/trustroots/blob/aaa8b9ed0269c8e4ea157bcc6097db18b7eb5719/modules/references/tests/server/reference-read-one.server.routes.tests.js#L78-L80

Not a blocker for merging. :-)

modules/users/server/routes/users.server.routes.js
gabrielsr: I thought of using it on a screen for unblocking users. We list all blocked users (api return the list of blocked users populated with simplified profiles) and then the user can choose one to remove.

simison: Good thinking 👍  :-)

simison:
One incremental PR towards this could be extracting this check into a utility function:

https://github.com/Trustroots/trustroots/blob/a4770d3128ee84721684b7cb09008bd8c029c66b/modules/users/server/controllers/users.profile.server.controller.js#L857-L874

We actually [have it already twice](https://github.com/Trustroots/trustroots/blob/a4770d3128ee84721684b7cb09008bd8c029c66b/modules/users/server/controllers/users.profile.server.controller.js#L791) in that controller and I'll need it 3rd time [for avatar API](https://github.com/Trustroots/trustroots/pull/1516).

modules/users/server/controllers/users.block.server.controller.js
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Remove a user from the blocked list
+ */
+exports.unblockUser = async function (req, res) {
+  /*
+   * req.profile was instantiated by a prev middleware
+   */
+  if (!req.profile) {
+    return res.status(500).send({
```
simison, line 85: I think we've sent 400 errors ("400 Bad Request") in other API routes instead of 500 ("500 Internal Server Error") — thoughts? I don't have a strong opinion either way.

gabrielsr: Yeah. It's definitely a bad request!

modules/users/server/controllers/users.block.server.controller.js
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/**
+ * Add a new userId to the blocked list
+ */
+exports.blockUser = async function (req, res) {
+  /*
+   * req.profile was instantiated by a prev middleware
+   */
+  if (!req.profile || req.profile._id === req.user._id) {
```
hlibbabii, line 41: Looks like `req.profile._id === req.user._id` is never `true`. So, currently, you can block your own profile:  

https://stackoverflow.com/questions/11637353/comparing-mongoose-id-and-stringsblocking

modules/users/server/controllers/users.profile.server.controller.js
```diff
@@ -854,19 +890,18 @@ exports.userByUsername = function (req, res, next, username) {
...
+              isBannedProfile,
+              isBlocked,
+            } = classifyPermission(req.user, profile);
 
             // Not own profile, and not public, or suspended, or shadowbanned user
             if (
               !isAdminOrModerator &&
               !isOwnProfile &&
-              (!profile.public || isBannedProfile)
+              (!profile.public || isBannedProfile || isBlocked)
```
hlibbabii, line 904: Definitely not a major thing, but maybe something to think of in the future. It took me some time to understand the logic whether a user should be visible or not (I started working with the codebase a few days ago). I wonder if it would make sense to refactor/rename/simplify things here and also in the `userMiniByID` function. I could imagine this could be an easy place for a bug to hide:

- `user` vs `profile` - who is me, who is the other user
- `isAdminOrModerator`, `isBannedProfile` - me or the other user?
- the comment above doesn't help much especially because it can get outdated easily

modules/users/tests/server/user-block.server.routes.tests.js
```diff
@@ -0,0 +1,328 @@
...
+    login(alice.credentials)
+      .catch(checkError('alice list blocked users: login', done))
+      .then(() => block(bob.credentials.username))
+      .catch(checkError('alice list blocked users: block bob', done))
+      .then(resp => {
+        /* alice see bob in blocked list */
+        log('info', `block response ${JSON.stringify(resp.body)}`);
+        agent
+          .get('/api/blocked-users')
+          .expect(200) // not found!
```
hlibbabii, line 202: found now? :)

modules/users/server/controllers/users.block.server.controller.js
simison: Good catch! `_id` is Mongo `ObjectId` which has `equal` method:

https://github.com/Trustroots/trustroots/blob/46dc40602812521a0d6320f190f307eff40af244/modules/messages/server/controllers/messages.server.controller.js#L215-L216

Similarly validation is possible:
https://github.com/Trustroots/trustroots/blob/46dc40602812521a0d6320f190f307eff40af244/modules/messages/server/controllers/messages.server.controller.js#L457-L458

simison (approved):
Smoke-tested locally, all good!

Thanks for all the work on this and thanks for patience with merging. 😅

simison:
> Administrative user should be able to list a user nonetheless
> Hide messages and events in both ways (anything more?)


Valid points! I added as a todos in https://github.com/Trustroots/trustroots/issues/1467

PR #1491, opened 2020-05-25
Docs/add translation guide

rumwerfer:
add i18n guide to volunteer docs


simison:
I'll get this in already as the volunteering page is getting some traffic now and would be great people understand translator skills are needed. We can then modify it in follow-ups.

Issue #1487, opened 2020-05-23
Incorporate i18n changes back from weblate


nicksellen:
(part of [i18n overview](https://github.com/Trustroots/trustroots/issues/1485))

## Original issue

(content lifted from comment by @mrkvon here https://github.com/Trustroots/trustroots/issues/492#issuecomment-615422263)

Currently, we have an issue with our `i18next` config in `hosted.weblate`. See [#1373#discussion_r410402108](https://github.com/Trustroots/trustroots/pull/1373#discussion_r410402108).

We use natural translation keys, so `.` is not a special character, but end of a sentence. For this reason, our `./config/client/i18n.js` says: 

```js
nsSeparator: false,
keySeparator: false,
```
however weblate assumes that we use `.` as key separator, so it does weird things like

```json
    "Wait a moment": {
        "": {
            "": {
                "": "Warte einen Moment ..."
            }
        }
    },
```

I haven't found the related config on the website or in weblate documentation.

## Updates

- there is [no setting for it](https://github.com/Trustroots/trustroots/issues/492#issuecomment-617010572)
- my [hope](https://github.com/Trustroots/trustroots/issues/492#issuecomment-617038857) about using basic JSON, instead of nested JSON, I think is unfounded, as we use `i18next JSON file` format, not `JSON file`

You can see the PR from weblate --> https://github.com/Trustroots/trustroots/pull/1373

nicksellen:
I came up with a short-term hacky solution - fix it with [a script](https://gist.github.com/nicksellen/cad99a57f877148a760ba2a9351c3afc) :) - it just re-nests the JSON data properly and puts the dots back in.

Gunna try it out a bit more first... maybe it will need manually applying on the weblate PR and manually merging it initially, but I'm keen to get the translations active.

nicksellen:
Still doesn't work...

https://hosted.weblate.org/projects/trustroots/core/#repository shows the failure.

It seems very odd to me that they seem to be attempting a _rebase_ instead of a _merge_ (a rebase needs to resolve conflicts for each commit, whereas a merge only needs to resolve the conflicts for the latest commit - of which there are none now, I can cleanly merge weblates repo into ours).

Hmmm...

simison:
There was a successful i18n commit yesterday: https://github.com/Trustroots/trustroots/commit/40d9ff4633b34854cfd4758f30159b4be5338804

— is that a good sign? :-)

Issue #1486, opened 2020-05-23
Extract translations from source


nicksellen:
(this issue is part of [i18n overview](https://github.com/Trustroots/trustroots/issues/1485))

We need a solution (technical + process) for extracting the translation strings from the source into JSON files.

Current situation is we have:
- [babel-plugin-i18next-extract](https://i18next-extract.netlify.app) installed
- it is enabled in our babel config when `NODE_ENV=production`
- ... therefore it runs in the CI build job

... but I'm not sure what happens next, does it get deployed to dev2.trustroots.org?, and what we _want_ to happen.

- [x] research current behaviour
- [x] explore/design/discuss new behaviour (if desired)
- [x] implement new behaviour!

- [slack discussion](https://trustroots.slack.com/archives/C0A3Q15SS/p1586363780076500?thread_ts=1586274524.070900&cid=C0A3Q15SS) that described an idea:
> an idea (which would implement the "do i18n for a PR when label is present" workflow):
> - when a PR build job in travis
> - during the build job (which runs npm run build)
> - after doing the build...
> - check if we have the i18n ready label (via the GitHub API)
> - commit any changes to public/locales back to the repo
> - also setup a filter rule to not run CI on those auto commits... I don't know whether weblate will be able to pull in those things (as it's configured to point to the git repo already), if not, then we might as well just do a similar workflow when the master branch builds are run...

... plus also the simpler alternative to just do that on master branch.

nicksellen:
For comparison here is the process we use with karrot (in the frontend at least):
- developer works on code, and whatever
- we use translations keys (as opposed to natural language), so as the developer works on code, they add the actual source language in english manually in the json file
- once the PR is ready we merge to master
- each night a [CI scheduled job](https://github.com/yunity/karrot-frontend/blob/2de8d64a8aaca09372fd16205ef13264bea2248d/.circleci/config.yml#L369-L374) comes by and syncs translations:
  - push any new keys (from the `en` file(s))
  - pull any updates from transifex
  - commit any changes from pulling to the git repo on an i18n branch and creates/updates a PR
  - we merge the PR when/if happy

I guess advantage of the scheduled job is it batches any changes done in the day in one go.

Given we already build the site in master branch in order to deploy it, the translations _are_ actually getting extracted at that point. We just don't commit them back to the repo. So, I think we can just do that! (but maybe with a `[ci skip]` comment to prevent it triggering the CI again :) )

PR #1373, opened 2020-04-11
Translations update from Weblate

weblate:
Translations update from [Weblate](https://hosted.weblate.org/projects/trustroots/translation/) for Trustroots/translation.

PR #1495, opened 2020-05-26
Script to update locales in CI

nicksellen:
#### Proposed Changes

* runs on master branch in CI
* fixes weblate JSON issues (see #1487)
* extracts new translations keys (see #1486)
* commits and pushes any changes to git repo
* then runs build/deploy as usual

Uses a robot user I created (https://github.com/trustroots-ci) to do the commits.

The workflow for incorporate the weblate translations will then be to just merge the weblate PR whenever we want (even with it's JSON mistakes), and then this would run and fix it all.

#### Testing Instructions

* hard to test! once reviewed visually, I will merge to master and then fiddle until it's working :)

Fixes https://github.com/Trustroots/trustroots/issues/1487 and https://github.com/Trustroots/trustroots/issues/1486


scripts/ci/fix-locales-and-push-changes
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+fix() {
+  ./scripts/ci/fix-weblate public/locales/*/*.json
+  npx prettier --write public/locales/*/*.json
```
simison, line 5: Doesn't this happen in git commit?

scripts/ci/update-locales
```diff
@@ -0,0 +1,45 @@
...
+
+[ci skip]
+EOF
+  fi
+}
+
+push_changes() {
+  if [ ! -z "$TRAVIS_BRANCH" ]; then
+    echo "Pushing changes!"
+    git config --global user.email "trustroots-ci@nicksellen.co.uk" # TODO: move this to @trustroots.org account...
```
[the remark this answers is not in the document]
nicksellen: Is _is_ a real GitHub account, I created it, GitHub call them _machine users_ (see https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users for them in another context). And in this case it's to be able to create a personal access token that only has access to what it needs, if I create one in my own account it has _way_ too much access...

scripts/ci/fix-locales-and-push-changes
nicksellen: Yes, good point, although it doesn't actually work when I tried it here... I just managed to commit:

```json
{
  "Polish":           "Polština",
     "foo": "bar"
}
```
locally... and no fixing occured :/

... digdigdig, `.eslintignore` says to ignore `public`...  we re-use the ignore file from eslint for prettier, _but_ they have different syntaxes/semantics (a bit)...

- https://eslint.org/docs/user-guide/configuring#eslintignore
- https://git-scm.com/docs/gitignore#_pattern_format

working on it...

nicksellen: Fixed in the last few commits... oops also seeing I set my local git credentials to the Trustroots CI user :)

nicksellen:
A big failure so far:
- it runs when it's not even on the master branch
- we can't push to master
- ... it _tries_ to push to master (I guess `$TRAVIS_BRANCH` is not what I expected...)

> $ [ "$TRAVIS_BRANCH" = "master" ] && ./scripts/ci/update-locales
> trustroots@0.3.3 i18n:extract /home/travis/build/Trustroots/trustroots
> babel --config-file ./babel.config.i18n.js -o /dev/null 'modules/**/*.js'
> [detached HEAD 9e18402] [CI] Extract locales
> 40 files changed, 3652 insertions(+), 20 deletions(-)
> Pushing changes!
> remote: error: GH006: Protected branch update failed for refs/heads/master.        
> remote: error: Required status check "continuous-integration/travis-ci" is expected. At least 1 approving review is required by reviewers with write access.        
> To https://github.com/Trustroots/trustroots.git
>  ! [remote rejected] HEAD -> master (protected branch hook declined)
> error: failed to push some refs to 'https://trustroots-ci:[secure]@github.com/Trustroots/trustroots.git'

nicksellen:
Ok, that should all be fixed now, but won't really know until it's merged. So any more reviews welcome :)

.eslintignore
```diff
@@ -3,7 +3,8 @@
 bin/admin
 deploy
 migrations/archived
-public
+public/*
+!public/locales
```
simison, line 7: `!public/**/*.json` could work too, I see there's another json file that we might want to Prettify.

Not a blocker.

simison:
Should we mergee this first? https://github.com/Trustroots/trustroots/pull/1373

nicksellen:
Ok, gunna go for it after the CI finishes... no need to merge #1373 before. It will test the flow better to do it after.

nicksellen:
.... ok it's not very happy, as committing/pushing to master is VERY NAUGHTY and I can't seem to make an exemption. Maybe if I make the bot user an admin...

nicksellen:
Having the CI user as an admin user worked, it's a bit overprivileged but hey ho.

> I wonder if we'd be better off to use GitHub Actions — could simplify this part?

GitHub actions is basically just another CI provider, just built in. Maybe the user rights are easier, but for everything else seems more painful having two CI providers at the same time.

PR #1496, opened 2020-05-28
Add stale issue/pr action

nicksellen:
#### Proposed Changes

* mark our unloved issues and PRs as unloved

#### Testing Instructions

* nothing will happen until it's merged and it runs on it's schedule

Fixes #aloadofoldissuesnobodycaresaboutanymore

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,40 @@
+name: 'Close stale issues'
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v2.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # See https://github.com/actions/stale/blob/master/action.yml for description of the options
+
+          stale-issue-label: unloved
+
+          stale-issue-message: |
+            This issue is marked as unloved because it has not had any activity for 180 days.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,40 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v2.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # See https://github.com/actions/stale/blob/master/action.yml for description of the options
+
+          stale-issue-label: unloved
```
simison, line 16: We have a few labels with [Status] prefix, might work here too?

"[Status] unloved" would be quite a combo of serious and less serious, haha!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,40 @@
...
+
+          stale-issue-message: |
+            This issue is marked as unloved because it has not had any activity for 180 days.
+
+            It doesn't mean it's not important, so please remove the unloved label if you like it, or add a comment saying what it means to you :)
```
simison, line 21: Could add something about "if this was a big, please test to see if this is still an issue?"

.github/workflows/stale.yml
nicksellen: I kinda feel those `[Status]` ones are for humans to mark... _unloved_ is quite short and sweet...

simison: Works for me 👍

PR #1516, opened 2020-06-07
User: add API to retrieve avatar

### LINKED ISSUE #982: Create avatar API endpoint
Currently all the logic for which avatar to show (none/uploaded/fb/gravatar) and which size, happens in the frontend in Angular directive. Would be better to handle that logic at the backend and on frontend we simply call an URL with user id and size.

While doing this, we can create a new react component for the avatar (can be two separate PRs).

Old directive;
https://github.com/Trustroots/trustroots/blob/master/modules/users/client/directives/tr-avatar.client.directive.js

See also convo;
https://github.com/Trustroots/trustroots/pull/965#issuecomment-447267930

simison:
#### Proposed Changes

* Adds an API endpoint to retrieve user avatar easily without frontend logic.

Currently does redirect to the final image, but next steps should likely be:
- Read and stream the image in the serverside instead
- Add caching layer so that we don't transfer gigabytes of images on the internet repeatedly for nothing
- Consider adding caching layer to ensure hit on DB when opening pages with multiple avatars (maybe do some measurements first)

We could probably test this with the mobile app (https://github.com/Trustroots/trustroots-mobile) first before doing so on the web.

#### Testing Instructions

* Upload avatar image to your user
* Open `http://localhost:3000/api/users/[USER_ID]/avatar`
* You should see your image
* Play with `source` and `size` URL args, you can see valid values from code

#### Known problems:
- Users with `public:false` return 404 instead of default avatar

Resolves #982


simison:
If we'd want to proxy responses, these seemed interesting:
- https://github.com/chimurai/http-proxy-middleware
- https://github.com/villadora/express-http-proxy
- https://medium.com/better-programming/how-to-build-a-reverse-proxy-with-express-87a31ec2ec80

modules/users/tests/server/user-avatar.server.routes.tests.js
```diff
@@ -0,0 +1,174 @@
...
+          .get(`/api/users/${facebookUser}/avatar`)
+          .expect(302);
+
+        res.text.should.equal(
+          `Found. Redirecting to https://graph.facebook.com/12345678912345678/picture/?width=1024&height=1024`,
+        );
+      });
+    });
+  });
+});
```
simison, line 174: Missing tests for different size options.

modules/users/tests/server/user-avatar.server.routes.tests.js
```diff
@@ -0,0 +1,174 @@
...
+          .get(`/api/users/5eefc92b4f63d31861f91f57/avatar`)
+          .expect(302);
+
+        res.text.should.equal(defaultRedirect);
+      });
+
+      it('Get 404 response for non public users', async () => {
+        await agent.get(`/api/users/${nonPublicUser}/avatar`).expect(404);
+
+        // @TODO: we should just reurn default image instead of 404
```
simison, line 103: Couldn't understand yet why this is responding 404. We want to always respond with image.

simison: Ah, I think it's from `usersPolicy.isAllowed`

modules/users/server/controllers/users.avatar.server.controller.js
```diff
@@ -0,0 +1,369 @@
...
+    req.profile.roles.includes('shadowban');
+  const isPublicProfile = req.profile.public;
+  const isAdminOrModerator =
+    req.user.roles.includes('moderator') || req.user.roles.includes('admin');
+
+  if (
+    !isAdminOrModerator &&
+    !isOwnProfile &&
+    (!isPublicProfile || isBannedProfile)
+  ) {
```
simison, line 305: These different scenarios could do with more tests.

modules/users/server/controllers/users.avatar.server.controller.js
```diff
@@ -0,0 +1,369 @@
...
+ *
+ * @TODO: serve by streaming instead of redirect and add caching layer.
+ *
+ * @param  {Object} res
+ * @param  {String} url
+ */
+function serveAvatarUrl(res, url) {
+  res
+    .status(302) // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302
+    // .setHeader('Cache-Control', 'public, max-age=0')
```
simison, line 267: @TODO: let's set some cache-time for these

simison: @rastapasta how does caching the image on mobile app work? Should we instruct it with this header from the API? That way we can change it remotely any time.

simison:
With Tests covering this and existing APIs pretty well, I'll go ahead and merge so that we can unblock using this for mobile.

Issue #1079, opened 2019-05-15
Useful admin tools to support our members


simison:
To support our members better and to keep the community safe from abusers and spammers, we need easy to use and secure support tools. (See also Roles #767.)

We now have some, here are more ideas. Feel free to modify this list directly instead of adding comments to the thread:

- [x] Search users by name, email and username
- [x] Observe user profile data, including suspended users
  - [ ] Observe support requests from or related to user
  - [ ] Ability to add notes about profile https://github.com/Trustroots/trustroots/pull/1540
- [x] Read messages between two members (we ask for consent)
- [ ] Observe number of messages/threads the user has written
- [x] Ability to suspend and shadow-ban member easily
  - [x] suspending should also hide the profile
  - [ ] buttons to remove user roles again https://github.com/Trustroots/trustroots/issues/2555
- [ ] Highlight detected spammers in some dashboard #1074 
- [ ] An auto-warning when somebody hits some threshold of negative ratings on their messages #781
- [ ] Ability to send user email & password confirmation emails or extend previously generated keys
- [ ] Masquerading #670 (especially great for debugging rare issues related to database, would also replace some of the other features on the list but at the same time, this is potentially very complex instead of debugging locally with identical database entry)
- [ ] Show count of "messages in the spirit of TR" in member report card (e.g. "17/20")
- [ ] Repair or remove Zendesk search link

Please add comments on what you feel should be higher priority tool to build or if something isn't really needed. Would be good to gather what is "a must" and what is "nice to have".

Note that all the admin tool actions are audit logged.

cc @guaka @doktorhorst @noahsmindfuck

doktorhorst:
That's great and makes support stuff much faster.
I'd say in a lot of report cases, it might be very useful to look into *more* messages sent by the reported member, not only the ones to the reporting member. But idk if that is ok with our privacy approach.

guaka:
Yes. Maybe we should divide this process into several steps.

There have been a couple of cases where the reporter looks like the offender. So we can start by getting the reporter's permission to read their messages, and if we need more we go ahead and also read the other party's messages, possibly only with their permission.

Maybe part of reporting a member is that you agree (with a checkbox) to have the support team read messages you sent to that reported member?

(It's tricky and complicated.)

noahsmindfuck:
[...]

PR #1540, opened 2020-06-20
Admin: add notes feature

simison:
Part of https://github.com/Trustroots/trustroots/issues/1079

#### Proposed Changes

* Add simple user "notes" feature to admin tools; important for e.g. logging info about warnings we give to users or other such info. 
* Notes get saved to separate collection to avoid leaking via users collection
* Added also ability to customize `TrEditor` component's placeholder text: observe that when writing messages to someone, placeholder still says _"Type your text"_. As a bonus, this can now be translated while previously it was coming from the library's depths always in English. :-)
* Only known problem is that written note doesn't get emptied when sending it — don't have time to dig into that now and it can be fixed later.

Screenshot 2020-06-20 at 23 44 07

#### Testing Instructions

- Have admin user
- Open `/admin`
- Find regular user
- Observe that you can write notes in their profile

### DIFF
--- a/modules/admin/client/api/admin-notes.api.js
+++ b/modules/admin/client/api/admin-notes.api.js
@@ -1,0 +1,11 @@
+import axios from 'axios';
+
+export async function addNote({ note, userId }) {
+  const { data } = await axios.post('/api/admin/notes', { note, userId });
+  return data;
+}
+
+export async function listNotes(userId) {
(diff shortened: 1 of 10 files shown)


simison:
Would it be useful to add some kind of tagging here?

simison:
Let's get this in so that support team can use it.

PR #1552, opened 2020-06-24
Add basic websockets server

simison:
Proof of concept of a basic WebSocket server implementation.

The use case would be for the native mobile app to be able to pull via sockets instead of doing constant polling of the API.

#### Proposed Changes

* Work in progress

#### Testing Instructions

* Work in progress

### DIFF
--- a/config/lib/websocket.js
+++ b/config/lib/websocket.js
@@ -1,0 +1,132 @@
+const connectMongo = require('connect-mongo');
+const cookieParser = require('cookie-parser');
+const http = require('http');
+const passport = require('passport');
+const path = require('path');
+const session = require('express-session');
+const WebSocket = require('ws');
+
+const config = require('../config');
+const MongoStore = connectMongo(session);
+
+function authenticateSession(req, connection, callback) {
+  // Create a MongoDB storage object
+  const mongoStore = new MongoStore({
+    collection: config.sessionCollection,
+    mongooseConnection: connection,
+  });
+
+  const cookieParserMiddleware = cookieParser(config.sessionSecret);
+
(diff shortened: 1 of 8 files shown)


modules/pages/client/components/Rules.component.js
```diff
@@ -1,39 +1,15 @@
...
-          </div>
-        </div>
-      </Board>
-
       <section className="container container-spacer">
         <div className="row">
           <div className="col-xs-12 col-sm-offset-1 col-sm-10 col-md-offset-2 col-md-8">
             <div>
-              <RulesText />
+              <WebSocketDemo />
```
simison, line 11: Haha, needs moving to a better place. This was here just for a demo.

config/lib/websocket.js
```diff
@@ -0,0 +1,132 @@
...
+
+          callback();
+        });
+      });
+    });
+  });
+}
+
+module.exports.start = function (app, connection) {
+  const map = new Map();
```
rastapasta, line 72: would recommend using an Array instead to allow multiple connections (like web + app)

simison: Yup 👍 makes sense. Feel free to push changes to this branch!

package.json
```diff
@@ -171,6 +171,7 @@
     "react-map-gl": "5.2.7",
     "react-medium-editor": "1.8.1",
     "react-responsive": "8.1.0",
+    "react-use-websocket": "2.0.3-b.4",
```
simison, line 174: Did a cursory look at different options and nothing seemed to be at very established state. Not too tied to this specific one either.

PR #1586, opened 2020-07-07
Modified the file extension of React files to JSX for clarity

Pramuspl:
#### Proposed Changes

Changed the extensions of files containing JSX syntax as well as fixed some linter warnings.

#### Testing Instructions

It's a minor change, the app doesn't seem to be impacted by it.

Fixes #


simison:
There might be some things to adjust in `config/assets`, `.eslintrc.js` and `package.json`

modules/core/client/components/StepNavigation.jsx
```diff
@@ -36,7 +36,7 @@ const BackButton = ({ small, ...props }) => {
       aria-label={t('Previous section')}
       {...props}
     >
-      <span className="icon-left" aria-hidden="true"></span>
+      <span className="icon-left" aria-hidden="true"/>
```
simison, line 39: I think Prettier/eslint wants these to be ` />` with space in between.

simison:
Must be somewhere still config change missing since this is failing the tests. Does it build?

Pramuspl:
> Must be somewhere still config change missing since this is failing the tests. Does it build?

Exactly, still struggling with it

simison:
Can you see if it helps to add `jsx` into `resolve` config here:

https://github.com/Trustroots/trustroots/blob/c30ce3b8a340537157f813462f77420a1e2a2a47/config/webpack/webpack.config.js#L64

See https://stackoverflow.com/a/36930833

Make sure to check how that works in latest Webpack docs https://webpack.js.org/concepts/

simison:
We'll also need a rebase here. :-)

PR #1607, opened 2020-07-19
Docs: instructions for activating translation

chagai95:
Added how to activate translation.

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -143,5 +143,4 @@ featureFlags: {
   },
 ```
 
-see an example [here](https://github.com/Trustroots/trustroots/blob/c30ce3b8a340537157f813462f77420a1e2a2a47/config/env/development.js#L14)
-
+[See an example.](https://github.com/Trustroots/trustroots/blob/c30ce3b8a340537157f813462f77420a1e2a2a47/config/env/development.js#L14)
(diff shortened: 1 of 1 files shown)


docs/i18n.md
```diff
@@ -131,3 +131,16 @@ import { Trans } from 'react-i18next';
 
 // ...
 ```
+
+
+To activate translation on any branch add the following to the `config/env/local.js` inside the `module.exports = {}` code block:
+
+```
+featureFlags: {
+    reference: true,
```
simison, line 140: Since `references` has nothing to do with i18n, let's not include it here as it can just confuse?

In the sample file there's a good example:

https://github.com/Trustroots/trustroots/blob/c30ce3b8a340537157f813462f77420a1e2a2a47/config/env/local.sample.js#L14-L17

docs/i18n.md
```diff
@@ -131,3 +131,16 @@ import { Trans } from 'react-i18next';
 
 // ...
 ```
+
+
```
simison, line 135: Can you add "Enabling translations" or similar title here? Now this is under "Translation with variables". ;-)

docs/i18n.md
```diff
@@ -131,3 +131,16 @@ import { Trans } from 'react-i18next';
...
+To activate translation on any branch add the following to the `config/env/local.js` inside the `module.exports = {}` code block:
```
hlibbabii, line 136: Maybe "To activate translation _in your local environment ..."_ ?

simison (approved):
I tuned the copy a bit.

🚀 merge time!

PR #1610, opened 2020-08-03
Messages: fix reply button covering text

simison:
#### Proposed Changes

* Adjust CSS so that reply button doesn't cover the message itself

#### Before

Screenshot 2020-08-03 at 23 04 15

#### After
Screenshot 2020-08-03 at 23 03 37

#### Testing Instructions

* Open new message thread
* On mobile width screen...
* Write text so that it otherwise goes behind the button. Switch to this branch and confirm text now flows around the button
* Test that desktop sizes are unaffected

Fixes #

### DIFF
--- a/modules/messages/client/less/thread.less
+++ b/modules/messages/client/less/thread.less
@@ -168,6 +168,7 @@
         margin: 0;
       }
       @media (max-width: @screen-xs-max) {
+        padding-right: 45px;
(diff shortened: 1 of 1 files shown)


simison:
A bit on short for devs currently around so went ahead and merged since this is bugfix and super simple.

PR #1611, opened 2020-08-04
Dynamic team page and move board to foundation page

simison:
### Proposed Changes

* Adds new `volunteer` role
* Adds API which returns everyone in `volunteer` role
* Lists said volunteers at `/team` page in random order: profile image, name, and link to TR profile
* Moves board members from `/team` page to `/foundation` page
* Small adjustments to links and language here and there
* Pulls images from Trustroots instead of using gravatar/stored image files; that's why in screenshots board images are gone; they'll work at the actual site then.
* I'll leave static images in the folder for now and we can clean them up a bit later until they no longer are in search engine memories and such.

### Testing Instructions

* Open `/foundation` and `/team` pages and observe it looks ok on desktop and mobile
* As an admin user, open user via `/admin` and grant them volunteer role

**Current live pages:**
- https://www.trustroots.org/team
- https://www.trustroots.org/foundation

**Here are screenshots of changes:**

### Team page before
old-team-page

### Team page after
new-team-page

### Foundation page before
old-foundation-page

### Foundation page after
new-foundation-page

### DIFF
--- a/modules/admin/client/components/AdminUser.component.js
+++ b/modules/admin/client/components/AdminUser.component.js
@@ -155,6 +155,16 @@ export default class AdminUser extends Component {
                     >
                       Make moderator
                     </button>
+                    <button
+                      className="btn btn-success"
+                      disabled={
+                        user.profile.roles.includes('volunteer') ||
+                        isSettingUserRole
+                      }
+                      onClick={() => this.handleUserRoleChange('volunteer')}
+                    >
+                      Make volunteer
+                    </button>
                   </div>
                 </div>
               </div>
--- a/modules/admin/server/controllers/admin.users.server.controller.js
+++ b/modules/admin/server/controllers/admin.users.server.controller.js
@@ -119,7 +119,9 @@ exports.listUsersByRole = (req, res) => {
   // Allowed roles to query
   if (
     !role ||
-    !['shadowban', 'suspended', 'admin', 'moderator'].includes(role)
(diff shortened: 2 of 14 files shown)


loberto:
What about having the support/safety team visible for anyone having questions about it?

ACTUALLY I don't think it's necessary, unless people think that TR has been hacked, or the email, and will likely emails one of the older members asking about it...

loberto:
more question... do you plan on updating the page with some regularity? or, how will be decided which faces are visible?

simison:
We'd have button in admin panel to add volunteer role to any profile and they'll appear here on this page right away. So we can add absolutely everyone. :-)

Lemoert:
I'd agree with Loberto's first thought: Good to have Support team separate or an initial/ending text saying "If you're looking to contact our Support Team, you can write us at support@trustroots.org."
Except that small detail, looking good!

simison:
@loberto It's a bit of extra work to get two separate lists going on from the API, so I'd prefer to keep it simple now.

Just to validate if someone who is sending emails out really is from Trustroots, this page should already be sufficient without separating who is in which team? (Many of us do kinda many things anyway)

I added this "contact us" bit, does that help for your purpose?

image

Hope this is good, lemme know if I can change any other texts!

I'll already get this out like this and we can iterate.

Lemoert:
Perhaps simply "Contact the support team if you have any questions"? Looking good anyway!

simison:
That sounds better indeed! 👍 I'll get it updated soonish.

PR #1653, opened 2020-08-11
Hide the background image of the footer on the home page on small screens

hlibbabii:
#### Proposed Changes

Hide the background image of the footer on the home page on small devices: white text on such background is not readable.

#### Testing Instructions
Go to the home page on a mobile device and scroll down to the bottom. The should be no background:

On desktop version, the background should be present:

### DIFF
--- a/modules/pages/client/components/Home.component.js
+++ b/modules/pages/client/components/Home.component.js
@@ -294,7 +294,11 @@ export default function Home({ user, isNativeMobileApp, photoCredits }) {
       </Board>
 
       {/* Footer */}
-      <Board className="board-primary board-inset home-footer" names="bokeh">
+      <Board
+        className="board-primary board-inset home-footer"
+        names="bokeh"
+        ignoreSmall={true}
+      >
         <div className="container">
           <div className="row">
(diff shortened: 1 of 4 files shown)


modules/pages/client/components/Home.component.js
```diff
@@ -294,7 +294,11 @@ export default function Home({ user, isNativeMobileApp, photoCredits }) {
       </Board>
 
       {/* Footer */}
-      <Board className="board-primary board-inset home-footer" names="bokeh">
+      <Board
+        className="board-primary board-inset home-footer"
+        names="bokeh"
+        ignoreSmall={true}
```
simison, line 300: Nit; can be shortened to `ignoreSmall` without `={true}`.

Would something like `ignoreBackgroundOnSmallScreen` be clearer? It's mouthful tho... Up to you!

simison:
Oh you made it generic, cool! :-) I'll give this a spin later tonight

modules/pages/client/components/Home.component.js
hlibbabii: Absolutely!

PR #1654, opened 2020-08-12
contacts_plural was in the string

chagai95:
changed to contacts

Fixes heading at contacts page:

image

#### Proposed Changes

* Update English string source

#### Testing Instructions

* see `/contacts`


simison:
Pretty sure I caused this here https://github.com/Trustroots/trustroots/pull/1553/files#diff-14395d742bd33e75b83fc7d3efdd3521R39 …or more accurately Weblate did something funky with strings I didn’t expect.

Issue #1493, opened 2020-05-25
[Overview] Experiences


nicksellen:
This is to summarize current state of references work and the steps we need to complete to get it ready for us.

## A bit of history / context

(pre-history, [a mention](https://ideas.trustroots.org/comparison-of-hospitality-exchange-platforms/#comment-14932) that _we’ve been working on a reference system for a while already._ ... in 2016).

The topic has been around for a long time with a lot of discussion, this is the [original discussion](https://github.com/Trustroots/trustroots/issues/718) created in October 2018. A bunch of the people commenting in that issue are not active any more. I consider it a piece of history by now. Although with some nice discussions on it.

@mrkvon actually implemented most of the references as described in that issue, but got stuck when it needed a bit more design/discussion input. It's been in the code for a while now in an incomplete state, but hidden behind a feature flag (turned off on https://trustroots.org and on on https://dev2.trustroots.org).

At the hackweek (April 2020) we discussed references again, to try and bring it back to life (see [meeting minutes](https://wiki.trustroots.org/en/Meeting_Minutes/2020-04-13_Hackweek_-_references_session) from that session).

We got a bunch of further work done based on those discussions, and some merged. But it stalled again after that ([this slack thread](https://trustroots.slack.com/archives/C08SENA9Z/p1587373767234900) was the last we heard about it).

Another sprint of work is happening in August 2020.

## TODO Non Technichal

- [x] See if https://github.com/Trustroots/trustroots/pull/918 has something to salvage 
- [x] See if https://github.com/Trustroots/trustroots/pull/1389 has something to salvage
- [x] Display user references (`/profile/USERNAME/references`) (overview https://github.com/Trustroots/trustroots/issues/1494)
  - [x] Basic skeleton for listing references (https://github.com/Trustroots/trustroots/pull/1662)
    - [x] Work on improving visuals for the list (@hlibbabii takes a look, we are not sure what this means)
    - [x] Hide "Write one!" nudge when looking at own profile and there are no references (https://github.com/Trustroots/trustroots/pull/1668)
    - [x] remove colors from the top-heading counters (positive/negative/maybe)

- [x] Write references.
    - [x] Basic functionality implemented (`/profile/USERNAME/references/new`)
    - [x] Update questions, Update possible answers to questions (ask non-tech people to help with this) (#1856 )

    - [x] Test if "support" feature (when adding "no" reference) works and if not, implement/fix (the email shows in MailDev)

[...]

PR #1667, opened 2020-08-18
References: include replies

simison:
Resolves API side of https://github.com/Trustroots/trustroots/issues/1686

Part of references list feature https://github.com/Trustroots/trustroots/issues/1493

#### Proposed Changes

* Switch from simple `.find()` to `.aggregate()` pipeline to allow more complex queries

###### TODO:

* Query for reference replies

#### Testing Instructions

* WIP

### DIFF
--- a/modules/references/client/api/references.api.js
+++ b/modules/references/client/api/references.api.js
@@ -60,7 +60,6 @@ export async function create(reference) {
  *
  * @param {string} userFrom - id of user who gave the reference
  * @param {string} userTo - id of user who received the reference
- * @params {bool} includeReplies - true if replies to the references need to be fetched
  * @returns Promise<Reference[]> - array of the found references
  */
 export async function read({ userFrom, userTo }) {
@@ -71,7 +70,6 @@ export async function read({ userFrom, userTo }) {
   if (userTo) {
     params.userTo = userTo;
   }
-
   const { data: references } = await axios.get('/api/references', { params });
   return references.map(reference =>
     mapObjectToObject(reference, referenceMapping, true),
--- a/modules/references/client/api/references.api.js
+++ b/modules/references/client/api/references.api.js
@@ -63,16 +63,13 @@ export async function create(reference) {
  * @params {bool} includeReplies - true if replies to the references need to be fetched
  * @returns Promise<Reference[]> - array of the found references
  */
-export async function read({ userFrom, userTo, includeReplies }) {
+export async function read({ userFrom, userTo }) {
   const params = {};
   if (userFrom) {
     params.userFrom = userFrom;
   }
   if (userTo) {
     params.userTo = userTo;
-  }
-  if (includeReplies) {
-    params.includeReplies = includeReplies;
   }
 
   const { data: references } = await axios.get('/api/references', { params });
--- a/modules/references/client/api/references.api.js
+++ b/modules/references/client/api/references.api.js
@@ -60,9 +60,10 @@ export async function create(reference) {
  *
  * @param {string} userFrom - id of user who gave the reference
  * @param {string} userTo - id of user who received the reference
+ * @params {bool} includeReplies - true if replies to the references need to be fetched
  * @returns Promise<Reference[]> - array of the found references
  */
-export async function read({ userFrom, userTo }) {
+export async function read({ userFrom, userTo, includeReplies }) {
   const params = {};
   if (userFrom) {
     params.userFrom = userFrom;
@@ -70,6 +71,10 @@ export async function read({ userFrom, userTo }) {
   if (userTo) {
     params.userTo = userTo;
   }
+  if (includeReplies) {
+    params.includeReplies = includeReplies;
+  }
+
   const { data: references } = await axios.get('/api/references', { params });
   return references.map(reference =>
     mapObjectToObject(reference, referenceMapping, true),
--- a/modules/references/client/components/CreateReference.component.js
+++ b/modules/references/client/components/CreateReference.component.js
@@ -35,15 +35,16 @@ export default function CreateReference({ userFrom, userTo }) {
 
   useEffect(() => {
     (async () => {
-      const reference = await api.references.read({
+      const references = await api.references.read({
         userFrom: userFrom._id,
         userTo: userTo._id,
-        includeReplies: false,
       });
 
-      if (reference.length === 1) {
-        setIsDuplicate(true);
-      }
+      references.forEach(reference => {
+        if (reference.userFrom.username === userFrom.username) {
+          setIsDuplicate(true);
+        }
+      });
 
       setIsLoading(false);
     })();
--- a/modules/references/client/components/CreateReference.component.js
+++ b/modules/references/client/components/CreateReference.component.js
@@ -38,6 +38,7 @@ export default function CreateReference({ userFrom, userTo }) {
       const reference = await api.references.read({
         userFrom: userFrom._id,
         userTo: userTo._id,
+        includeReplies: false,
       });
 
       if (reference.length === 1) {
--- a/modules/references/client/components/ListReferences.component.js
+++ b/modules/references/client/components/ListReferences.component.js
@@ -37,7 +37,10 @@ export default function ListReferences({ profile, authenticatedUser }) {
       const publicExperiencePairsDict = publicExperiences
         .filter(experience => experience.userTo.username === profile.username)
         .reduce(
-          (a, exp) => ({ ...a, [exp.userFrom.username]: { theirs: exp } }),
+          (a, exp) => ({
+            ...a,
+            [exp.userFrom.username]: { sharedWithUser: exp },
+          }),
           {},
         );
 
@@ -45,18 +48,18 @@ export default function ListReferences({ profile, authenticatedUser }) {
         if (experience.userFrom.username === profile.username) {
           publicExperiencePairsDict[
             experience.userTo.username
-          ].mine = experience;
+          ].writtenByUser = experience;
         }
       });
 
       // Object.values(publicExperiencePairs).forEach(pair => {
-      //   if (pair.mine === undefined) {
-      //     throw Error(`The following experience is public: ${pair.theirs._id}.
+      //   if (pair.writtenByUser === undefined) {
+      //     throw Error(`The following experience is public: ${pair.sharedWithUser._id}.
       //     The reply to it should also exist but it doesn't.`);
       //   }
-      //   if (pair.theirs.userTo.username !== pair.mine.userFrom.username) {
+      //   if (pair.sharedWithUser.userTo.username !== pair.writtenByUser.userFrom.username) {
       //     throw Error(
-      //       `This experience ${pair.mine._id} should be reply to this one ${pair.theirs._id} but it is not`,
+      //       `This experience ${pair.writtenByUser._id} should be reply to this one ${pair.sharedWithUser._id} but it is not`,
       //     );
       //   }
       // });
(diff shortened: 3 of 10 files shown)


modules/references/server/controllers/reference.server.controller.js
```diff
@@ -378,43 +378,87 @@ exports.readMany = async function readMany(req, res, next) {
...
+      matchQuery.userFrom = new mongoose.Types.ObjectId(userFrom);
+    }
+
+    // Filter by userTo
+    if (userTo) {
+      matchQuery.userTo = new mongoose.Types.ObjectId(userTo);
+    }
+
+    // Aggregate projection for User in reference
+    const userKeys = {
```
simison, line 399: This is basically `userProfile.userMiniProfileFields + ' created'`

I first started converting it to this format but then figured it might not be worth if:
- We'd soon use new Avatar API instead of returning most of this from the API
- We'd anyway add created + gender, thus making it custom.

simison:
@hlibbabii would you be interested to take this over? I'm happy to support you by reviewing! :-)

hlibbabii:
> @hlibbabii would you be interested to take this over? I'm happy to support you by reviewing! :-)

@simison, sure, why not

hlibbabii:
Latest screenshot:

image

simison:
@hlibbabii I think we can get this in as long as it technically works. We can work on visuals in follow-up PRs.

This will need a rebase, and I can then test this if it's ready for test+review?

modules/references/client/components/read-references/Recommendation.js
```diff
@@ -7,16 +7,19 @@ export default function Recommendation({
...
   const { t } = useTranslation('references');
 
   return (
     <div>
-      <ul className="list-inline">
-        {hostedMe && <li>{t('Guest.')}</li>}
-        {hostedThem && <li>{t('Host.')}</li>}
-        {met && <li>{t('Met in person.')}</li>}
-      </ul>
+      {isMyExperience && (
```
simison, line 16: Perhaps these we can keep visible regardless?

modules/references/client/components/ListReferences.component.js
```diff
@@ -82,15 +109,42 @@ export default function ListReferences({ profile, authenticatedUser }) {
...
+      </div>
+
+      {experiencePairs.map(experiencePair => (
+        <div key={experiencePair.theirs._id}>
+          <div className="row">
+            <div className="col-xs-12">
+              <Reference reference={experiencePair.theirs} isMine={true} />
+            </div>
+          </div>
+          {experiencePair.mine && (
```
simison, line 127: `mine` term confuses me a bit: I could be person A looking at experiences between B and C, right?

hlibbabii:
> @hlibbabii I think we can get this in as long as it technically works. We can work on visuals in follow-up PRs.
> 
> This will need a rebase, and I can then test this if it's ready for test+review?

Sure, I'll work no this today

modules/references/client/components/ListReferences.component.js
hlibbabii: @simison How do you like `writtenByUser`/`sharedWithUser`?

modules/references/client/api/references.api.js
```diff
@@ -60,16 +60,21 @@ export async function create(reference) {
...
-export async function read({ userFrom, userTo }) {
+export async function read({ userFrom, userTo, includeReplies }) {
   const params = {};
   if (userFrom) {
     params.userFrom = userFrom;
   }
   if (userTo) {
     params.userTo = userTo;
   }
+  if (includeReplies) {
```
simison, line 74: Any situation practically where we wouldn't want replies? If not, we could just always include them?

modules/references/client/components/ListReferences.component.js
```diff
@@ -14,28 +14,61 @@ import ReferencesSection from './read-references/ReferencesSection';
...
+        experience => experience.public,
       );
-
-      const pendingNewestFirst = references.filter(
-        reference => !reference.public,
+      const pendingNewestFirst = experiences.filter(
+        experience => !experience.public,
       );
+
+      /* TODO: compare by ids (couldn't make it work, so I used usernames to compare for now) */
```
simison, line 36: Just leaving a note here to have a look. :-)

modules/references/client/components/ListReferences.component.js
hlibbabii: `incoming`/`outgoing`?

simison: Further ideas: `written`/`received`, `from`/`to`

Too late in the evening to form an opinion. 😆

simison: Just for inspiration I checked how CS does it (in my profile json dump):

Screenshot 2020-12-06 at 23 17 07

hlibbabii:
This is what we currently see if the other side doesn't share an experience (see last two replies):

image

modules/references/client/components/ListReferences.component.js
hlibbabii: When we fetch references, looks like ids are not fetched with `userTo` and `userFrom` objects:

image

Is there anything wrong with $aggregate query and $project thing? (haven't looked much into mongoose yet)

simison:
> This is what we currently see if the other side doesn't share an experience (see last two replies):

Yeah, it's not optimal and we need to work on it but I'd be fine in this PR, and we can follow-up with visual bits.

modules/references/client/api/references.api.js
hlibbabii: I guess there might be potentially a situation where we wouldn't display replies. But even in this case, I can imagine that pre-loading them could be useful.  So, yes, I agree, let's always fetch replies. I think I got blinded by this code where we check if the user already shared an experience:

      const reference = await api.references.read({
        userFrom: userFrom._id,
        userTo: userTo._id,
        includeReplies: false,
      });

      if (reference.length === 1) {
        setIsDuplicate(true);
      }

This is the only case now where we pass `includeReplies=false`, but I think this logic just should be rewritten

hlibbabii:
> Yeah, it's not optimal and we need to work on it but I'd be fine in this PR, and we can follow-up with visual bits.

Yeah, I understand why you want to merge this PR faster. Also, for me, it's not much fun rebasing it again :) I will fix some small things and convert it to non-draft

modules/references/client/api/references.api.js
simison: Yeah — additionally `userFrom _id, userTo: _id`  feels a very explicit condition already and it shouldn't need to include the "reply" with that condition.

For the purposes of the current UI, it even makes sense to just return boolean yes/no but it's also nice to have the whole experience object so that we can include some of it in the UI.

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -375,48 +375,117 @@ exports.readMany = async function readMany(req, res, next) {
...
+      if (userFrom) {
+        matchQueryReply.userTo = new mongoose.Types.ObjectId(userFrom);
       }
+    }
 
-      return query;
-    })();
+    // Aggregate projection for User in reference
+    const userKeys = {
+      id: 1,
```
simison, line 423: This might need to be `_id` instead.

simison:
Yeah, smaller, more focused PRs are easier to deal with and it would block us both to collaborate on the follow-ups without stepping on each other's toes, so to say. :-)

modules/references/client/api/references.api.js
hlibbabii: The only thing I like about having `includeReplies` explicitly set to true is that it makes more clear why replies are returned: `/api/references?userTo=user1&includeReplies=true`.  Alternatively, we should change the route to `/api/referencePairs?user=user1` or something like this

hlibbabii: Btw, I think we don't really need `userFrom` param, we just need some boolean param like `onlyFromAuthUser` to check if there are already pending or published references from the signed in user:

https://github.com/Trustroots/trustroots/pull/1860/files#diff-7b97d7f1b923f70a20dd2d719b6b11324e2a13f048872f41fd1ca4e6b2d86bf3

simison: Fair point! Looks like we already use authenticated user's ID here: 

https://github.com/Trustroots/trustroots/blob/2c11ed4e7e59be349ba7719ee0510e5b7165a619/modules/references/server/controllers/reference.server.controller.js#L159

That `checkDuplicate()` method is currently used just inside `create()`, but perhaps it should be it's own API endpoint used just for checking this? That way the API for listing references doesn't need to be concerned with any of this?

hlibbabii: @simison What do you think about doing it like this: #1910

simison: @hlibbabii I like it! I'm not sure how soon I get to review but I'll try prioritize this.

hlibbabii: Great! After we merge that one, I'll remove `fromTo` and this PR will be ready to be reviewed

modules/references/client/components/CreateReference.component.js
```diff
@@ -35,14 +35,16 @@ export default function CreateReference({ userFrom, userTo }) {
...
         userFrom: userFrom._id,
         userTo: userTo._id,
       });
 
-      if (reference.length === 1) {
-        setIsDuplicate(true);
-      }
+      references.forEach(reference => {
+        if (reference.userFrom.username === userFrom.username) {
+          setIsDuplicate(true);
```
simison, line 45: Should we short-circuit the `forEach` loop here to avoid looping through the rest?

modules/references/client/components/ListReferences.component.js
```diff
@@ -14,28 +14,58 @@ import ReferencesSection from './read-references/ReferencesSection';
...
 export default function ListReferences({ profile, authenticatedUser }) {
   const { t } = useTranslation('references');
-  const [publicReferences, setPublicReferences] = useState([]);
-  const [pendingReferences, setPendingReferences] = useState([]);
+  const [publicExperiencePairs, setPublicExperiencePairs] = useState([]);
+  const [pendingExperiences, setPendingExperiences] = useState([]);
   const [isLoading, setIsLoading] = useState(true);
 
+  function pairUpExperiences(experiences) {
+    /* TODO: compare by ids (couldn't make it work, so I used usernames to compare for now) */
```
simison, line 22: Is there any benefit in doing this at client side? We could do it at the backend already and get API response in a format the client needs it.

Fine for this PR but perhaps something we can revisit in separate PR. :-)

simison:
Rebased

simison (commented):
✅ Can't actually "approve" PR I started. 😅 

Looks good and works!

Let's get this in and iterate on UI and other aspects in follow-up PRs.

hlibbabii (approved):

simison:
Failing server tests still need a look into before merging.

hlibbabii:
> Failing server tests still need a look into before merging.

Right ... I first wanted to understand what the final version of API would look like, and then write the tests. I'll check the tests now since we are going to merge this.

simison:
Another option is to comment out those tests, or mark them as `.skip()` (if I remember the syntax correct), until we settle on final API.

I think my personal preference would still be to have the "reference pairs" formatting happen on the backend rather than at the frontend but we can ponder more on it. :-)

hlibbabii:
> Another option is to comment out those tests, or mark them as `.skip()` (if I remember the syntax correct), until we settle on final API.
> 
> I think my personal preference would still be to have the "reference pairs" formatting happen on the backend rather than at the frontend but we can ponder more on it. :-)

Yeah, I skipped the tests for now. I also agree that we should return experience pairs from the backend. This also resolves the question of whether we need to ask for replies explicitly - if we return pairs, we definitely don't.

simison:
Aight! :-) Feel free to merge once tests pass. 👍 Nice job here!

PR #1668, opened 2020-08-19
Hide 'Write one!' nudge when looking at own profile and there are no experiences

hlibbabii:
Ongoing work on  #1493 : [Overview] references.

When the logged-in user is viewing the profile of another user that doesn't have any experiences, they see the 'write one!' nudge:

image

#### Before
The nudge is also present when the logged-in user is viewing their own page.

#### After 
The nudge is hidden when the logged-in user is viewing their own page:

image


simison (approved):
Looks good and tests well. 👍 

Feel free to merge.

Issue #1494, opened 2020-05-25
Display user references


nicksellen:
(issue part of overview issue https://github.com/Trustroots/trustroots/issues/1493)

There are three bits of information collected in a reference:

1. interactions (met, hosted me, hosted them)
2. recommend/experience (did you feel comfortable) - includes option to report them too
3. (optional) feedback text

Of the three bits of information, 1 and 3 would be publicly visible in some form, and 2 would be private (initially not used for anything, but for future uses, always kept private though).

The API is available for this (I think? need to double check), but it has not been designed yet, so there is some scope for thinking how it should be displayed, you can see this [previous PR](https://github.com/Trustroots/trustroots/pull/918) for some inspiration/code, but note that it does _not_ follow the above approach (i.e. 2 is public there), so don't use it directly.

One idea that came up could be to display 1 in a generalized way (e.g. _“this person has hosted someone within 6 months”_).

_Note: we did get into some confusion about this after the hackweek, I re-read all the discussions and I think this addresses everything I could find - so if you disagree and have a *critical concern* (i.e. blocking) about this approach, do say so, but otherwise remember we can always iterate/improve in the future_

Related discussions:
- https://wiki.trustroots.org/en/Meeting_Minutes/2020-04-13_Hackweek_-_references_session
- https://trustroots.slack.com/archives/C08SENA9Z/p1587373767234900

(If you want to work on this but have not been involved in previous discussions about it, it would be worth pairing up with somebody who has).

noahsmindfuck:
I played around today with experience listing design... I don't know what the status is atm, maybe this can inspire someone or foster some debate. If anyoen wants to join the miro board to experiment: https://miro.com/app/board/o9J_lU8XQ6Q=/

![2021-02-12-193641_1920x1080_scrot](https://user-images.githubusercontent.com/24893086/107808616-2df62880-6d6a-11eb-928f-7cdcbc8f852a.png)

PR #1697, opened 2020-08-29
Experiences: fix sorting by creation date

hlibbabii:
#### Proposed Changes

Fix sorting of experiences when displaying them in a user's profile as a part of ongoing work on #1494, which is a part of #1493.

Currently, the sorting in master is not working correctly:

https://github.com/Trustroots/trustroots/blob/0b4026830735eaa585d7051a93fa1dd206a50ee3/modules/references/client/components/ListReferences.component.js#L28-L30

image

I thought that we most probably would want to do sorting and filtering on the backend (if implementing pagination or just not loading all the experiences), so instead of fixing it on the front-end, I moved it there. Please let me know you think if I'm wrong.

#### After the change:

image

### DIFF
--- a/modules/references/client/api/references.api.js
+++ b/modules/references/client/api/references.api.js
@@ -55,7 +55,9 @@ export async function create(reference) {
 }
 
 /**
- * API request: read references and filter by userFrom and userTo
+ * API request: read references, filter them by userFrom and userTo,
+ * and sort by 'created' field starting from the most recent date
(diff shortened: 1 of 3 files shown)


simison (approved):
Thanks! Feel free to merge when convenient for you. :-)

I appreciate how straightforward change and simple this PR is, as well as clear testing instructions with screenshots.

PR #1731, opened 2020-10-01
added links

NaiveNarwhal:
intallation links for easiest use of the pages highlighted

#### Proposed Changes

*

#### Testing Instructions

*

Fixes #


simison:
Thanks @NaiveNarwhal ! Could the NodeJS link point to `/en` instead of `/es`? Otherwise looks good!

PR #1739, opened 2020-10-26
Fix merge conflicts at weblate repo

hlibbabii:
Ongoing work on #1487 (Incorporate i18n changes back from weblate)

I thought why wouldn't we try merging Weblate changes and see if it can pull the changes without conflicts this time. If it fails, we could see the errors again and understand more about the problem


simison (approved):
Lets see how this works!

PR #1798, opened 2020-11-05
tribes -> circles: rename i18n file

hlibbabii:


simison (approved):
I like how focused this PR is, changing only what is essential. 👍

PR #1800, opened 2020-11-05
tribes -> circles: rename i18n files owned by weblate

hlibbabii:
We need to rename the files owned by weblate (`public/locales/*/tribes.json`) manually in order not to lose existing translations. Then weblate will pick up the renamed files


simison (approved):
Sounds good!

PR #1803, opened 2020-11-06
add most popular languages to weblate for translation

hlibbabii:
Adding most popular languages taken from here: https://grafana.trustroots.org/d/H-JQpyRMk/profile-details?orgId=1&from=now-30d&to=now

This could be also done using the weblate UI. However:

1) this will create a commit for each language and each component -> 274 commits
2) languages have to be added manually, which is slow, I wrote a bash script which creates files instead.

In the future, if there is some language missing, a single language can also be added manually using the weblate UI.

Related to #1487


simison (approved):
Thanks!

> languages have to be added manually, which is slow, I wrote a bash script which creates files instead.

Feel free to add the script to the `scripts` folder in case it comes in handy in the future!

hlibbabii:
> Feel free to add the script to the `scripts` folder in case it comes in handy in the future!

Yep, let me make it a bit more usable and I'll add it a bit later with another PR

PR #1828, opened 2020-11-14
remove 'nb_NO' locale ('nb' already exists)

hlibbabii:
Also, weblate complains that there is a duplicate language.

This is a newly added locale (that I shouldn't have added) with no translations


simison (approved):
👍 

Just to understand right, this is Norwegian (Bokmål), and we already had Norwegian (Nynorsk)? Ideally, we could support multiple dialects of the same language and multiple languages within one country, so it's odd to me we have to do this?

hlibbabii:
As far as I can see, Weblate recognizes 'nb' (sorry, I mistakenly wrote 'no' before) as Norwegian (Bokmål). So we already have it, and there are quite some translations. Weblate also recognizes 'nb_No' as Norwegian (Bokmål), therefore it complains.  

We can definitely add also Nynorsk, which has the 'nn' code.

PR #1857, opened 2020-11-21
adding gender to references

chagai95:
#### Proposed Changes

* Adding gender to references

#### Testing Instructions

* Go to references on someone's profile Ruth references - dev2 or local build with generated users

Fixes #

### DIFF
--- a/modules/references/client/components/read-references/Reference.js
+++ b/modules/references/client/components/read-references/Reference.js
@@ -9,7 +9,7 @@ import Avatar from '@/modules/users/client/components/Avatar.component';
 import TimeAgo from '@/modules/core/client/components/TimeAgo';
 import UserLink from '@/modules/users/client/components/UserLink';
 import Recommendation from './Recommendation';
-
+import { getGender } from '@/modules/core/client/utils/user_info';
 // @TODO, pull from config
 const DAYS_TO_REPLY = 14;
 
@@ -44,25 +44,6 @@ export default function Reference({ reference }) {
     userTo,
   } = reference;
 
-  /*
-   * Functions passing strings to translation fuction for translation scripts
-   * TODO this method should be extracted for reuse in References for example.
-   */
-  const getGender = genderCode => {
-    switch (genderCode) {
-      case 'female':
-        return t('female');
-      case 'male':
-        return t('male');
-      case 'non-binary':
-        return t('non-binary');
-      case 'other':
-        return t('other');
-      default:
-        return undefined;
-    }
-  };
-
   const created = new Date(reference.created);
 
   const getDaysLeft = created =>
@@ -84,9 +65,7 @@ export default function Reference({ reference }) {
               {t('Member since {{memberSince}}', {
                 memberSince: new Date(userFrom.created).getFullYear(),
               })}
-              {getGender(userFrom.gender) === undefined
-                ? ''
-                : ', ' + getGender(userFrom.gender)}
+              {userFrom.gender && `, ${getGender(userFrom.gender)}`}
             </span>
           </UserMeta>
           <time dateTime={created} className="text-color-links">
--- a/modules/references/client/components/read-references/Reference.js
+++ b/modules/references/client/components/read-references/Reference.js
@@ -44,6 +44,25 @@ export default function Reference({ reference }) {
     userTo,
   } = reference;
 
+  /*
+   * Functions passing strings to translation fuction for translation scripts
+   * TODO this method should be extracted for reuse in References for example.
+   */
+  const getGender = genderCode => {
+    switch (genderCode) {
+      case 'female':
+        return t('female');
+      case 'male':
+        return t('male');
+      case 'non-binary':
+        return t('non-binary');
+      case 'other':
+        return t('other');
+      default:
+        return undefined;
+    }
+  };
+
   const created = new Date(reference.created);
 
   const getDaysLeft = created =>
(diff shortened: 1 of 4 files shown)


modules/references/client/components/read-references/Reference.js
```diff
@@ -65,6 +84,9 @@ export default function Reference({ reference }) {
               {t('Member since {{memberSince}}', {
                 memberSince: new Date(userFrom.created).getFullYear(),
               })}
+              {getGender(userFrom.gender) === undefined
```
simison, line 87: Shorthand would work nicely here:
```
{userFrom.gender && getGender(...)}
```

simison (commented):
Thanks for working on it!

Left a couple small notes that would be good to do.

Couldn't test (I'm on mobile) so will leave approving for someone else.

chagai95:
Thanks for taking a look I'll try adding that as well tomorrow!

modules/references/client/components/read-references/Reference.js
chagai95: I tried this but I'm not quite sure what you mean... we checked to see if the gender is undefined because if it is not we wanted to leave the comma out. Do you mean we should solve this differently?

modules/users/client/components/ProfileViewBasics.js
```diff
@@ -14,6 +14,7 @@ export default function ProfileViewBasics({ profile }) {
 
   /*
    * Functions passing strings to translation fuction for translation scripts
+   * TODO this method should be extracted for reuse in References for example.
```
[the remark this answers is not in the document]
chagai95: that would still be in the users module so references would have to depend on users module? kind of makes sense, are you sure?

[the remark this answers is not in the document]
simison: You could also put it in core module if you prefer. As long as there isn't duplicate code around. :-)

modules/references/client/components/read-references/Reference.js
simison: I meant that `userFrom.gender` will already be undefined without passing it through function, and its enough to check for it being falsy.

Since the alternative ternary output is just an empty string, I think I'd do something like this:

```
{{userFrom.gender && `${getGender(userFrom.gender)}, `}}
```

Returning undefined in JSX will just render nothing.

Does that make sense?

It's not a biggie but reads nicer to me. :-)

simison: I meant that `userFrom.gender` will already be undefined without passing it through function, and its enough to check for it being falsy.

Since the alternative ternary output is just an empty string, I think I'd do something like this:

```
{{userFrom.gender && `${getGender(userFrom.gender)}, `}}
```

Returning undefined in JSX will just render nothing.

Does that make sense?

It's not a biggie but reads nicer to me. :-)

simison (commented):
Left a note.

modules/references/client/components/read-references/Reference.js
chagai95: it reads a lot nicer! thanks, I'm learning loads!

chagai95:
@hlibbabii can you review so we can merge?

modules/references/client/components/read-references/Reference.js
```diff
@@ -65,6 +65,7 @@ export default function Reference({ reference }) {
               {t('Member since {{memberSince}}', {
                 memberSince: new Date(userFrom.created).getFullYear(),
               })}
+              {userFrom.gender && `, ${getGender(userFrom.gender)}`}
```
simison, line 68: "Other" gender label will probably read a bit funny and be confusing here, but ok to figure it out some time later.

simison (approved):
Tests well!

Tested both when gender was not set, and when gender was set. 

Thanks for working on it.

modules/references/client/components/read-references/Reference.js
chagai95: I think it will look weird on the profile as well, maybe we can add that to the string of other as a work around until we find something better? gender: Other?

simison: Yes, agreed. How about just "other gender"?

There might be some translation-difficulties with partial sentences with commas where translators don't get the whole sentence to see. Can that be improved somehow too?

chagai95: other gender is good, I'll push that until we find a better solution.

MrSalami: Hey how about having a gender icon/symbol and then it can be defined in a pop up bubble after hovering over the symbol?

simison: Thanks for the idea!

The trouble with hover is that it doesn't work on mobile and is really basically hiding info until folks realize to hover, so it's best to not rely on it.

Feels like it can be just fine solved with text so let's not complicate it? :-)

chagai95: how about a symbol and text, without hovering, then it'll be clear it's about gender?

MrSalami: Agree, Chagai was mentioning something about hovering, but definitely don't want to complicate anything further and yes, it won't work with mobile. As a solution, a text symbol could be added such as this, which is what we were suggesting really. Check the ones here: https://gender.wikia.org/wiki/Gender_Symbol

PR #1860, opened 2020-11-22
Make pending experiences you shared with someone visible for you in their profile

### LINKED ISSUE #1842: Show the user which experiences shared by them are not yet public
Ongoing work on #1493 

We need to distinguish on the UI the experiences that are already visible for everyone, and the ones that are for now only visible for those who wrote them (Because the other side has not shared theirs yet). Also, the time after which the experience will become public the latest should be shown.

Implementation options: put non-public experiences on the top of the list + use a different color (grey?)

hlibbabii:
# Proposed changes:

image

# Testing instructions:

- go to the profile of any user with which you haven't yet shared experience. Write an experience. An experience should appear under "Experiences pending publishing"  

WIP on #1493 
Fixes #1842

### DIFF
--- a/modules/references/client/components/read-references/Reference.js
+++ b/modules/references/client/components/read-references/Reference.js
@@ -116,7 +116,7 @@ export default function Reference({ experience }) {
             <TimeAgo date={createdDate} />
           </ReferenceLink>
         </Header>
-        {!isPublicReference && (
+        {!isPublicReference && feedbackPublic === undefined && (
           <>
             <PendingNotice>
               <PendingNoticePlaceholder aria-hidden="true">
@@ -159,6 +159,16 @@ export default function Reference({ experience }) {
               <FeedbackPublic>{response.feedbackPublic}</FeedbackPublic>
             )}
           </Response>
+        )}
+        {!isPublicReference && feedbackPublic !== undefined && (
+          <>
+            <PendingNotice>
+              {t(
+                'Your experience will become public in {{count}} days, or when they share their experience with you.',
+                { count: getDaysLeft(createdDate) },
+              )}
+            </PendingNotice>
+          </>
         )}
       </div>
     </div>
--- a/modules/references/server/controllers/reference.server.controller.js
+++ b/modules/references/server/controllers/reference.server.controller.js
@@ -422,13 +422,23 @@ exports.readMany = async function readMany(req, res, next) {
     const selfId = req.user._id;
 
     const userToId = new mongoose.Types.ObjectId(userTo);
-    const matchQuery = {
+    let matchQuery = {
       $or: [{ userTo: userToId }, { userFrom: userToId }],
     };
     // Allow non-public references only when userTo is self
     if (!selfId.equals(userToId)) {
       matchQuery.public = true;
     }
+
+    const privateFromSelfQuery = {
+      userFrom: selfId,
+      userTo: userToId,
+      public: false,
+    };
+
+    matchQuery = {
+      $or: [matchQuery, privateFromSelfQuery],
+    };
 
     // Aggregate projection for User in reference
     const userKeys = {
--- a/modules/references/tests/server/reference-read-many.server.routes.tests.js
+++ b/modules/references/tests/server/reference-read-many.server.routes.tests.js
@@ -210,8 +210,8 @@ describe('Read references by userTo Id', () => {
       should(body[0].userFrom._id).eql(users[0].id);
       should(body[0].userTo._id).eql(users[3].id);
 
-      should(body[0].response).eql(null);
-      should(body[0].public).eql(false);
+      should(body[0].response).be.null();
+      should(body[0].public).be.false();
 
       should(body[0]).have.properties(
         '_id',
--- a/modules/references/tests/server/reference-read-many.server.routes.tests.js
+++ b/modules/references/tests/server/reference-read-many.server.routes.tests.js
@@ -202,6 +202,25 @@ describe('Read references by userTo Id', () => {
       should(body[2].userFrom._id).eql(users[5].id);
     });
 
+    it('[param userTo] response should contain private experience from self', async () => {
+      const { body } = await agent
+        .get(`/api/experiences?userTo=${users[3]._id}`)
+        .expect(200);
+
+      should(body[0].userFrom._id).eql(users[0].id);
+      should(body[0].userTo._id).eql(users[3].id);
+
+      should(body[0].response).eql(null);
+      should(body[0].public).eql(false);
+
+      should(body[0]).have.properties(
+        '_id',
+        'created',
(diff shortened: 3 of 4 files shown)


modules/references/server/controllers/reference.server.controller.js
```diff
@@ -375,41 +375,48 @@ exports.readMany = async function readMany(req, res, next) {
     // validate the query
     validate(validateReadMany, req);
 
-    const { userFrom, userTo } = req.query;
+    const { userTo, userFromPublic, userFromPending } = req.query;
```
simison, line 378: Can the `userFromPending` be inferred from authenticated user `(available always as `req.user._id`), and it thus wouldn't need to be passed through from client layers?

simison:
This seems like an excellent candidate for writing up a few API tests first and then coding to match expected results from API, but totally up to you of course. :-)

modules/references/server/controllers/reference.server.controller.js
simison: To elaborate more on this; we should surface pending references only to people who wrote them, i.e. the authenticated user.

hlibbabii: > Could the userFromPending be inferred from the authenticated user (available always as req.user._id), and it thus wouldn't need to be passed through from client layers?

Right, very valid point.

In general, there so many changes that depend on the change of the experience fetching API. I'll look thoroughly look into the question of changing API in the context of all tasks.

simison:
Needs a rebase. :-)

hlibbabii:
> Needs a rebase. :-)

rather total reimplementation :)

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -404,14 +404,24 @@ exports.readMany = async function readMany(req, res, next) {
...
 
+    const privateFromSelfQuery = {
+      userFrom: selfId,
+      userTo: userToId,
+      public: false,
+    };
+
+    matchQuery = {
+      $or: [matchQuery, privateFromSelfQuery],
+    };
```
simison, line 423: I find the `matchquery` reshuffling a bit hard to grok.

How about something more simple, like so:
```js
    const matchQuery = {
      $or: [
        // Experiences to/from this member
        {
          $or: [{ userTo: userToId }, { userFrom: userToId }],
        },
        // Authenticated user's pending references to this member
        {
          userFrom: selfId,
          userTo: userToId,
          public: false,
        },
      ],
    };

    // Allow non-public references only when userTo is self
    if (!selfId.equals(userToId)) {
      matchQuery.public = true;
    }
```

What do you think?

modules/references/tests/server/reference-read-many.server.routes.tests.js
```diff
@@ -202,6 +202,25 @@ describe('Read references by userTo Id', () => {
...
+    it('[param userTo] response should contain private experience from self', async () => {
+      const { body } = await agent
+        .get(`/api/references?userTo=${users[3]._id}`)
+        .expect(200);
+
+      should(body[0].userFrom._id).eql(users[0].id);
+      should(body[0].userTo._id).eql(users[3].id);
+
+      should(body[0].response).eql(null);
+      should(body[0].public).eql(false);
```
simison, line 214: Nitpick; this works, and alternative would be `.be.null()` and `.be.false()`

modules/references/client/components/read-references/Reference.js
```diff
@@ -116,7 +116,7 @@ export default function Reference({ experience }) {
             <TimeAgo date={createdDate} />
           </ReferenceLink>
         </Header>
-        {!isPublicReference && (
+        {!isPublicReference && feedbackPublic === undefined && (
```
simison, line 119: Checking for `feedbackPublic` being undefined for this scenario feels a bit fragile, but I also don't have better suggestions. 🤔

simison (approved):
Tests well.

I left a couple smaller comments about code but don't let those stop from merging.

CI is still acting up, I'll add you as github space owner so you can force-merge if need.

modules/references/server/controllers/reference.server.controller.js
hlibbabii: I agree. Should we, though, put all the queries inside one `$or`?

modules/references/client/components/read-references/Reference.js
hlibbabii: I don't' like it either. It feels like the `Reference` component needs some parameter from the outside like `feedbackVisible` or `feedbackObfuscated`, or instead of "feedback" use some other term that would summarize textual feedback, recommendation, and interactions.

hlibbabii: In general, I think the root cause of such robustness issues is that, in our `Experience` object, there is this group of keys (`recommend`, `feedbackPublic`, `interactions`) that conceptually belong together: 1) they are all created by the user directly when creating experience 2) they all represent the opinion of the user 3) they all need to be obfuscated when returning from the back-end if the experience is still private.  If there were some `feedback` key, e.g.:
```
experience = {
    _id,
    userFrom,
    userTo,
    created,
    feedback: {
         textualFeedbackPublic,
         interactions,
         recommend,
    },
    responseCreated,
    responseFeedback: {
         textualFeedbackPublic,
         interactions,
         recommend,
    }
}
```
..., the back-end wouldn't return the whole `feedback` object. Would comparing `feedback` with `undefined` make things look more robust?

modules/references/server/controllers/reference.server.controller.js
simison: Yeah, I think that would work. 👍

simison: Not sure where exactly that `matchQuery.public = true;` would need to be though, inside first block? 🤔

hlibbabii: How about doing it like this:

```
function buildReadManyQuery(userToId, selfId) {
  const allExperiencesFromOrToUser = {
    $or: [{ userTo: userToId }, { userFrom: userToId }],
  };

  const privacyRestrictions = {
    $or: [
      // all published experiences:
      { public: true },
      // pending experiences only from or to Authenticated user:
      { public: false, userTo: selfId },
      { public: false, userFrom: selfId },
    ],
  };

  return {
    $and: [allExperiencesFromOrToUser, privacyRestrictions],
  };
}

```

simison:
@hlibbabii good to merge? :-)

hlibbabii:
> @hlibbabii good to merge? :-)

@sisimon, sure! Should we think more about the issues above and fix them if necessary in the follow-up PRs?

simison:
@hlibbabii I'm always fan of doing follow-ups to avoid stalling the progress. ;-)

PR #1861, opened 2020-11-22
Feedback question changes depending on recommend state

chagai95:
#### Proposed Changes

* Feedback question changes depending on recommend state.

#### Testing Instructions

* Choose yes, no, I don't know or no with reporting and then next, you should see a slightly different question every time for the Feedback.

Fixes #


simison:
Looks like you have commits from another branch to add gender?

chagai95:
> Looks like you have commits from another branch to add gender?

My bad, I'll try fixing that...

chagai95:
I think I fixed that, did some weird stuff with git, sorry

modules/references/client/components/create-reference/Feedback.js
```diff
@@ -43,5 +68,7 @@ export default function Feedback({ feedback, onChangeFeedback }) {
 
 Feedback.propTypes = {
   feedback: PropTypes.string.isRequired,
+  recommend: PropTypes.string,
```
hlibbabii, line 71: maybe `recommend` could be a boolean with "unknown" state being `undefined` because it feels like recommendation has boolean semantics at the moment with "don't know" being just an absence of value

but I guess it's also fine to leave it as it is.

modules/references/client/components/create-reference/Feedback.js
```diff
@@ -2,9 +2,34 @@ import PropTypes from 'prop-types';
...
+      case 'yes':
+        return t('Did you enjoy their cooking? singing?');
+      case 'no':
+        return report
+          ? t('Did you hate their cooking? singing?')
+          : t('Did you not enjoy their cooking? singing?');
+      case 'unknown':
+        return t('Did you maybe enjoy their cooking? singing?');
+      default:
+        return undefined;
```
hlibbabii, line 29: if it were not JS, I would throw an exception from the default case but I'm not sure if it's the best practice in JS

modules/references/client/components/create-reference/Feedback.js
chagai95: we discussed changing this in the future if I remember correctly.

modules/references/client/components/create-reference/Feedback.js
chagai95: I think an exception might be dangerous, if it's not caught and I don't understand enough to know that...

hlibbabii (approved):
I tested it, works fine. I guess can be merged!

PR #1875, opened 2020-11-28
Add email confirmation tracking

simison:
### TODO:

- [x] Add datestamp to DB
  - directly into the email token hash? 👎 ideal, but too much work at this point.
  - as a separate field 👍 
- [ ] calculate time between confirm / sending email
  - It's there but isn't working, `.diff()` always gives `0` :-( https://momentjs.com/docs/#/displaying/difference/
- send tracking event to InfluxDB: 
  - [ ] failure
  - [x] success
- InfluxDB tags: do they need to be strings or are booleans ok? Suspecting they need to be strings.

#### Proposed Changes

* Adds tracking for email confirmation, mainly to track time between the email sent and email confirmed

#### Testing Instructions

* Enable influx DB from `env/config/local.js` (best if you set it up locally so you can actually see it working, otherwise you see just error https://www.influxdata.com/)
* Change email from the account page
* Receive email http://localhost:1080/
* Check from DB that user has new `emailTokenDate`
* Confirm your email
* Check from DB that user no longer has `emailTokenDate`
* Check console

**Test:**
- [ ] signup, 
- [ ] email change, and 
- [ ] email change resend

### DIFF
--- a/modules/users/server/controllers/users.authentication.server.controller.js
+++ b/modules/users/server/controllers/users.authentication.server.controller.js
@@ -81,6 +81,7 @@ exports.signup = function (req, res) {
         // (from where it's then again moved to email field)
         user.emailTemporary = user.email;
 
+        user.emailTokenDate = Date.now();
         user.emailToken = authenticationService.generateEmailToken(user, salt);
 
         // Then save the user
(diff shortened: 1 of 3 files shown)


modules/users/server/controllers/users.authentication.server.controller.js
```diff
@@ -650,14 +651,24 @@ exports.extendFBAccessToken = function (shortAccessToken, callback) {
...
     },
     function (err, user) {
       if (!user) {
+        // Add failure tracking
         return res.redirect('/confirm-email-invalid');
       }
+      // Success tracking here
+
+      const isSignup = !user.public;
+      const timeFromEmail = moment().diff(user.emailTokenDate);
```
simison, line 667: Could be interesting to check if `user.emailTokenDate` is Date object, or does it need to be converted into one with `new Date(user.emailTokenDate)`

chagai95:
It's there but isn't working, .diff() always gives 0 

I added a moment() to the date, not sure if that matters but I found other places it was like this.

I also found that the emailTokenDate is being unset from the database before it is being used that's why it was showing 0.

Can I just leave it in the database for now?

modules/users/server/controllers/users.authentication.server.controller.js
chagai95: I'm pretty sure it's working now with moment().diff(moment(user.emailTokenDate))

PR #1878, opened 2020-11-28
Rewrite "user not found" message in React

simison:
A super simple example for rewriting bit of Angular code into React.

#### Proposed Changes

* Rewrite "user not found" Angular template part to a simple React component
* While at it, change the copy to be more meaningful
* Add additional "find people" link instead of only map search, which should be more useful next action for these folks.

#### Testing Instructions

* Open http://localhost:3000/profile/doesnotexist
* ✨ 

**Before:**

Screenshot 2020-11-28 at 20 08 11

**After:**

image

### DIFF
--- a/modules/users/client/components/UserDoesNotExist.component.js
+++ b/modules/users/client/components/UserDoesNotExist.component.js
@@ -18,9 +18,7 @@ export default function UserDoesNotExist() {
       </a>
       <br />
       <br />
-      <a href="/search" className="btn btn-secondary">
-        {t('Map search')}
-      </a>
+      <a href="/search">{t('Map search')}</a>
     </div>
   );
 }
--- a/modules/users/client/components/UserDoesNotExist.component.js
+++ b/modules/users/client/components/UserDoesNotExist.component.js
@@ -16,6 +16,11 @@ export default function UserDoesNotExist() {
       <a href="/search/members" className="btn btn-primary">
         {t('Find people')}
       </a>
+      <br />
+      <br />
+      <a href="/search" className="btn btn-secondary">
+        {t('Map search')}
+      </a>
     </div>
   );
 }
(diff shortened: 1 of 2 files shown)


simison:
While this is so simple that it might not seem worth the effort — the great thing is that we get these bits for translation. ✨

chagai95 (approved):
I did not understand how the components are linked,

is UserDoesNotExist automatically translate to user-does-not-exist?

modules/users/client/components/UserDoesNotExist.component.js
```diff
@@ -9,7 +9,7 @@ export default function UserDoesNotExist() {
     <div className="row text-center" role="alert">
       <h1>{t('Oops!')}</h1>
       <em className="lead">
-        {t('Member you are looking for is not available.')}
+        {t('The Trustrooter you are looking for is not available.')}
```
simison, line 12: @chagai95 thanks for suggestion! What do you think if we'd use "Trustroots member" instead? It would be a dozen times easier to translate to different languages and little less confusing for those who are bad in English.

simison:
> is UserDoesNotExist automatically translate to user-does-not-exist?

Yup! Any file with `.component.js` on the filename is actually turned into Angular directive, so that it can be included in Angular templates (stuff under `/views/*`).

More details here: https://team.trustroots.org/React.html

modules/users/client/components/UserDoesNotExist.component.js
chagai95: Sounds great! It still needs the "the" though otherwise it sounds a bit weird in English... The Trustroot's member or maybe even just a simple "the person"

simison: 👍 Sure! I'll update and merge.

hlibbabii:
When I go to the profile of an existing user, I see this:

image

simison:
Oh, funny! I'll just revert this one and do another round.

PR #1880, opened 2020-11-29
Hide "public" title unless also "pending" experiences are showing

simison:
## Proposed Changes

* Split references section into its own component for clarity
* Change copy of pending/public for clarity
* Don't show "public" title unless we show also "pending" title

## Testing Instructions

* Open your profile with references
* Note no "public" title visible
* Open another profile in incognito and write an experience about the 1st profile
* Refresh the 1st profile's references. Do titles make more sense now?

### Before

"Public" title is always visible. Confusing without context.

image

### After

**New copy for section titles**

Screenshot 2020-11-29 at 20 19 36

**No titles visible if there are no pending experiences**

Screenshot 2020-11-29 at 20 19 21

### DIFF
--- a/modules/references/client/components/ListReferences.component.js
+++ b/modules/references/client/components/ListReferences.component.js
@@ -50,11 +50,11 @@ export default function ListReferences({ profile, authenticatedUser }) {
     return <LoadingIndicator />;
   }
 
-  const hasPublicReferencs = publicReferences.length > 0;
+  const hasPublicReferences = publicReferences.length > 0;
   const hasPendingReferences = pendingReferences.length > 0;
 
   // No references
-  if (!hasPendingReferences && !hasPublicReferencs) {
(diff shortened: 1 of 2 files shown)


modules/references/client/components/ListReferences.component.js
```diff
@@ -50,8 +50,11 @@ export default function ListReferences({ profile, authenticatedUser }) {
     return <LoadingIndicator />;
   }
 
+  const hasPublicReferencs = publicReferences.length > 0;
```
hlibbabii, line 53: typo

hlibbabii (approved):
Tested it, works great!

modules/references/client/components/ListReferences.component.js
simison: Sharp eye!

PR #1896, opened 2020-12-05
Update: refactor experiences time

simison:
#### Proposed Changes

* Small refactor of `TimeAgo` component to be based on [`<time>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) HTML element: just semantics...

    time tag

* Simplify HTML and improve styling for the time; works better especially on mobile now when either name, or translated strings might be longer than in English.

     No changes on desktop version:
    Screenshot 2020-12-05 at 18 27 18

    Mobile version before:

    Screenshot 2020-12-05 at 18 27 14

    Mobile version after:

    Screenshot 2020-12-05 at 18 27 14

#### Testing Instructions

* Check References list and "time ago" bit in the corner
* Confirm other places continue unchanged where `TimeAgo` component is used: message list most prominently.


hlibbabii (approved):
Looks good

PR #1901, opened 2020-12-05
Profile & Experiences: simplify gender label

simison:
Instead of trying to handle comma and lowercase/Sentence-case in different scenarios, I figured we can simplify this a lot and just use natural sentences. Should be easier to translate, too.

#### Proposed Changes

* Change gender labels to Sentence case from lowercase
* Instead of structuring sentences with a comma, split to separate sentences.
* Swap "member since" and "gender" the other way around; makes it easier to skim the list of experiences to stop if they're all from the same gender.

#### Testing Instructions

* Check profile with and without gender or birthdate
* Check experiences list

### DIFF
--- a/modules/core/client/utils/user_info.js
+++ b/modules/core/client/utils/user_info.js
@@ -1,20 +1,20 @@
 import { useTranslation } from 'react-i18next';
 
-/*
- * Functions passing strings to translation fuction for translation scripts
+/**
+ * Get label for profile gender options
  */
 export function getGender(genderCode) {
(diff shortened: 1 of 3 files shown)


hlibbabii:
Does the dot look alright if we specify gender but no date of birth? I am totally fine with having it like it is now, just a random thought

simison:
Yeah, I agreed it could do without it but it's not a biggie either way to me. I can have a look at other sites to see how they do it.

hlibbabii (approved):

PR #1902, opened 2020-12-05
Reference list: clearer meta labels & better mobile support

simison:
#### Proposed Changes

- Pull recommendations and met/guest/host info as "labels" instead of having them as plain text
- Add `max-width: 600px` to the description; keeps it a bit more balanced and easier to read.
- Move `DAYS_TO_REPLY` constant to shared file to avoid duplication

**Before**

image

**After**

image

#### Testing Instructions

* Try different variations of references


simison:
Rebased, good for final look.

modules/references/client/components/read-references/Meta.js
```diff
@@ -0,0 +1,42 @@
...
+        recommend === 'no' && t('Not recommend'),
+        hostedMe && t('Guest'),
+        hostedThem && t('Host'),
+        met && t('Met in person'),
+      ]
+        .filter(label => !!label)
+        .map(label => (
+          <span key={label} className="label label-default">
+            {label}
+          </span>
```
hlibbabii, line 31: is this case too extreme? :D

<image src="https://user-images.githubusercontent.com/2955794/101995492-717cd980-3cca-11eb-806d-527f0a3cb640.png" width="300px"/>

simison: Haha, not at all. :-D I'm actually not a big fan of just a list of labels like this, but it was a quick solution. I'm still thinking about how to present them better. Any ideas welcome!

Meanwhile, I can make sure it doesn't break on mobile.

Mobile navigations clearly need some work, too.

simison: Aight, fixed in https://github.com/Trustroots/trustroots/pull/1902/commits/2c9c8038398f6a7e7d88ed45bfe4588dd6d47fb0

image

simison:
@hlibbabii Thanks for testing! I fixed the mobile version, I can merge once CI passes.

As I wrote in https://github.com/Trustroots/trustroots/pull/1902#discussion_r541789530, I'd prefer some other visuals than this kind of labels but they're fine intermediate and gets us moving in the right direction.

PR #1903, opened 2020-12-05
Experiences: always show report toggle

simison:
#### Proposed Changes

* Add `Switch` component (there's a similar Angular Directive and these share CSS styles for now)
* Make report toggle always visible when writing an experience, not just when choosing "no" to the recommendation
* Elaborate a bit more in length that reporting is important

#### Testing Instructions

* Write an experience about someone

* Choose "no", text should appear:

    Screenshot 2020-12-06 at 01 40 39

* No matter if you choose yes/no/I don't know, you can see the toggle:

    Screenshot 2020-12-06 at 01 40 27

* Pressing the toggle shows the text field:

    Screenshot 2020-12-06 at 01 40 31

### DIFF
--- a/modules/core/client/components/Switch.js
+++ b/modules/core/client/components/Switch.js
@@ -2,13 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-export default function Switch({ checked, children, onChange, small }) {
-  // Angular Directive supports also these CSS classes:
-  // tr-switch-right
+export default function Switch({ checked, children, onChange, isSmall }) {
+  // Angular Directive supports also `tr-switch-right` CSS class
   return (
     <label
       className={classnames('tr-switch', {
-        'tr-switch-sm': small,
+        'tr-switch-sm': isSmall,
       })}
     >
       <input type="checkbox" checked={checked} onChange={onChange} />
@@ -21,5 +20,5 @@ Switch.propTypes = {
   checked: PropTypes.bool.isRequired,
   children: PropTypes.node.isRequired,
   onChange: PropTypes.func.isRequired,
-  small: PropTypes.bool,
+  isSmall: PropTypes.bool,
 };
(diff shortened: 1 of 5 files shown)


chagai95:
Is this so people don't have to give others a bad review which will be visible for others in order to report them? Makes sense, maybe we can add this to the text? "We will not let the other person know you reported them"

simison:
> Is this so people don't have to give others a bad review which will be visible for others in order to report them? 


Yep, several females made comments that this might be helpful.


> maybe we can add this to the text? "We will not let the other person know you reported them"

Good idea. 👍

simison:
> maybe we can add this to the text? "We will not let the other person know you reported them"

@Trustroots/safety folks :wave: Thoughts on this? Any situations where this necessarily isn't true? The sentiment is that we'd encourage people to report members in lower reshold.

noahsmindfuck:
* We do approach reported members with the information that they were reported. We do not tell them who reported them but in some cases it might be obvious.
* It does make sense to put it there but we shouldn't put too much text. maybe we cann add a link to a FAQ question that says sth. like "How do I report someone and how do you deal with it?". There we can elaborate more clearly and transparent how we handle it.

simison:
Thanks @noahsmindfuck! The FAQ entry makes probably sense outside this context, too.

Yeah, I agree less text is better — would adding "Privately..." be short and effective?

image

chagai95:
> Thanks @noahsmindfuck! The FAQ entry makes probably sense outside this context, too.
> 
> Yeah, I agree less text is better — would adding "Privately..." be short and effective?
> 
> image

Anonymously?

noahsmindfuck:
I remember writing the text that would be the answer to this FAQ someday but I have no idea why and where. :) We documented once how to deal with stuff. Maybe check the safety repository.

simison:
> Anonymously?

That could indicate that support wouldn't know _who_ reported? Would "Privately" have the same issue?

noahsmindfuck:
> Anonymously?

It is not anonymously. We still know. Secretly or another synonym yould work.

simison:
Updated the copy to "Privately report this person to moderators", good for another look before merge!

We can loop back later to add clarifying copy if needed.

hlibbabii (approved):
Looks good!

PR #1905, opened 2020-12-06
Fix: profile description field height jumping

simison:
#### Proposed Changes

* Tiny fix to add CSS specificity on profile description element, to avoid underlying component CSS (min-heigh: 30px) overriding the profile description CSS (min-height: 150px).

#### Testing Instructions

* Have empty profile description
* Try to edit it

**Before**

broken

**After**

works


hlibbabii (approved):
Looks good!

PR #1907, opened 2020-12-06
Reactify "user does not exist" message

simison:
#### Proposed Changes

* New try at https://github.com/Trustroots/trustroots/pull/1878 — see instructions/screenshot there.

#### Testing Instructions

* Open http://localhost:3000/profile/doesnotexist — should show "not found" message
* Open http://localhost:3000/profile/profile-that-exists — should work normally

I'll merge once CI passes.

### DIFF
--- a/modules/users/client/components/UserDoesNotExist.component.js
+++ b/modules/users/client/components/UserDoesNotExist.component.js
@@ -1,0 +1,26 @@
+// External dependencies
+import { useTranslation } from 'react-i18next';
+import React from 'react';
+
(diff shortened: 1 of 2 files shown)


simison:
So this turned out to be simpler than I thought:

Breaks:
```html
  <user-does-not-exist />
```

Works:
```html
  <user-does-not-exist ></user-does-not-exist>
```

Even if we're writing react here, it's still converted into Angular directive and looks like that's what they require to work. :-)

PR #1909, opened 2020-12-06
Experiences: redesign success messages

simison:
#### Proposed Changes

* Redesign "done" and "you've already left reference" messages
* Shaping copy towards "experiences"

**TODO:** 
Tests still seem to need adjusting but this is good for functional/code/copy review.

### Screenshots
**"Done" before**

Screenshot 2020-12-06 at 23 28 14

**"Done" after**

Screenshot 2020-12-06 at 23 19 18

**"Already" before**

Screenshot 2020-12-06 at 23 27 48

**"Already" after**

Screenshot 2020-12-06 at 23 26 11

#### Testing Instructions

* Leave reference to someone, with and without reporting them
* Try the same again and see "already wrote" message


simison:
Some thoughts:

- Ideally "you already shared" would have some messaging to confirm how many days they still need to wait, or if it's public or not, but that's for another time.

- Copy could be different if they were reported, feels a little too positive in that case.

- We could just bring them straight to the experiences-list and show the success message there, but that's for another time and PR.

hlibbabii:
> * Copy could be different if they were reported, feels a little too positive in that case.

I also noticed that when I saw the screenshot, but I thought that it's even good. It doesn't make you feel bad about reporting people in a way?

modules/core/client/components/SuccessMessage.js
```diff
@@ -0,0 +1,42 @@
...
+const Spacer = styled.div`
+  margin: 15px 0;
+`;
+
+export default function SuccessMessage({ children, cta, title }) {
+  return (
+    <Container>
+      <Sparkles src={sparklesSvg} alt="" aria-hidden />
+      <h2 className="font-brand-light">{title}</h2>
+      {children && <Spacer>{children}</Spacer>}
```
hlibbabii, line 32: I see that `children` is not used now. Are we going to need it for future uses of `SuccessMessage`?

hlibbabii (approved):
Works great!

modules/core/client/components/SuccessMessage.js
simison: It's used here:

https://github.com/Trustroots/trustroots/blob/ef21faf9c43aeb34399e49b6db815e417fe0c136/modules/references/client/components/create-reference/SubmittedInfo.js#L25-L52

It's all the content between `<SuccessMessage></SuccessMessage>

hlibbabii: Oh, my bad! Need to work on my react. Thanks for the explanation!

PR #1910, opened 2020-12-07
create a separate endpoint to retrieve logged-in user's reference given to `userTo`

hlibbabii:
#### Proposed changes

Add route *GET: /api/my-reference?userTo=\<userTo\>* and make client code call this route when checking if the currently auth user has already left a reference to `userTo` 

#### Testing instructions

1. Try exercising the route above
1. Log in; leave a reference to someone; try leaving reference one more time - should not be possible

TBA: 
- [x] tests
- [x] validation
- [x] remove obsolete`userFrom` param from `/api/references` which is now obsolete -- **will be done in a separate PR**

part of ongoing work on #1493 

related to #1667 and #1860 (untangles a bit changes that need to be done there)

### DIFF
--- a/modules/references/client/api/references.api.js
+++ b/modules/references/client/api/references.api.js
@@ -85,13 +85,18 @@ export async function read({ userFrom, userTo }) {
  */
 export async function readMine({ userTo }) {
   const params = { userTo };
-  const { data: reference } = await axios.get('/api/my-reference', {
-    params,
-  });
-
-  return reference !== null
-    ? mapObjectToObject(reference, referenceMapping, true)
-    : null;
+  try {
+    const { data: reference } = await axios.get('/api/my-reference', {
+      params,
+    });
+    return mapObjectToObject(reference, referenceMapping, true);
+  } catch (err) {
+    if (err.response?.status === 404) {
+      return null;
+    } else {
+      throw err;
+    }
+  }
 }
 
 /**
--- a/modules/references/client/api/references.api.js
+++ b/modules/references/client/api/references.api.js
@@ -84,7 +84,7 @@ export async function read({ userFrom, userTo }) {
  * @returns Promise<Reference[]> - array of the found references
  */
 export async function readMine({ userTo }) {
-  const params = { userTo: userTo };
+  const params = { userTo };
   const { data: reference } = await axios.get('/api/my-reference', {
     params,
   });
--- a/modules/references/client/api/references.api.js
+++ b/modules/references/client/api/references.api.js
@@ -77,6 +77,24 @@ export async function read({ userFrom, userTo }) {
 }
 
 /**
+ * API request: read references written by loggedIn user, filter them by userTo,
+ * and sort by 'created' field starting from the most recent date
+ *
+ * @param {string} userTo - id of user who received the reference
+ * @returns Promise<Reference[]> - array of the found references
+ */
+export async function readMine({ userTo }) {
+  const params = { userTo: userTo };
+  const { data: reference } = await axios.get('/api/my-reference', {
+    params,
+  });
+
+  return reference !== null
+    ? mapObjectToObject(reference, referenceMapping, true)
+    : null;
+}
+
+/**
  * API request: report a member
  * @TODO this request belongs to a different module
  * @param {object} user - member to report
--- a/modules/references/client/components/CreateReference.component.js
+++ b/modules/references/client/components/CreateReference.component.js
@@ -35,16 +35,13 @@ export default function CreateReference({ userFrom, userTo }) {
 
   useEffect(() => {
     (async () => {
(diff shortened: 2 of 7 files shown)


modules/references/client/api/references.api.js
```diff
@@ -76,6 +76,24 @@ export async function read({ userFrom, userTo }) {
...
+ * @param {string} userTo - id of user who received the reference
+ * @returns Promise<Reference[]> - array of the found references
+ */
+export async function readMine({ userTo }) {
+  const params = { userTo: userTo };
+  const { data: reference } = await axios.get('/api/my-reference', {
+    params,
+  });
+
+  return reference !== null
```
hlibbabii, line 92: Should we return 404 here rather than 200 and `null`? Seems like returning 404 is a normal thing to do when something is not found. On the other hand, reference not being found here is a part of the normal flow, not an exceptional situation.

simison: I think 404 is a normal behavior in REST API principles (although I know our APIs are more of a mixture, especially when it comes to routes). We have a few APIs already doing the same.

Then you just need to take better care to `catch()` it at the client-side so that you don't bubble up errors.

Not a strong opinion tho, I'd leave this up to you :-)

simison: Another way to look at it that might help you decide:

If you're fetching for an object, lack of the object should be 404.

If you're fetching an answer to a question, then 200 with yes/no seems appropriate.

Does that make sense?

hlibbabii: Yeah, absolutely, everything makes sense. Here the question is not what the endpoint should return (I agree that 404 is more suitable) but rather whether this endpoint is the right thing to call when we just need to ask this yes/no question

hlibbabii: I am thinking again about it, now having one route back again looks pretty good to me :D

`/api/experiences-by/:experienceUserId?userTo=<userTo>&includeReplies=<includeReplies>`

**Checking if haven't shared experience yet:**
`/api/experiences-by/:experienceUserId?userTo=<userTo>&include_replies=false`
status 200 and `len(fetched_experiences) == 0`

** Displaying on the profile page: **
`/api/experiences-by/:experienceUserId?include_replies=true`

modules/references/client/api/references.api.js
```diff
@@ -76,6 +76,24 @@ export async function read({ userFrom, userTo }) {
...
 
+/**
+ * API request: read references written by loggedIn user, filter them by userTo,
+ * and sort by 'created' field starting from the most recent date
+ *
+ * @param {string} userTo - id of user who received the reference
+ * @returns Promise<Reference[]> - array of the found references
+ */
+export async function readMine({ userTo }) {
+  const params = { userTo: userTo };
```
simison, line 87: This can be shortened to:

```js
const params = { userTo };
```

https://alligator.io/js/object-property-shorthand-es6/

simison: Added a lint rule to guide us automatically: https://github.com/Trustroots/trustroots/pull/1927

modules/references/client/api/references.api.js
simison: I think this has two aspects:

- What makes more sense as an API route?
- How does the underlying code look like?

I agree it can make sense as a single route for all this, but at the same time does it make the controller for that route more complex and harder to reason about, when it has to deal with:
- Returning replies or not (instead of just defaulting always returning them)
- If it should be limited only to `sharedWithUserId` (instead of just ignoring this aspect)

Furthermore, we already need a very similar query doing the same thing (`checkDuplicate()`) so it would make sense to re-use that here for this task, instead of using the already complex aggregate query.

So I think a separate API endpoint might still make sense for simplicity's sake, even if API paths point-of-view is different.

Not super strong opinion tho but just my trail of thought when I think about this. What do you think?

hlibbabii: @simison yeah, in general, I agree with your reasoning. Check out the new changes

simison (approved):
Code looks good and this tests well. 👍

modules/references/client/api/references.api.js
```diff
@@ -76,6 +76,29 @@ export async function read({ userFrom, userTo }) {
...
 
+/**
+ * API request: read references written by loggedIn user, filter them by userTo,
+ * and sort by 'created' field starting from the most recent date
+ *
+ * @param {string} userTo - id of user who received the reference
+ * @returns Promise<Reference[]> - array of the found references
+ */
+export async function readMine({ userTo }) {
+  const params = { userTo };
```
simison, line 87: Total nit-pick and feel free to ignore; might make sense to pass `userTo` directly to `get()` instead of assigning it to intermediate `params` which is used anyway just once?

PR #1927, opened 2020-12-12
Lint: add object-shorthand rule

simison:
#### Proposed Changes

* Add [`object-shorthand`](https://eslint.org/docs/rules/object-shorthand) Eslint rule. Works in both backend and frontend code.
* Run `eslint --fix` to automatically handle all the existing rule violations in codebase 

#### Testing Instructions

Should complain:
```js
const foo = 1;
const bar = { foo: foo };
```

image

Should work:
```js
const foo = 1;
const bar = { foo };
```


hlibbabii (approved):
👍🏼

PR #1930, opened 2020-12-12
Build: stop auto-detecting outdated dependencies

simison:
#### Proposed Changes

Modify the script that detects outdated `node_modules` and runs install. It now just warns when dependencies are missing or outdated.

image

image

I added this to avoid confusion new and inexperienced volunteers were having with installing and keeping dependencies up to date in their setup.

Still a fair concern, but I'd rather deal with educating folks into NPM procedures than deal with getting dependencies installed each time I want to review someone's PR. :-)

We just need to be mindful and run `npm ci` to refresh dependencies ourselves when needed.

#### Testing Instructions

* Run `touch package.json`
* Run `npm start`; no install should occur


hlibbabii:
I agree. It's annoying to reinstall dependencies all the time.

Have you thought about 1) checking if the dependencies are up-to-date and just giving a warning if they are not 2) have a separate npm command to run things without updating dependencies

simison:
@hlibbabii good questions!

> checking if the dependencies are up-to-date and just giving a warning if they are not 

That's a nice idea. I'll implement it like that for now to aid with new less (npm-) experienced volunteers. I can also add a blurb if dependencies are completely missing.

If it feels still over the top, we can later loop back to remove it completely.

> have a separate npm command to run things without updating dependencies

Thanks for the idea. I'd rather see us simplify processes and not have separate commands for multiple purposes.

simison:
@hlibbabii I updated the script and PR description.

hlibbabii (approved):
👍🏼

PR #1936, opened 2020-12-13
Compare references on the client based on the id rather than username

hlibbabii:
#### Proposed Changes

* use `_id` field to compare experiences instead of `username`.

as discussed here: https://github.com/Trustroots/trustroots/pull/1667#discussion_r537352161

#### Testing Instructions

* Go to a profile with many experiences among which there are ones with replies. Check that they are displayed properly and that the replies are present.
* Share an experience with someone. It should not be possible to share another experience with the same person. 

Ongoing work on #1493


simison (approved):
Looks good and tested that it works. 👍

PR #1937, opened 2020-12-13
Add experience list indentation

simison:
I'm playing with the "reply" visuals a bit.

- Does it make sense to include "reply" in cases where there isn't experience shared yet? Does it bring any value?
- If the above does make sense, we would probably need to pull at least the name+picture of that person, or otherwise, it's out of context since it's unclear who "Zita" is writing about in the below screenshots.
- It might make sense to show "pending" references in this way nevertheless? The more you see in that case the clearer it is.

Other:
- Thoughts on list indentation? I was thinking to add some other indicator that the two experiences are related: either by visually having them both in the same container/panel, having some kind of arrow, playing with left/right alignment kinda like in some chat apps, or something else...
- Thoughts on keeping dates visible both ways? I'm not sure what value there is to hide it, since the two experiences could be written at fairly different times.

#### Proposed Changes

* Make time always visible on each experience

Screenshot 2020-12-13 at 19 21 10

* Add indentation in experience pairs
  
image

* Add empty placeholder experience when one doesn't exist but "reply" exists

Screenshot 2020-12-13 at 19 21 37

No "reply" looks like such:

Screenshot 2020-12-13 at 19 22 39
 

#### Testing Instructions

* WIP

### DIFF
--- a/modules/references/client/components/read-references/Reference.js
+++ b/modules/references/client/components/read-references/Reference.js
@@ -55,6 +55,16 @@ const FeedbackPublic = styled.div`
   max-width: 600px;
 `;
 
+const ReferenceContainer = styled.div`
+  margin-left: 0;
+
+  ${props =>
+    props.inCreatorProfile &&
+    `
+      margin-left: 20px;
+    `}
+`;
+
 export default function Reference({ reference, inRecipientProfile }) {
   const { t } = useTranslation('references');
 
@@ -84,10 +94,13 @@ export default function Reference({ reference, inRecipientProfile }) {
   const inCreatorProfile = !inRecipientProfile;
 
   return (
-    <div className="panel panel-default" id={_id}>
+    <ReferenceContainer
+      className="panel panel-default"
+      id={_id}
+      inCreatorProfile={inCreatorProfile}
+    >
       <div className="panel-body reference">
         <ReferenceHeading>
-          {inCreatorProfile && <div>{t('their reply')}</div>}
           <Avatar user={userFrom} size={36} />
           <UserMeta>
             <strong>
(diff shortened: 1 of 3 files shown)


simison:
Just for a reference (🥁), a couple examples:

**Couchsurfing**

image

**AirBnb**

image

image

simison:
Aight, new version which I feel is much better than my previous one:

Mobile
Screenshot 2020-12-22 at 19 19 56

Desktop
Screenshot 2020-12-22 at 19 19 41

- If there is no original experience (only "response"), nothing shows up
- Responses are shown inside the same container and slightly indented
- Response's avatar is smaller, date shows up little differently and isn't a link,

modules/users/client/components/UserLink.js
```diff
@@ -7,14 +8,21 @@ import React from 'react';
  * @param {string} user.displayName
  * @param {string} user.username
  */
-export default function UserLink({ user }) {
+export default function UserLink({ user, className }) {
+  const { t } = useTranslation('users');
+
+  if (!user?.username) {
+    return <span className={className}>{t('Anonymous member')}</span>;
```
simison, line 15: "New feature". Just to protect against cases where we don't have username when user has been deleted.

modules/references/client/components/read-references/Reference.js
```diff
@@ -81,37 +97,25 @@ export default function Reference({ reference, inRecipientProfile }) {
...
 
-  const inCreatorProfile = !inRecipientProfile;
-
   return (
     <div className="panel panel-default" id={_id}>
-      <div className="panel-body reference">
-        <ReferenceHeading>
-          {inCreatorProfile && <div>{t('their reply')}</div>}
+      <div className="panel-body">
+        <Header>
```
simison, line 103: Contemplated encapsulating contents of the header into its own component and re-using it across both experiences (original & response).

They ended up looking quite different at the end so opted in to not doing so.

PR #1951, opened 2020-12-20
Remove logged-out search page

simison:
#### Proposed Changes

Removes logged-out search page. I started looking at turning this into React but realized it isn't really much in use:

**Before**

Screenshot 2020-12-20 at 19 46 45

**After**

image

It was created because we linked into search directly from Hitchwiki, and since Trustroots was new to most folks in Hitchwiki it was appropriate to show some explanation of what TR is. By now most Hitchwiki folks should know it pretty well. :-) 

It still is kinda useful for that, but at the same time, I like to simplify and have less code around. :-)

Screenshot 2020-12-20 at 19 50 54

#### Testing Instructions

* As logged out, open `/search`
* You should see a regular login form instead of special invite to login


hlibbabii (approved):
👍🏼

PR #1953, opened 2020-12-22
Update copy from references to experiences

simison:
#### Proposed Changes

* Update UI copy from "references" to "experiences"
* Left out rest of the email copy for now, I'll open a separate PR for that

#### Testing Instructions

* Smoke test UI for references

### DIFF
--- a/modules/users/client/views/profile/profile-view.client.view.html
+++ b/modules/users/client/views/profile/profile-view.client.view.html
@@ -244,7 +244,7 @@ this is a fallback when the contacts are loading
                   ng-if="app.appSettings.referencesEnabled"
                 >
                   <a ui-sref="profile.references.list" role="tab">
-                    Experences
+                    Experiences
                   </a>
                 </li>
               </ul>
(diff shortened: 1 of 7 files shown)


simison:
Rebased

modules/users/client/views/profile/profile-view.client.view.html
```diff
@@ -172,7 +172,7 @@
                   ui-sref="profile.experiences.new({username: profileCtrl.profile.username})"
                 >
                   <i class="icon-plus-squared-alt"></i>
-                  Write a reference
+                  Share your experience
```
hlibbabii, line 175: this, again, does not change the title on mobile, but I guess we can merge this and update it later?

hlibbabii (approved):

modules/users/client/views/profile/profile-view.client.view.html
simison: Yep, correct, let's handle the mobile navigation in one swoop separately.

PR #1957, opened 2020-12-22
Turn 404 not found page into React component

simison:
#### Proposed Changes

* Just convert a simple Angular template to React

#### Testing Instructions

* Open any page URL that doesn't exist


hlibbabii (approved):
👍🏼

PR #1958, opened 2020-12-22
add gender when seeding users for test db

hlibbabii:
#### Proposed Changes

* when seeding users for test Db, add also gender

#### Testing Instructions

* run `npm run dropdb` (warning, will remove all test data!), then `npm seed` to generate test data again. Most of the users should have gender specified in their profiles


simison (approved):
Good idea 👍 

Tested that this works well. I like how from enums we get even random empty value.

PR #1959, opened 2020-12-23
Statistics: rewrite in React

simison:
#### Proposed Changes

Instead of copying the old template html, I just threw it into a simple CSS grid instead.

* Rewrite `/statistics` in React; one less Angular module 🎉 
* Add simple util to retrieve network label and use it for profile, too.
* Modernize test file to use new utils
* Modify API response to have data we need at client-side, instead of calculating it all at clientside.

#### Testing Instructions

* Open `/statistics` as logged in and logged out user

**Before**

stats-old

**After**

stats-new


testutils/common/data.common.testutil.js
```diff
@@ -51,14 +51,30 @@ function generateClientUser(overrides = {}) {
...
-  { public: pub, locale } = {},
+  {
+    additionalProvidersData,
+    extSitesBW,
+    extSitesCS,
+    extSitesWS,
+    locale,
+    newsletter,
+    public: pub,
+  } = {},
```
simison, line 62: Was a bit surprised this doesn't just take anything, but passes a specific list of keys allowed for modification. 🤷‍♂️

simison:
I'll get this in & for translating — it's not critical feature at all so fine if it's not perfect. 🙃

PR #1960, opened 2020-12-23
Fix experiences tab link on the profile page

hlibbabii:
#### Proposed Changes

* Fix experiences tab link on the profile page

#### Testing Instructions

- Go to any profile page
- Click on the "Experiences" tab
- Experiences should be displayed


simison (approved):
👍 Good catch

PR #1961, opened 2020-12-23
Tooling: adjust newsletter list export script

simison:
The script is used to export emails for Mailtrain for sending our newsletter.

Found an old script that I just place in different folder so that it gets linted.

I also modernized the file a bit.


simison:
Since this isn't production code, I'll just go ahead and merge this.

PR #1963, opened 2020-12-23
Docs: add newsletter documentation

simison:
#### Proposed Changes

* Adds basic newsletter documentation. Mostly just wanted to get "Updating subscribers on Mailtrain" documented at this point.

### DIFF
--- a/docs/Newsletter.md
+++ b/docs/Newsletter.md
@@ -1,0 +1,37 @@
+# Newsletter
+
+Occasionally we send a newsletter to our members.
+
+## Sending newsletter
+
+### Creating campaign
+
+Create campaign on [Mailtrain](https://mailtrain.trustroots.org). It's the app we use for sending newsletters.
+
+### Updating subscribers on Mailtrain
+
+Mailtrain's list and users on the site are not synchronised. Therefore you need to manually update the list each time you want to send out a campaign.
+
+Generate the list on the production server:
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/Newsletter.md
```diff
@@ -0,0 +1,37 @@
...
+
+Go to importing tool in Mailtrain (_Lists → Trustroots Newsletter → Imports_) and create a new import with the file. It doesn't matter if the csv file has some existing emails — Mailtrain's import will ignore those.
+
+You can see from import "status" page in case the import failed due failed formatting on the file. If so, manually edit your file and consider adjusting the export script to deal with the case.
+
+Remember to delete both files after you don't need them anymore. Especially the local one.
+
+### Rollout
+
+Test the campaign first by sending it to a small, internal test list. Once that works out, do the full-rollout on the main list. Stagger it across long enough time, allowing around 1K subscribers per hour seems good pace.
```
MrSalami, line 35: We can only send out the campaign to 1k per hour?

simison: I'm not quite sure what's right pace. I can elaborate a bit more reasoning for it, and make the exact amount fuzzy so that folks don't take it literally.

simison: Looks like was had the setting set to 200 per hour at our settings. I lifted it up to 400 per hour now and clarified reasons in the doc.

PR #1966, opened 2020-12-23
Profile: convert data download to React

simison:
#### Proposed Changes

* Another quick Angular->React conversion of tiny bit of account page. Gets these strings to translation.

I left out the analytics bits because they aren't important.

#### Testing Instructions

* Open http://localhost:3000/profile/edit/account
* See the "download" bit, each link should work.
Screenshot 2020-12-24 at 00 32 55

### DIFF
--- a/modules/users/client/views/profile/profile-edit-account.client.view.html
+++ b/modules/users/client/views/profile/profile-edit-account.client.view.html
@@ -1,8 +1,6 @@
 <!-- Email -->
 <div class="panel panel-default">
-  <div class="panel-heading">
-    Email
-  </div>
+  <div class="panel-heading">Email</div>
   <div class="panel-body">
     <!-- Email Address -->
     <form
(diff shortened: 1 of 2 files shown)


modules/users/client/views/profile/profile-edit-account.client.view.html
```diff
@@ -1,8 +1,6 @@
 <!-- Email -->
 <div class="panel panel-default">
-  <div class="panel-heading">
-    Email
-  </div>
+  <div class="panel-heading">Email</div>
```
simison, line 3: These were just some changes done by Prettier/Eslint.

hlibbabii (approved):
👍🏼

hlibbabii:
Do we need to add a link for experiences, btw?

simison:
Yeah, it would be good to add experiences here once it's live or earlier behind the feature flag. Part of the reason I already looked at this now.

PR #1968, opened 2020-12-25
Refactor: use flattened 'interaction' object only in react component

hlibbabii:
#### Proposed Changes

* Do not flatten-unflatten interactions objects (which is inside the experience object) back and forth. Unflatten it once when creating an experience in the react component and always use the unflattened structure after that.

Slack discussion:
https://trustroots.slack.com/archives/C0A3Q15SS/p1608857982047000?thread_ts=1608764438.035300&cid=C0A3Q15SS

#### Testing Instructions

* Check that the experiences tab in someone's profile loads without errors and that interaction labels 'met', 'hostedMe', 'hostedThem' are present. Check everything is fine with the labels also when creating an experience.

### DIFF
--- a/modules/references/client/components/read-references/Meta.js
+++ b/modules/references/client/components/read-references/Meta.js
@@ -13,7 +13,7 @@ const Labels = styled.div`
   }
 `;
 
-export default function Meta({ met, hostedMe, hostedThem, recommend }) {
+export default function Meta({ interactions, recommend }) {
   const { t } = useTranslation('references');
 
   return (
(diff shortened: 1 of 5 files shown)


modules/references/client/components/read-references/Meta.js
```diff
@@ -36,8 +36,6 @@ export default function Meta({ met, hostedMe, hostedThem, recommend }) {
 }
 
 Meta.propTypes = {
-  hostedMe: PropTypes.bool.isRequired,
-  hostedThem: PropTypes.bool.isRequired,
-  met: PropTypes.bool.isRequired,
+  interactions: PropTypes.object.isRequired,
```
simison, line 39: Could do something like this to describe the shape here:

https://github.com/Trustroots/trustroots/blob/07249bba5adfe650c9c62469617f8440cb137eeb/modules/users/client/users.prop-types.js#L1-L7

Totally also ok if not.

hlibbabii: oh, absolutely, also for experience object then, and experience pairs. I guess we can do it in a separate PR

simison (approved):
Looks good and tests well 👍

modules/references/client/components/read-references/Meta.js
simison: Yes, totes optional and fine in follow-ups. 👍

PR #1969, opened 2020-12-25
Add LanguageList component and fill test users with random languages

simison:
#### Proposed Changes

* Adds simple `LanguageList` component to be re-used across two places.
* The profile next to message thread was showing just language codes previously, so this fixes that.
* Later on I can use this at search results as well once it's in React.

#### Testing Instructions

Component
* Open message thread, observe languages
* Open a profile, observe languages

Test data
- Run `npm run seed:users 5`
- Check latest users with `db.getCollection('users').find({},{languages:1}).sort({created:-1})`
- They should have random languages:


hlibbabii (approved):
👍🏼

Issue #1840, opened 2020-11-19
Add experience count to profile


hlibbabii:
Part of the ongoing work on the experiences feature #1493 

**This issue summarizes the cons and pros of having the count of experiences added to the profile and ideas of how this count could be shown.** In addition to the total count of experiences, we can also show counts for some specific categories of experiences (e.g. negative, neutral, positive, if we finalize our decision to differentiate them like that, TODO link to the non-existent (?) yet issue).

**Might be related to the following replies** (What people don’t like with references) **from the** [user survey](https://codi.kanthaus.online/s/member_survery_references):

_OPINION-1: creates a market/competition of exchanging/having references (8 people)
OPINION-2: contacting hosts with many references bc. chance of being hosted is increased. Leads to the wrong assumption about people and keeps you from meeting unexperienced hosts (1 person)
OPINION-3:  lots of positive references made people feel uncomfortable to leave negative references (1 person)_

**The total number of experiences:**
- Specify the exact total number 
    + allows to immediately understand how much experience the user has
    + this is the UI people would expect(one expects some number when there is a list of something)
- Experience levels: 1 to 10 exp - level 1, 10-50 - level 2, etc (Loberto):
    + Might be good enough to get an idea about the experience of the person (while 2 vs 50 experiences matters, 30 vs 50 does not)
    + partially solves the issues described in _OPINION-1, 2_
- Verbal experience levels: e.g. "Have had a few/some/lots of experiences" or with a single word (Mikael)
    + Might be good enough to get an idea about the experience of the person (while 2 vs 50 experiences matters, 30 vs 50 does not)
    + partially solves the issues described in _OPINION-1, 2_
- Do not specify the number at all
    + Only if can be estimated from the UI (list size, number of pages, tree size)
    + partially solves the issues described in _OPINION-1, 2_
- Make exact number private (Loberto), publicly - nothing it all or levels

**The number of experiences in different categories (negative/positive/neutral if we have them in the end):**
- specify the exact numbers:
    + for the users who are especially concerned about safety will find it convenient to immediately see the number of negative references 
- don't specify the exact numbers:
    + makes sense if we decide not to specify the exact total number
    + partially solves the issue from _OPINION-3_

Link to the slack-discussion: https://trustroots.slack.com/archives/C017JUMR6SZ/p1597749747230700

**Option to be implemented:** TBD

chagai95:
Need to ask more volunteers, they get to decide, if no decision is made it gets left as is, which means the experiences are shown.

PR #1970, opened 2020-12-25
Experiences: add count to tab

simison:
Resolves https://github.com/Trustroots/trustroots/issues/1840 (or is that a bit different issue?)

#### Proposed Changes

* Convert "profile tabs" into React component
* Show number of experiences in the tab
* Show a notification dot if there are pending experiences and I'm looking at my own profile
* Adds new API endpoint just to receive the count. Alternatively we could fetch the whole list at this point, but it would be wasteful if user never opens the experiences tab.

#### Testing Instructions

* When user has public experiences, tab is shown with a count
* When user doesn't have public experiences, tab is not shown
* Tab shows up always also if you look at your own profile
* If you're looking at your own profile...
   * ...also private experiences are counted in
   * ...you see a dot if there are pending experiences
* If experiences are disabled, tab is hidden

When looking at other's profile (even when they have pending experiences):

Screenshot 2020-12-26 at 13 08 46

When looking at my own profile when there are pending experiences:

Screenshot 2020-12-26 at 13 03 03

### DIFF
--- a/modules/core/client/components/Badge.js
+++ b/modules/core/client/components/Badge.js
@@ -6,11 +6,11 @@ import styled from 'styled-components';
 const Dot = styled.div`
   background: #d9534f;
   border-radius: 50%;
-  height: 5px;
+  height: 7px;
   position: absolute;
-  right: 10px;
+  right: 9px;
   top: 6px;
-  width: 5px;
+  width: 7px;
 `;
 
 export default function Badge({ children, withNotification }) {
--- a/modules/core/client/components/Badge.js
+++ b/modules/core/client/components/Badge.js
@@ -1,0 +1,28 @@
+// External dependencies
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from 'styled-components';
+
+const Dot = styled.div`
+  background: #d9534f;
+  border-radius: 50%;
+  height: 5px;
+  position: absolute;
+  right: 10px;
+  top: 6px;
+  width: 5px;
+`;
+
+export default function Badge({ children, withNotification }) {
+  return (
+    <span className="badge">
+      {children}
+      {withNotification && <Dot />}
+    </span>
+  );
+}
+
+Badge.propTypes = {
+  children: PropTypes.node.isRequired,
+  withNotification: PropTypes.bool,
+};
--- a/modules/references/server/controllers/reference.server.controller.js
+++ b/modules/references/server/controllers/reference.server.controller.js
@@ -548,7 +548,6 @@ exports.readMine = async function readMine(req, res) {
 };
 
 exports.getCount = async function getCount(req, res, next) {
-  console.log('getCount — userTo:', req.query.userTo); //eslint-disable-line
   const { userTo } = req.query;
 
   if (!mongoose.Types.ObjectId.isValid(userTo)) {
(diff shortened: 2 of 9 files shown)


simison:
Just to note that I didn't look at mobile experience yet this time.

simison:
One aspect I was a bit un-decisive about: if to show the tab with 0 for everyone, or keep the tab hidden when there are no references.

I remember earlier leaning towards just showing it to everyone for simplicity and consistency. Thinking it more, I felt like showing a tab where there really isn't anything to see isn't great experience either. Since we already didn't show contacts tab in that case either, I made the decision to keep it consistent across contacts/experiences.

modules/users/client/components/ProfileTabs.component.js
```diff
@@ -0,0 +1,103 @@
...
+  isOWnProfile,
+  userId,
+  username,
+}) {
+  const { t } = useTranslation('users');
+  const [experiencesCount, setExperiencesCount] = useState(0);
+  const [pathName, setPathName] = useState(initialPathName);
+
+  useEffect(async () => {
+    const count = await getExperiencesCount(userId);
```
hlibbabii, line 24: if experiences are disabled, the call to back-end will still be made, right?

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -546,3 +546,32 @@ exports.readMine = async function readMine(req, res) {
...
+  }
+
+  try {
+    const query = {
+      userTo: new mongoose.Types.ObjectId(userTo),
+    };
+
+    // Allow non-public references only when userTo is self
+    if (!req.user._id.equals(userTo)) {
+      query.public = true;
```
hlibbabii, line 567: Just a random idea: always show the number of published references. If you look at your profile and there are non-published ones, put an asterisk: (4*)

hlibbabii (approved):
Tested - works well!

modules/references/server/controllers/reference.server.controller.js
simison: Thanks for the idea!

I gave it a think and concluded:
- It's best to show the same number in the tab than how many experiences they also see in the list. If there's a mismatch, it can get folks confused.
- Asterix would be potentially unclear and we'd need to explain it
- Little more common pattern is a red dot for "unread" or for "notifications" and I think that's a good improvement to call attention to the fact that there are pending experiences.

I added the dot in latest iteration, and update PR description with fresh screenshots. Lemme know what you think!

modules/users/client/components/ProfileTabs.component.js
simison: Good catch! Fixed in https://github.com/Trustroots/trustroots/pull/1970/commits/8fbda57ab677e83472bbe87be2796a75b183b652

simison:
@hlibbabii thanks for the review & testing! I pushed new iteration based on idea I got from your comment. Take a look and lemme know what you think. :-)

modules/references/server/controllers/reference.server.controller.js
```diff
@@ -546,3 +546,42 @@ exports.readMine = async function readMine(req, res) {
...
+      ? await Reference.find({
+          ...query,
+          public: false,
+        }).count()
+      : 0;
+
+    return res.status(200).json({
+      count: privateCount + publicCount,
+      // `hasPending` included only for own profile
+      ...(isSelf ? { hasPending: Boolean(privateCount) } : {}),
```
hlibbabii, line 582: Is it worth returning private and public counts separately for the self profile (since we already got these values from the DB)? Then, `hasPending` can be derived from `privateCount` on the client. Additionally, the private count may be useful in the future.

simison: My thinking was that the exact count for pending ones isn't useful in client and I'd prefer just returning data in format we actually need it at the client side.

> Additionally, the private count may be useful in the future.

We can add it back at that point if it becomes useful.

What do you think?

hlibbabii: sure

modules/core/client/components/Badge.js
```diff
@@ -0,0 +1,28 @@
...
+
+const Dot = styled.div`
+  background: #d9534f;
+  border-radius: 50%;
+  height: 5px;
+  position: absolute;
+  right: 10px;
+  top: 6px;
+  width: 5px;
+`;
```
hlibbabii, line 14: Maybe make it slightly bigger?

image

Anyway, looks good to me!

hlibbabii (approved):

modules/core/client/components/Badge.js
simison: Valid point! Made it a bit bigger.

PR #1977, opened 2020-12-27
Refactor: return experiences paired up with responses from backend

hlibbabii:
#### Proposed Changes

* * All experiences end-points now return a response under a separate key

#### Testing Instructions

* Check that the list of experiences in someone's profile is the same and the replies are present; check that the counter shows the same number;
* Log in as someone who has received an experience that is still private. It should be visible in the "pending experiences" section
* Sharing experience with someone should work without errors. It should not be possible to share an experience again with that person.

Fixes #


modules/references/experiences.prop-types.js
```diff
@@ -0,0 +1,25 @@
...
+  met: PropTypes.bool.isRequired,
+  hostedMe: PropTypes.bool.isRequired,
+  hostedThem: PropTypes.bool.isRequired,
+});
+
+export const experienceType = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  public: PropTypes.bool.isRequired,
+  userFrom: PropTypes.object.isRequired,
+  userTo: PropTypes.object.isRequired,
```
hlibbabii, line 13: I decided to rename stuff later to make this PR cleaner

simison:
Seeing some React warnings in console:

image

simison:
In case where User A has written a experience to User B, but User B didn't respond and the experience went public (you can just do this directly via DB), we should hide the one experience in User A's profile:

**In this PR:**
Screenshot 2020-12-27 at 15 54 36

**In `master`:**
Screenshot 2020-12-27 at 15 55 12

hlibbabii:
@simison Thanks for the comments! I've fixed the things you mentioned

modules/references/experiences.prop-types.js
```diff
@@ -0,0 +1,25 @@
...
+  recommend: PropTypes.bool.isRequired,
+  feedbackPublic: PropTypes.string.isRequired,
+  response: PropTypes.shape({
+    _id: PropTypes.string.isRequired,
+    created: PropTypes.strimg.isRequired,
+    interactions: interactionsType.isRequired,
+    recommend: PropTypes.bool.isRequired,
+    feedbackPublic: PropTypes.string.isRequired,
+  }),
+});
```
simison, line 25: Let's put this file under client folder (`modules/references/client`) like it's at `user` module as well?

modules/references/client/components/read-references/ReferenceCounts.js
```diff
@@ -2,6 +2,7 @@ import { useTranslation } from 'react-i18next';
 import PropTypes from 'prop-types';
 import React from 'react';
 import styled from 'styled-components';
+import { experienceType } from '../../../experiences.prop-types';
```
simison, line 5: This import was failing, and I changed it  to `@/modules/references/experiences.prop-types` locally to get it work.

I don't have strong opinion but slight preference to `@/...`  imports over multiple `../`s, just for clarity.

simison, line 5: Hm, take the above back — there is some other issue with it. Just not sure why changing import made the app run for a moment:

image

hlibbabii: @simison fixed!

simison (approved):
Looks great and works well 👍

PR #1978, opened 2020-12-27
Convert "reference thread" feature into React

simison:
#### Proposed Changes

* Convert "reference thread" feature into React
* Put it back to message thread. It went missing when messages were converted to React

#### Testing Instructions

* Open message thread when there are no messages; you should not see anything
    Screenshot 2020-12-27 at 13 11 13

* Write a message to someone, and refresh the thread. You should still see nothing.

* Only when someone else writes you, you should see the question
    Screenshot 2020-12-27 at 13 11 47

* Response stays visible. Refresh the thread to test it.
    Screenshot 2020-12-27 at 13 47 17
    Screenshot 2020-12-27 at 13 47 13

* You can change your response by pressing "change"


hlibbabii:
Let's say, the user thinks it's too early to judge if the message they received is in the spirit of Trustroots but accidentally clicks "no". If the user then clicks "change", wouldn't the expectation be that the "no" reply is cancelled even without clicking "yes" after that?

simison:
@hlibbabii yeah, I think a feature like that would be good improvement. I don't think it matters a ton since a single false signal is fairly edge-case. How this PR is though, I didn't change the existing APIs or change the feature otherwise, just added it back.

hlibbabii:
> @hlibbabii yeah, I think a feature like that would be good improvement. I don't think it matters a ton since a single false signal is fairly edge-case. How this PR is though, I didn't change the existing APIs or change the feature otherwise, just added it back.

@simison Alright! Otherwise, I tested - everything works great!

simison:
Cheers — I'll merge this one and open an issue for potential convo about changing the question copy.

PR #1979, opened 2020-12-27
Reference thread stats: add gender

simison:
#### Proposed Changes

* Adds gender value to reference thread stats

#### Testing Instructions

* Enable influx from config, it doesn't matter if it's actually functioning or not
* Write yes / no reference to someone with and without gender
* Observe from console logs stats being sent


hlibbabii (approved):
Tests and look good!

PR #1980, opened 2020-12-27
Fix admin route check

simison:
#### Proposed Changes

* Fix checking for `admin` role requirement for `/admin/*` routes.

This was implemented earlier but broke in some refactor. Having it visible for everyone isn't a big deal since it's just client, admin APIs are still limited for admins only. Might as well just hide it though.

#### Testing Instructions

* Have user with `admin` in `roles` array.
* Have user without `admin` in `roles` array.
* Test opening `/admin`
* Test also as non-authenticated user

### DIFF
--- a/modules/core/client/controllers/app.client.controller.js
+++ b/modules/core/client/controllers/app.client.controller.js
@@ -239,10 +239,7 @@ function AppController(
           toState.requiresAuth = true;
         }
         // Check if user has the required role
-        else if (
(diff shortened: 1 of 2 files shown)


simison:
I tried adding client side route tests for this but Angular and its route tests are a nightmare to understand, and I didn't get it working. It's not too critical so we can just fix it again if it breaks.

PR #1981, opened 2020-12-27
Admin: list reference threads

simison:
#### Proposed Changes

* List latest 500 negative reference threads at admin tool

Screenshot 2020-12-27 at 17 36 51

#### Testing Instructions

* Give some positive and negative references to message threads (use https://github.com/Trustroots/trustroots/pull/1978)
* Open `/admin/reference-threads`
* See list of negative ones there

### DIFF
--- a/modules/admin/client/api/admin-reference-threads.api.js
+++ b/modules/admin/client/api/admin-reference-threads.api.js
@@ -1,0 +1,6 @@
+import axios from 'axios';
+
+export async function getReferenceThreads() {
+  const { data } = await axios.get('/api/admin/reference-threads');
+  return data;
+}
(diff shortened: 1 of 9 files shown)


noahsmindfuck:
Am I correct that these are experiences?
- why are they called references?
- the list has only the recent negative ones 
  - that were also reported to admins?
  - all of them
- why are they threads? bc one can respond with a counterrefernce?

simison:
@noahsmindfuck yep yep! These are for message thread yes/no ones indeed. 👍 

It's just to give little bit of visibility into what's happening with these. Some kind of alerting would be more adequate but this is still something.

Additionally there are these  in each profile:

image

simison:
Merged this so that volunteers can test more in production.

PR #1992, opened 2021-01-01
Docs: mention Prettier on install docs

simison:
#### Proposed Changes

* Add Prettier to install docs

I'll just merge this as it's a simple docs update.


simison:
Weird CI failure. Should be unrelated so could be in `master`. 🤔

PR #1993, opened 2021-01-01
Tests: clear all DB collections between tests runs & use abstraction everywhere

simison:
Test maintenance, get rid of some boilerplate.

#### Proposed Changes

* Clear all collections between tests;
   * See changes in `testutils/server/data.server.testutil.js`
   * Previously we were clearing out only `References` & `Users` collections
 * Use clear database abstraction everywhere
 * Use `utils.signIn` and `utils.signOut` utils in few places for readability
 * Disable/remove two flaky tests

#### Testing Instructions

* CI green

### DIFF
--- a/testutils/server/data.server.testutil.js
+++ b/testutils/server/data.server.testutil.js
@@ -65,32 +65,17 @@ async function saveExperiences(_docs, done = () => {}) {
 }
 
 /**
- * Clear specified database collections
- * @param {string[]} collections - array of collection names to have all documents removed
- * @returns {Promise<void>}
- */
-async function clearDatabaseCollections(collections) {
-  const models = collections.map(collection => mongoose.model(collection));
-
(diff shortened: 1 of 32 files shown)


testutils/server/data.server.testutil.js
```diff
@@ -65,34 +65,19 @@ async function saveReferences(_docs, done = () => {}) {
...
 
 /**
- * Clear specified database collections
- * @param {string[]} collections - array of collection names to have all documents removed
+ * Clear all collections in a database
+ * Usage in mocha: afterEach(clearDatabase)
  * @returns {Promise<void>}
  */
-async function clearDatabaseCollections(collections) {
+async function clearDatabase() {
```
simison, line 72: These changes could actually be done in separate PR to not wait for fixing failing tests here.

PR #1997, opened 2021-01-01
references -> experiences: react components

hlibbabii:
#### Proposed Changes

* Rename references to experiences inside react components and components themselves

#### Testing Instructions

* wait until ci passes
* smoke test all experiences functionality: viewing experiences, sharing an experience, share an experience with the same person for the second time


simison (approved):
👍  Smoke tested

PR #2008, opened 2021-01-03
Admin reference threads: fix listing when there are deleted users

simison:
#### Proposed Changes

* Fix in admin tool for reference threads in cases when user is deleted

Screenshot 2021-01-03 at 19 08 36

#### Testing Instructions

* Have a reference thread
* Delete user or change IDs
* Load `/admin/reference-threads`

**Before**
image

**After**

Screenshot 2021-01-03 at 19 08 36


simison:
I'll merge since this is admin tool so I'm less concerned about the code reviews for this.

PR #2011, opened 2021-01-03
Config: add few non allowed default usernames

simison:
#### Proposed Changes

* Adds `circles`, `moderator` and `official` to non-allowed usernames list.
* Also alphabetise the list

I've checked that these profiles don't exist yet in production.


hlibbabii:
> * Adds `circles`, `moderator` and `official` to non-allowed usernames list.

You meant `circle`, right?

simison:
> You meant circle, right?


Yes, sharp eyes! :-)

PR #2012, opened 2021-01-03
Build: add bundle analyzer

simison:
I've found a need for this a couple times already, so figured we might as well have it pre-configured for ease of use.

#### Proposed Changes

* Adds Webpack bundle analyzer (disabled by default)

#### Testing Instructions

* Add config to your `config/env/local.js` and run `npm start`
* You should get browser tab at http://127.0.0.1:8888/ with:

image

### DIFF
--- a/config/env/default.js
+++ b/config/env/default.js
@@ -238,4 +238,12 @@ module.exports = {
       dsn: '',
     },
   },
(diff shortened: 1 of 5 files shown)


hlibbabii (approved):
👍🏼

PR #2013, opened 2021-01-03
Cleanup invite code feature

simison:
Removing old code that's been un-used for years. If in any situation we might need this in future, it's here in the PR ready to be recovered.

image

🎉 

As it's now, it wouldn't work anyway because Maitre-app (service the feature was using) has changed too much and some parts of the code have already been removed previously.

Looks like someone already spent some time converting old code to React, and maybe some strings were also translated, and I'd hate to see that kind of unnecessary work happen again.

The code also still gets bundled and shipped to everyone, making initial site load slower.

#### Proposed Changes

* Removes old "invite" code feature

#### Testing Instructions

* Smoke test the app, especially signup. Everything should just work.


hlibbabii (approved):
Smoke-tested: works well 👍🏼

PR #2028, opened 2021-01-10
Adding mention of donations in form of media

chagai95:
Tanja asked me to add this.

#### Proposed Changes

* adding a bit of text with hyperlink

#### Testing Instructions

* go to https://www.trustroots.org/contribute and check Volunteering section

Fixes #

### DIFF
--- a/modules/pages/client/components/Contribute.component.js
+++ b/modules/pages/client/components/Contribute.component.js
@@ -59,6 +59,12 @@ export default function Contribute() {
                 </a>{' '}
                 to get an idea of what’s in the making and what you can do to
                 help out.
+                You could help Trustroots out with as little effort as{' '}
+                <a href="https://team.trustroots.org/Chat.html">
+                  donating your photo's
+                </a>{' '}
+                to use for our website or create content. If you're
+                a (semi-)professional photographer or have any questions you can contact us trough share@trustroots.org.
               </Trans>
             </p>
           </div>
--- a/modules/pages/client/components/Contribute.component.js
+++ b/modules/pages/client/components/Contribute.component.js
@@ -53,18 +53,15 @@ export default function Contribute() {
             <p>
               <Trans t={t} ns="pages">
                 Whether you’re a hardcore techie or know nothing of coding, we
-                can use your help! Head over to the{' '}
-                <a href="https://team.trustroots.org/Chat.html">
-                  volunteer chat
+                can use your help! See{' '}
+                <a href="https://team.trustroots.org/Volunteering.html">
+                  volunteer page
                 </a>{' '}
-                to get an idea of what’s in the making and what you can do to
-                help out.
-                You could help Trustroots out with as little effort as{' '}
-                <a href="https://team.trustroots.org/Chat.html">
-                  donating your photo's
-                </a>{' '}
-                to use for our website or create content. If you're
-                a (semi-)professional photographer or have any questions you can contact us trough share@trustroots.org.
+                get an idea of what’s in the making and what you can do to help
+                out. You could help Trustroots out with as little effort as
+                donating your photo&apos;s to use for our website or create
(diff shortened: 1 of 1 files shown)


TMC89:
Asked to make a proposition ;) But yeah same result.

modules/pages/client/components/Contribute.component.js
```diff
@@ -59,6 +59,12 @@ export default function Contribute() {
                 </a>{' '}
                 to get an idea of what’s in the making and what you can do to
                 help out.
+                You could help Trustroots out with as little effort as{' '}
+                <a href="https://team.trustroots.org/Chat.html">
+                  donating your photo's
+                </a>{' '}
+                to use for our website or create content. If you're
+                a (semi-)professional photographer or have any questions you can contact us trough share@trustroots.org.
```
simison, line 67: I think it would be best to just link to contact form here (`<a href="/contact">contact us</a>`) because we have a bit of tendency not to follow that email all the time, but we do follow support queue. Folks can then forward to right place from there if need.

modules/pages/client/components/Contribute.component.js
```diff
@@ -59,6 +59,12 @@ export default function Contribute() {
                 </a>{' '}
                 to get an idea of what’s in the making and what you can do to
                 help out.
+                You could help Trustroots out with as little effort as{' '}
+                <a href="https://team.trustroots.org/Chat.html">
```
simison, line 63: One way to contact will probably work best — two is just more confusing?

simison:
You can get Prettier pass here if you run `npm run reformat-files modules/pages/client/components/Contribute.component.js`. That should always happen when you `git commit` — not quite sure why that didn't happen here.

modules/pages/client/components/Contribute.component.js
TMC89: Okay, that's also fine. I thought about share because Jans and Shawn also gets those emails and I asked them to keep an eye on it. Whichever you prefer is fine :)

modules/pages/client/components/Contribute.component.js
TMC89: I haven't seen what is behind the hyperlink but the idea is that the Google form is behind it: that is what we created to collect the photos. The email is in case someone has questions afterwards or is a pro.

TMC89: The first part isn't my proposition. Only the second about donating photos. Behind "donating photo's" should be the link to the Google form we created for this.

TMC89:
This is the text after someone checked for grammer:
You could also help Trustroots out with as little effort as donating
your photos to use for our website or create content. If you're a
(semi-)professional photographer or have any questions, please contact
us  share@trustroots.org.

modules/pages/client/components/Contribute.component.js
TMC89: This is the text after someone checked for grammer:
You could also help Trustroots out with as little effort as donating
your photos to use for our website or create content. If you're a
(semi-)professional photographer or have any questions, please contact
us share@trustroots.org.

TMC89: What if we change the first sentence to: "One way to help us out is with little effort as donating your photos...."

TMC89:
Soooo, what's the status :)?

TMC89:
@chagai95 @simison not sure how I can help here

chagai95:
What's left to do? maybe make a list?

TMC89:
There's technical stuff to do right? I can't do that so I'm not sure what list you mean. 
I suggested that the Google form was included on the platform so OR members would find it when clicking on "contribute" OR it was mentioned in as a small hyperlink at the bottom/in the right menu :)

simison:
Test failure is unrelated, I'll merge this. Thanks folks!

chagai95:
Thank you!

PR #2029, opened 2021-01-13
Video call prompt

### LINKED ISSUE #2571: Nudge members to have video call before meeting
We had a discussion already about this on Slack I believe and now I got the same suggestion through support: add some info that hints members to make a video call before meeting. The member at support said it would be less uncomfortable when the platform suggests its, than asking for it themselves.

chagai95:
#### Proposed Changes

* Prompting the user to a video call before hosting

#### Testing Instructions

* click on the Yes, I can host! , you can see this message before sending your first message to someone in the messages layout

Fixes #2571

### DIFF
--- a/modules/messages/client/components/QuickReply.js
+++ b/modules/messages/client/components/QuickReply.js
@@ -44,11 +44,25 @@ export default function QuickReply({ onSend, onFocus }) {
           if (write) {
             onFocus();
           } else {
-            onSend(`
+            if (host) {
+              onSend(`
+              <p data-hosting="${host ? 'yes' : 'no'}">
+                <b><i>${content}</i></b>
+                <br>
+                <br>
+                Click here if you are interested in a quick face to face before you meet in person, set a time for a video call!
+                <br>
+                <br>
+                We set up a room for you here <a href = https://meet.jit.si/slkjfdhvreiuvsnf>meet.jit.si/slkjfdhvreiuvsnf</a>
+              </p>
+            `);
+            } else {
+              onSend(`
               <p data-hosting="${host ? 'yes' : 'no'}">
                 <b><i>${content}</i></b>
               </p>
             `);
+            }
           }
         }
 
--- a/modules/messages/client/components/QuickReply.js
+++ b/modules/messages/client/components/QuickReply.js
@@ -45,6 +45,17 @@ export default function QuickReply({ onSend, onFocus }) {
             onFocus();
           } else {
             if (host) {
+              let result = '';
+              const characters =
+                'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+              const charactersLength = characters.length;
+              for (let i = 0; i < 15; i++) {
+                result += characters.charAt(
+                  Math.floor(Math.random() * charactersLength),
+                );
+              }
+              alert(result);
+              const jitsi_random = result;
               onSend(`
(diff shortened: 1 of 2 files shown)


noahsmindfuck:
What about moving it to the right and out of the message window? In the are of "is this convo in the spirit of trustroots" and in mobile just make a green camera symbol to klick.

noahsmindfuck (changes requested):
What about moving it to the right and out of the message window? In the are of "is this convo in the spirit of trustroots" and in mobile just make a green camera symbol to klick.

Also: why is the whatsapp me in this commit?

chagai95:
> What about moving it to the right and out of the message window? In the are of "is this convo in the spirit of trustroots" and in mobile just make a green camera symbol to klick.
> 
What would be the purpose of that? It'll be much less visual and I didn't quite understand what you mean with the camera icon... The users need to be aware this is not just a video call button but a room they would both have to enter at the same time, which is why we need text to explain it.... This is unfortunately not a video calling feature that we are adding to the platform, that'll be a little bit more complicated...


> Also: why is the whatsapp me in this commit?

The WhatsApp is there because I didn't discard the changes, weird I didn't think VSC would push that, thanks for noticing!

chagai95:
Hey @simison do you think this is something you could merge as is, it's been stagnating for a while and I think it might be better on production the way it is as opposed to not at all?

sitarane:
Are we even allowed to integrate jitsi's server in an app like that? Did anyone check the jitsi fine prints?

Jitsi is nice and all to let people use it's infra for free, but there probably are limits...

chagai95:
I'm pretty sure it's allowed (otherwise we can use any other free jitsi instance out there or run our own). If we are going to merge this then I'll personally communicate this to the jitsi team and maybe they'll put us on their website as an open source community project which uses jitsi, then we'll definitely be in the clear, (otherwise we can remove it)

Also I don't know how integrated jitsi really is in this case we're basically just generation a random URL and telling people to meet...

modules/users/client/views/profile/profile-view.client.view.html
```diff
@@ -155,6 +155,14 @@
                   Send a message
                 </a>
               </li>
+              <li ng-if="app.appSettings.referencesEnabled">
+                <a
+                  ui-sref="profile.experiences.new({username: profileCtrl.profile.username})"
+                >
+                  <i class="fa icon-message-alt"></i>
+                  Contact me via Phone
```
sitarane, line 163: Hard coded string? Shouldn't we use i18n keys?

modules/messages/client/components/QuickReply.js
```diff
@@ -44,11 +44,36 @@ export default function QuickReply({ onSend, onFocus }) {
...
+              const characters =
+                'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+              const charactersLength = characters.length;
+              for (let i = 0; i < 15; i++) {
+                result += characters.charAt(
+                  Math.floor(Math.random() * charactersLength),
+                );
+              }
+              alert(result);
+              const jitsi_random = result;
```
sitarane, line 58: This risk intersecting with an existing jitsi room. I bet there's a way to just query jitsi for a random room and use that.

modules/users/client/views/profile/profile-view.client.view.html
chagai95: We should

PR #2031, opened 2021-01-15
Writing "response" experience with limited choice after the "initial" experience went public

hlibbabii:
#### Proposed Changes

Writing "response" experience with limited choice after the "initial" experience went public. 

Initial draft implementation: the tab where the recommendation can be chosen is not shown at all. Probably we don't want to do it like that, we should still give a possibility to report a person? Also, we should let the user know that the recommendation is "yes". (If we decide to keep recommendations)

- [x] disable tab on UI if the initial experience is public
- [x] change the `/api/my-experiences` endpoint to always return both A->B and B->A experiences if available
- [x] server-side validation had already been implemented.

#### Testing Instructions

* Log in as A; share experience with B; find the experience in the DB and change the `public` field to 'true'; log in as B and share experience with A; notice that you are not allowed to choose recommendation; look at the profile of A and notice that the left experience has `recommended` label
* Make sure experience cannot be shared twice with the same person

WIP #1493 

https://user-images.githubusercontent.com/2955794/104788013-6d445000-5791-11eb-9d2e-5cb9bca82177.mov

### DIFF
--- a/modules/references/client/components/CreateExperience.component.js
+++ b/modules/references/client/components/CreateExperience.component.js
@@ -125,9 +125,17 @@ export default function CreateExperience({ userFrom, userTo }) {
   );
 
   let tabs = [
-    { title: t('How do you know them'), component: interactionsTab },
-    { title: t('Recommendation'), component: recommendationTab },
-    { title: t('Feedback'), component: feedbackTab },
+    {
+      id: 'interactions',
+      title: t('How do you know them'),
+      component: interactionsTab,
+    },
+    {
+      id: 'recommendation',
+      title: t('Recommendation'),
+      component: recommendationTab,
+    },
+    { id: 'feedback', title: t('Feedback'), component: feedbackTab },
   ];
 
   if (!sharedOnTime) {
@@ -191,9 +199,9 @@ export default function CreateExperience({ userFrom, userTo }) {
         onSelect={() => {}}
         id="create-reference-tabs"
       >
-        {tabs.map(({ component, title }, i) => {
+        {tabs.map(({ id, component, title }, i) => {
           return (
-            <Tab eventKey={i} key={`${i}`} title={title} disabled>
+            <Tab eventKey={i} key={id} title={title} disabled>
               {component}
             </Tab>
           );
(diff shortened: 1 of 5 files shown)


hlibbabii:
TBD: nice error if the user starts sharing experience while it is still private, but submits it when it already went public? Probably not critical for this iteration?

modules/references/client/components/CreateExperience.component.js
```diff
@@ -169,15 +191,13 @@ export default function CreateExperience({ userFrom, userTo }) {
...
-        </Tab>
-        <Tab eventKey={1} title={t('Recommendation')} disabled>
-          {tabs[1]}
-        </Tab>
-        <Tab eventKey={2} title={t('Feedback')} disabled>
-          {tabs[2]}
-        </Tab>
+        {tabs.map(({ component, title }, i) => {
+          return (
+            <Tab eventKey={i} key={`${i}`} title={title} disabled>
```
simison, line 196: How come not just `key={i}`? 

Index isn't perfect since list/index can change based on `sharedOnTime`? https://reactjs.org/docs/lists-and-keys.html

hlibbabii: > How come not just `key={i}`?

You know that I like overcomplicating things :D

> Index isn't perfect key here since list/index can change based on `sharedOnTime`? https://reactjs.org/docs/lists-and-keys.html

Fixed that, thanks!

simison:
Needs a rebase :-)

simison (approved):
Tests well!

Tested with "expired" experience as well non-expired.

I noticed a couple warnings in console, not sure if related:

image

After those are sorted (if related), good to merge!

hlibbabii:
@simison I fixed the errors on the second screenshot. Regarding the first one, it complains that there is no `interactions` object in the displayed experiences, however from what I see there is no way to create an experience without it - there is both client and server-side validation. So, probably, such experience was added to the db directly?  Anyway, I guess it's not related, and also not such a big deal?

simison:
Yeah, sounds like some direct DB manipulation indeed 👍

PR #2032, opened 2021-01-16
Add prompt to share response under public experiences on the logged-in user's page

hlibbabii:
#### Proposed Changes

Add prompt to share responses under public experiences on the logged-in user's page.

<image src="https://user-images.githubusercontent.com/2955794/104791025-7b966a00-5799-11eb-9a66-3617eccbd53d.png" width=500/>

Didn't put much thought into the design, so implemented the easiest option. 

#### Testing Instructions

- Go to my profile -> experiences. Check that there is a link "Share experience also with them." below public experiences to which the logged-in user hasn't yet shared a response. Click the link and check that it goes to the experience sharing page with the user who shared the initial experience.

- Go to someone else's profile make sure that the link is not present below the public experiences with no response

WIP #1493

### DIFF
--- a/modules/references/client/components/ListExperiences.component.js
+++ b/modules/references/client/components/ListExperiences.component.js
@@ -54,11 +54,13 @@ export default function ListExperiences({ profile, authenticatedUser }) {
   const hasPublicExperiences = publicExperiences.length > 0;
   const hasPendingExperiences = pendingExperiences.length > 0;
 
+  const onReceiverProfile = authenticatedUser._id === profile._id;
+
   // No experiences
   if (!hasPendingExperiences && !hasPublicExperiences) {
     return (
       <NoContent icon="users" message={t('No experiences yet.')}>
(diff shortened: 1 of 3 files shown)


simison:
One issue — it's now double in this pending experience:

Screenshot 2021-01-17 at 18 20 37

Otherwise this is working great, thanks for adding it!

I'd prefer to separate the call-to-action to be separate from content, and less like a link. Something like this?

image

Here's HTML for that — comes after `panel-body`:
```html
<div class="panel-footer text-right">
  <a href="#" class="btn btn-default">Write about your experience</a>
</div>
```

hlibbabii:
> One issue — it's now double in this pending experience:

Oh, you are right, I didn't notice it.

thanks, I'll fix everything!

simison (approved):
Apart from one bug & suggested visual change, good to go! Merge when ready.

PR #2046, opened 2021-01-16
Translations update from Weblate

weblate:
Translations update from [Weblate](https://hosted.weblate.org/projects/trustroots/translation/) for Trustroots/translation.

Current translation status:

Weblate translation status

PR #2047, opened 2021-01-16
Update copy for consictency: facebook (Group) -> facebook group

hlibbabii:
Not to translate similar strings twice:

https://github.com/Trustroots/trustroots/pull/2046#discussion_r559050253


simison (approved):
⚡

PR #2076, opened 2021-01-31
Redirecting to edit profile

chagai95:
Not a perfect solution but a good start.

#### Proposed Changes

* adding subscribe to newsletter on about section

#### Testing Instructions

* go to arrow next to profile picture and click on about then scroll all the way down and chack out the Subscribe to Newsletter button

Fixes #

### DIFF
--- a/modules/pages/client/components/Home.component.js
+++ b/modules/pages/client/components/Home.component.js
@@ -382,6 +382,9 @@ export default function Home({ user, isNativeMobileApp, photoCredits }) {
                   <a href="https://shop.trustroots.org/">{t('Shop')}</a>
                 </li>
                 <li>
+                  <a href="https://www.trustroots.org/profile/edit/account/">{t('Subscribe to Newsletter')}</a>
+                </li>
+                <li>
                   <a href="/volunteering">{t('Volunteering')}</a>
                 </li>
                 <li>


simison:
I think the trouble is that unless they're logged in, they'll only see login form:

image

modules/pages/client/components/Home.component.js
```diff
@@ -381,6 +381,9 @@ export default function Home({ user, isNativeMobileApp, photoCredits }) {
                 <li>
                   <a href="https://shop.trustroots.org/">{t('Shop')}</a>
                 </li>
+                <li>
+                  <a href="https://www.trustroots.org/profile/edit/account/">{t('Subscribe to Newsletter')}</a>
```
simison, line 385: Can you remove the domain part? So just `/profile/edit/account` — thanks!

modules/pages/client/components/Home.component.js
```diff
@@ -381,6 +381,9 @@ export default function Home({ user, isNativeMobileApp, photoCredits }) {
                 <li>
                   <a href="https://shop.trustroots.org/">{t('Shop')}</a>
                 </li>
+                <li>
```
simison, line 384: Shall we wrap this in user-check `{user && ( <li>...</li> )}` so that we don't render it for un-logged folks?

simison (approved):
I'm not quite sure this will have any impact, but feel free to merge after addressing feedback.

Thanks for thinking about the problem!

modules/pages/client/components/Home.component.js
simison, line 385: Why _Newsletter_ with capital N? :-)

PR #2090, opened 2021-02-13
Fix: experience deep link in the list points to "references"

simison:
The date link was incorrectly linking to `/references` while it should be `/references`

image

#### Proposed Changes

* Fix correct URL

#### Testing Instructions

* See list of experiences and click on date -> should create a deep link to that experiences


simison:
I'll merge since this is such a simple fix.

PR #2093, opened 2021-02-15
Media: stop translating date literals

simison:
#### Proposed Changes

* Stop translating date literals — we shouldn't send strings like "Jul 22nd" for translation. We can use i18n-next date pattern instead https://www.i18next.com/translation-function/formatting
* Add one news article from 2020

#### Testing Instructions

* Open `/media` and compare with what's on trustroots.org — should match

### DIFF
--- a/modules/pages/client/components/Media.component.js
+++ b/modules/pages/client/components/Media.component.js
@@ -5,6 +5,9 @@ import { Trans, useTranslation } from 'react-i18next';
 export default function Media() {
   const { t } = useTranslation('pages');
 
+  const mediaDate = (...dateArgs) =>
+    t('{{date, MMMM Do}}', { date: new Date(...dateArgs) });
+
   return (
     <>
       <Board names="bokeh">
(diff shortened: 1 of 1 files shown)


modules/pages/client/components/Media.component.js
```diff
@@ -43,7 +46,10 @@ export default function Media() {
             <h4 className="media-list-divider text-muted">2017</h4>
             <dl className="dl-horizontal media-list">
               <dt>
-                <span className="text-muted">{t('October 13th')}</span>
+                <span className="text-muted">
+                  {/* October 13th */}
+                  {mediaDate(2017, 9, 13)}
```
simison, line 51: JS has month starting from 0 so that's why it can look odd, and that's why I added the comment for clarity.

simison:
Instead of lots of boiler-HTML, ideally we'd just loop through some kind of data-object, or have each media-line via a simple component. I just didn't wanna spend time on this now. 🙃

hlibbabii (approved):
👍🏼

Issue #2095, opened 2021-02-16
Search map: "Circles in common" doesn't get upgraded between profiles


simison:
On the new map:

> When I look at the map with users and go from one user to the other, Circles in common does not get updated, it stays the same as initiated with first user who’s details I look at.

<img width="508" alt="image" src="https://user-images.githubusercontent.com/87168/108054260-17481e00-7057-11eb-9878-80bbb4cfb1ab.png">

at https://github.com/Trustroots/trustroots/blob/b94d8080a440d36529e7aa9921b8d174192a63f5/modules/search/client/views/search-sidebar-results.client.view.html#L155-L159

Likely overlooked this in the big rewrite https://github.com/Trustroots/trustroots/pull/1329

simison:
Might also get fixed in React rewrite of this sidebar in https://github.com/Trustroots/trustroots/pull/1971

PR #2096, opened 2021-02-16
Fix: circles in common not getting updated

simison:
#### Proposed Changes

* Fixes circles in common for profiles not getting updated when opening profiles on the map

#### Testing Instructions

* open `/search`
* open profiles, observe circles in common being different after fix. Before fix they didn't change.

Fixes https://github.com/Trustroots/trustroots/issues/2095


simison:
Nope, this didn't fix it. :-/

PR #2106, opened 2021-02-20
Rename references to experiences

simison:
Recommend to look at individual commits for clarity.

#### Proposed Changes

* Rename all remaining "reference" things to "experience" things, across client-side and server side
* Not renaming the feature flag in config, just for convenience
* Note that also MongoDB collection is renamed. Recommend to rename it locally to keep previous test-data.

#### Testing Instructions

* Smoke test everything


modules/experiences/server/models/reference.server.model.js
```diff
@@ -61,19 +55,14 @@ const ReferenceSchema = new Schema({
   feedbackPublic: {
     type: String,
     trim: true,
-  } /*
-  feedbackPrivate: {
-    type: String,
-    trim: true
```
simison, line 67: Removed some commented-out code here.

modules/experiences/server/models/reference.server.model.js
```diff
@@ -6,20 +6,14 @@ const uniqueValidation = require('mongoose-beautiful-unique-validation');
...
     type: Date,
     default: () => Date.now(), // Date.now is wrapped for sinon.useFakeTimers()
     required: true,
   },
-  /*
-  modified: {
-    type: Date,
-    default: Date.now
-  },
-  */
```
simison, line 22: Removed some commented-out code here.

modules/experiences/client/less/experiences-new.less
```diff
@@ -1,7 +1,7 @@
 @import '~less/global-variables';
 @import '~bootstrap/less/mixins/buttons';
 
-.reference-new-tabs {
+.experience-new-tabs {
```
simison, line 4: This CSS seems not to be in use currently — I'll sort it out in separate PR.

hlibbabii (approved):
I smoke-tested, nothing seems to have broken. Btw, there are still things in `bin/fillTestData/Experiences.js` not renamed

simison:
Thanks! Good catch with test-fill data, I fixed that.

PR #2110, opened 2021-02-23
Experiences: add question mark to question

simison:
The same sentence is already being translated with a question mark here: 

https://github.com/Trustroots/trustroots/blob/d146c16f6547a5e3397ca4cc7fb164d28f8482fe/modules/experiences/client/components/create-experience/Interaction.js#L15

Screenshot 2021-03-07 at 02 21 15

#### Proposed Changes

* Add a question mark to the tab when adding new experience

#### Testing Instructions

* Add experience to someone and see the tabs


hlibbabii (approved):
👍🏼

PR #2121, opened 2021-02-28
Block users UI

antalAkos:
#### Proposed Changes

* Let users block and unblock each other from profile and message thread UI

#### Testing Instructions

* Open a profile
* Block them from their profile or message thread

  In profile
  image
  
  In profile via mobile
   image

 In message thread
   image

* You will see a banner informing that you've blocked them
    image

* Login as the other user
* You shouldn't now see the first user anymore nor be able to send them messages:

Part of https://github.com/Trustroots/trustroots/issues/1467


modules/support/client/components/BlockMemberLink.component.js
```diff
@@ -0,0 +1,38 @@
...
+export default function BlockMemberLink({ username }) {
+  const { t } = useTranslation('support');
+
+  async function handleBlockMember({ username }) {
+    const data = await api.block(username); 
+  }
+
+  return (
+    <a
+      href={`/support?report=${username}`}
```
simison, line 25: The `href` isn't needed here, so we can probably switch the whole `<a>` tag to a `<button>` instead?

To get it to render like it's a link you can add `btn btn-link` into `className`. Not sure if it does the trick but let's see how it looks :-)

modules/support/client/components/BlockMemberLink.component.js
```diff
@@ -0,0 +1,38 @@
...
+      className="small text-muted"
+      onClick={() => handleBlockMember(username)}
+      aria-label={t('Block member {{username}} to support', { username })}
```
simison, line 28: This line is for blind keyboard users. Mentioning support doesn't make sense here, how about just `Block member {{username}}`?

modules/support/client/components/BlockMemberLink.component.js
```diff
@@ -0,0 +1,38 @@
+/**
```
simison, line 1: I think the `user` module would feel a more logical place for this component instead of the support module, what do you think?

modules/support/client/components/BlockMemberLink.component.js
```diff
@@ -0,0 +1,38 @@
...
+
+import * as api from '../api/block.api';
+
+
+
+
+export default function BlockMemberLink({ username }) {
+  const { t } = useTranslation('support');
+
+  async function handleBlockMember({ username }) {
```
simison, line 19: The `{}` brackets here make JS think this the function parameter should be an object, with key `username`:

```js
{ username:  'something' }
```

so changing it to ` handleBlockMember(username)` will help.

modules/support/client/components/BlockMemberLink.component.js
```diff
@@ -0,0 +1,35 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import '@/config/client/i18n';
+import { useTranslation } from 'react-i18next';
+
+import * as api from '../api/block.api';
+
+export default function BlockMemberLink({ username }) {
+  const { t } = useTranslation('support');
```
simison, line 14: In https://github.com/Trustroots/trustroots/pull/2121/files#r584354289 I suggested moving to `user`module instead, so just noting that translation namespace would need changing in that case as well.

modules/support/client/api/block.api.js
```diff
@@ -0,0 +1,6 @@
+import axios from 'axios';
+
+export async function block(username) {
+  alert('alert2' + username);
+  return await axios.post(`/api/blocked-users/${username}`);
```
simison, line 5: Looks like the API route actually expects `put()` instead of `post()`:

https://github.com/Trustroots/trustroots/blob/48921b398eac5dcf2dd074b3b0afb5a3e6fa8416/modules/users/server/routes/users-block.server.routes.js#L14-L18

modules/users/client/views/profile/profile-view-basics.client.view.html
```diff
@@ -7,3 +7,8 @@
   ng-if="::app.user._id !== profileCtrl.profile._id"
   username="::profileCtrl.profile.username"
 ></report-member-link>
+<br>
```
simison, line 10: These links felt a bit close and hard to tap with big fingers on mobile. Let's try a bigger space with `<br/><br/>`?

modules/users/server/policies/users.server.policy.js
```diff
@@ -133,7 +133,7 @@ exports.invokeRolesPolicies = function () {
         },
         {
           resources: '/api/blocked-users/:username',
-          permissions: ['put', 'delete'],
+          permissions: ['post', 'delete'],
```
simison, line 136: `put` was correct: https://github.com/Trustroots/trustroots/pull/2121/files#r584355524

simison (approved):

PR #2123, opened 2021-03-01
Fix: circles in common not getting updated

saemy:
#### Proposed Changes

* Fixes circles in common for profiles not getting updated when opening profiles on the map

#### Testing Instructions

* open `/search`
* open profiles, observe circles in common being different after fix. Before fix they didn't change.

Fixes: #2095


simison (approved):
Excellent, code looks good and tests well! 

Thanks!

PR #2128, opened 2021-03-06
Update/sparkpost webhooks

simison:
#### Proposed Changes

* Update `eventTypes` list from [docs](https://www.sparkpost.com/docs/tech-resources/webhook-event-reference/), in two commits: https://github.com/Trustroots/trustroots/pull/2128/commits/7dabf6ce6607f222a37d61a73d0ec10748549e44 (alphabetize), https://github.com/Trustroots/trustroots/pull/2128/commits/531e7d106cc29e4d18927f6b5f4094317e79e905 (Fresh list of event names)
* Move `campaignId` and `country` to tags from meta object, so that we can query against them in Grafana. https://github.com/Trustroots/trustroots/pull/2128/commits/9a17281853f9c78ff947646f6f15966710ec1b79
* Add `reason` data (for `bounce` and other rejection events), see [docs](https://www.sparkpost.com/docs/tech-resources/webhook-event-reference/)
* Add `mailboxProvider` data (e.g. "Gsuite" etc) to help detecting if we're blocked by any specific providers e.g. due to email blacklists.
* Modernize code, don't use so much `lodash` etc

#### Testing Instructions

A bit tricky to test from end-to-end... 😅 

You could push to the webhook API endpoint an example data object and observe it from console get recorded to InfluxDB.

I'd be fine just for code-review and confirm that the test that I updated touches all the code paths.

I'll observe this in production after deploying and I'll be ready to revert if it behaves unexpectedly.

Note that we don't really run Stathat anymore so that code and test is obsolete — we should in fact just strip the code out.

### DIFF
--- a/modules/sparkpost/server/controllers/sparkpost-webhooks.server.controller.js
+++ b/modules/sparkpost/server/controllers/sparkpost-webhooks.server.controller.js
@@ -171,7 +171,7 @@ exports.processAndSendMetrics = (event, callback) => {
 /**
  * Basic authentication middleware
  */
-exports.basicAuthenticate = function (req, res, next) {
+exports.basicAuthenticate = (req, res, next) => {
   // Get the basic auth credentials from the request.
   // The Authorization header is parsed and if the header is invalid,
   // undefined is returned, otherwise an object with name and pass properties.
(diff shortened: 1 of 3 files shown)


simison:
Depends on removing Stathat first: https://github.com/Trustroots/trustroots/pull/2129

PR #2133, opened 2021-03-06
Remove/angular directive host offer

simison:
#### Proposed Changes

* Use existing React component for mobile view for hosting info on profiles
* Remove now un-used Angular directive
    image

#### Testing Instructions

* Open user's own profile in mobile view, open hosting info
* Open other user's profile in mobile view, open hosting info
* Note that the dropdown doesn't work in current `master` (same issue as with language dropdown I assume)
* Everything should be identical as in production otherwise

Screenshot 2021-03-07 at 01 04 27


simison:
This should be relatively safe since the React component is already used for desktop — I'll merge to get to test more.

PR #2134, opened 2021-03-06
Fix hosting button

simison:
#### Proposed Changes

When looking at your own profile, this button was broken:

https://user-images.githubusercontent.com/87168/110224467-7f668300-7ee4-11eb-84a4-cf791fbb3407.mov

The dropdown component is worth fixing but another problem here was that it was needlessly complicated. We don't need a dropdown and all the pre-setting hosting offer in editor, we can just link to it with simple button.

This simplifies things quite a bit. The whole Offers component could do a bit of renovation/modernizing. :-)

#### Testing Instructions

- Test own profile with different statuses
- Test profile that doesn't have hosting offer saved in DB (should show up as "no"
- Test other profiles
- Clicking it should bring to messages for other profiles, or editing the offer for your own profile

Screenshot 2021-03-07 at 01 24 39

### DIFF
--- a/modules/offers/client/components/OfferStatusButton.js
+++ b/modules/offers/client/components/OfferStatusButton.js
@@ -1,0 +1,39 @@
+// External dependencies
+import { useTranslation } from 'react-i18next';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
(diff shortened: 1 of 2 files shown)


hlibbabii (approved):
tested, works well 👍🏼

PR #2135, opened 2021-03-07
Update copy in experiences

simison:
#### Proposed Changes

* Add "Based on your experience, ..." in recommendation questions
* Change "I don't know" to "skip" in recommendation question
* Always show "It's extremely important you report anyone behaving against rules to us." instead of showing it only when people choose "no"
* Correct typo under feedback field

#### Testing Instructions

* Write experience, and observe steps work

image
image

### DIFF
--- a/modules/experiences/client/components/create-experience/Feedback.js
+++ b/modules/experiences/client/components/create-experience/Feedback.js
@@ -57,7 +57,7 @@ export default function Feedback({
         ></textarea>
         <span className="help-block">
           {t(
-            'The answer you write will be publicaly available on their profile',
+            'The answer you write will be publicly available on their profile.',
           )}
           <br />
         </span>
--- a/modules/experiences/client/components/create-experience/Feedback.js
+++ b/modules/experiences/client/components/create-experience/Feedback.js
@@ -12,8 +12,6 @@ export default function Feedback({
 
   /*
    * Functions passing strings to translation fuction for translation scripts
-   * Added not, hate, maybe to differenciate between the 4 options.
-   * This should be changed to something more fitting (perferably fitting experiences concept)
    */
   const getRecommend = (recommendCode, report) => {
     switch (recommendCode) {
@@ -21,7 +19,7 @@ export default function Feedback({
         return t('Did you enjoy their cooking? singing?');
       case 'no':
         return report
-          ? t('Did you hate their cooking? singing?')
+          ? t('Did you not like their cooking? singing?')
           : t('Did you not enjoy their cooking? singing?');
       case 'unknown':
         return t('Did you maybe enjoy their cooking? singing?');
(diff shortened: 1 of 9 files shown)


modules/experiences/client/components/create-experience/Feedback.js
```diff
@@ -57,7 +57,7 @@ export default function Feedback({
         ></textarea>
         <span className="help-block">
           {t(
-            'The answer you write will be publicaly available on their profile',
+            'The answer you write will be publicly available on their profile.',
```
simison, line 60: Just a typo correction here:

Screenshot 2021-03-07 at 01 53 39

hlibbabii:
I like "Skip" much better than "I don't know"! Maybe make it grey?

simison:
Thanks @hlibbabii ! I tried but I disliked how it became less obvious it's actionable. We also have a lot of different styles for "click me" type things (even in just this one view) so I'd like to try limit adding more. 😅 

 
Screenshot 2021-03-07 at 12 40 58

Your sentiment might've been "can it be separated a bit from the other two", for which I tried this — what do you think?

Screenshot 2021-03-07 at 12 42 55

Simple empty space goes long way, too:

image

simison:
As a side note, those 3-step-styles have regressed at some point and need tweaking. Active tab shouldn't look like a button.

They were clickable at some point but looks like no longer.

hlibbabii:
> Your sentiment might've been "can it be separated a bit from the other two",

exactly! 

Yeah, I think some visual separation would be nice, and your suggestions are good, I am thinking about it but still not sure what's the best

simison:
I'll add the line so that we get to play and feel it little bit.

noahsmindfuck (changes requested):
I am not sure. I like the "Based on your experience" thing, generally, but wasn't it that we would rather ask them:
"Besides your personal experience, would you still recommend others to host them?" or
"Would you recommend <username> to others, independent from your personal experience with them?

The feeling that I have is that we - with my version of the question - support to not rate a person just on one experience and to make the experience less about the other. This way we are already starting to make the experience about us and we recommend the other person unrelated to our experience.

With the question of @simison we are doing the opposite and we create an environment where the single experience with a person is dominating over the public persona of the other. I do not refer to abuse of the experience system here, I mean that it even happens subconsciously and that I would prefer not to create a place where we are rating members. :)

I very much like the seperation with a | for the I don't know and I very much agree on @simison for the styling of the tabs.

simison:
@noahsmindfuck yasss I agree with those copy suggestions — I was hoping someone could chime in with more specific suggestion than what I could remember from the call. :-)

> "Besides your personal experience, would you still recommend others to host them?"

This sounds clear 👍

simison:
Updated copy based on @noahsmindfuck 's feedback as well added small changes suggested by @MrSalami at https://github.com/Trustroots/trustroots/pull/2148

noahsmindfuck:
Maybe change it to
>It is very important for us that you report anyone who breaks the [https://www.trustroots.org/rules rules]!

simison:
@noahsmindfuck I did want to, but at the same time I don't want to make it unclear what happens by clicking links while filling in this form. Needs a modal methinks.

noahsmindfuck:
@simison we could add the "opens in new tab" thingy symbol. I think its quite a convention to link to the thing that you mention ant that ppl know that but I might be wrong.

simison:
Yep, I think we could try how that feels.

PR #2148, opened 2021-03-08
Changes to spelling and grammar in sharing your experience 

MrSalami:
#### Proposed Changes

**Change "Would you like to describe something about them?" to "Would you like to describe something about your experience with them"** 

**Change 'publicaly' (wrong spelling) to 'publicly'**.

_Reasoning is to support the conversations had about not giving a direct reference to someone, which the sentence implies in saying 'about them.' If it is about the experience, then adding the language of "your experience with" would help guide the user to sharing that experience as opposed to a description of 'them.'_ 

image

**Change 'against rules' which is wrong grammar, and add 'the' or 'the community' or 'the community rules and standards'** 
_Just don't write is as against rules, it needs a 'the' before it at least._ 

**Add 'the' before moderators** 
_while ok to just leave it as moderators, it adds a bit of authority to add 'the'_

**Change the statement 'please write in English if possible." to "Please write in English if possible, thank you" and take out period**

_The person who is reporting to moderators is about to write about a bad experience. Adding thank you will communicate to this person that the moderators are empathetic and hopefully give them the added confidence to writing the bad experience as opposed to not. "Will this moderator hear me? Will they do something?" is probably a question they are wondering as they decide to write and report. That is why I suggest adding a thank you at the end._ 

image

**Change 'hate to 'not like'** 

_Hate is a strong word and does not seem necessary to use in this context._ 

image

#### Testing Instructions

*

Fixes #


simison:
Thanks! I incorporated this feedback in https://github.com/Trustroots/trustroots/pull/2135

PR #2158, opened 2021-03-13
Remove Stalebot

simison:
#### Proposed Changes

* Remove Stalebot. I suspect this has been less helpful than useful. It's true we have old things to clean out, but we also have very old things we'd like to keep.

#### Testing Instructions

* N/A


hlibbabii (approved):
👍🏼

Issue #1334, opened 2020-03-27
[Overview] React migration


mrkvon:
_This is an issue to organize a bigger picture of React migration. It should serve as overview of the work and crossroad to more specific topics_

- To add something to this issue, please edit this description directly, if you can.
- Inform us on slack or in comments or edit the issue to grab things you want to work on
- Add [`React` label](https://github.com/Trustroots/trustroots/labels/react) to any issue or PR you may open that is related to React migration.
- Add your stuff to [Framework refactor](https://github.com/Trustroots/trustroots/projects/4) project, too.

### Components to migrate:

_this should be broken down into subtasks and completed more, maybe in separate issues_

Here's the list of our modules:

#### admin

**DONE** already implemented in react

#### core
- [x] 404 https://github.com/Trustroots/trustroots/pull/1957
- [ ] [Push notification question](https://github.com/Trustroots/trustroots/blob/2b57c46fafd9b48015dd253d07a383283ecfdfb0/modules/core/client/views/push-notification-question-modal.client.view.html)
- [x] [Header](https://github.com/Trustroots/trustroots/blob/86f9c2d88dd52e7fdfa9062cda915c1548ac5343/modules/core/server/views/partials/header.server.view.html) WIP PR: https://github.com/Trustroots/trustroots/pull/2171
- [ ] [Footer](https://github.com/Trustroots/trustroots/blob/86f9c2d88dd52e7fdfa9062cda915c1548ac5343/modules/core/server/views/partials/footer.server.view.html)

#### messages

- [x] inbox #1171 (@nicksellen)
- [x] threads #1173 (@nicksellen)
- [x] unread messages counter https://github.com/Trustroots/trustroots/pull/1370 (@nicksellen)
- [x] reference question next to message thread (aka references-thread)  https://github.com/Trustroots/trustroots/pull/1978

#### offers

- [ ] **TODO** Meetup pages (https://www.trustroots.org/offer/meet)
- [ ] **WIP** Host offer #1301 
- [x] Hosting & meet map https://github.com/Trustroots/trustroots/pull/1156

#### pages

- [x] (modules/pages/client/views) https://github.com/Trustroots/trustroots/pull/1412

#### Experiences

- [x] Implemented in react from the start

#### search

- [x] map  https://github.com/Trustroots/trustroots/pull/1329
- [ ] map sidebar, WIP: https://github.com/Trustroots/trustroots/pull/1971

#### statistics
[...]

PR #2171, opened 2021-03-14
Reactify header

simison:
#### Proposed Changes

* Rewrite header in React, allowing it to be translated

Screenshot 2021-04-03 at 19 08 01

#### Testing Instructions

* Ensure header collapses and functions in different screen widths
* Ensure dropdowns work
* Ensure active pages get marked as "active" visually; inside dropdowns, I didn't implement this as it was a bit elaborate. We can get this feature back with a different, React-based router easier.
* Ensure keyboard navigation works (walking through links and dropdowns with tab/arrows/enter)

Part of https://github.com/Trustroots/trustroots/issues/1334


simison:
I need to still ensure this fully works but it's pretty far along already so I'd welcome reviews/testing.

I still need to add testing instructions + PR description.

modules/pages/client/components/Navigation.component.js
```diff
@@ -147,7 +147,7 @@ export default function Navigation({ user, onSignout, isNativeMobileApp }) {
               <a
```
simison, line 147: Fixes in this file I can PR separately as they're unrelated.

modules/core/server/views/partials/header.server.view.html
```diff
@@ -1,270 +1,6 @@
-<header
-  id="tr-header"
-  role="banner"
-  class="navbar navbar-default navbar-fixed-top hidden-print"
+<app-header
+  is-i18n-enabled="app.appSettings.i18nEnabled"
```
simison, line 2: We can probably clean out the i18n feature flag out and thus no need to pass this config via props.

simison: PR for feature flag-removal: https://github.com/Trustroots/trustroots/pull/2217

modules/users/client/components/InterfaceLanguagePanel.component.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import LanguageSwitch from '@/modules/core/client/components/LanguageSwitch.component';
+import LanguageSwitch from '@/modules/core/client/components/LanguageSwitch';
```
simison, line 2: This component isn't used anymore in Angular templates, so we can remove `.component.js` part in the file — thus build won't turn it into Angular component anymore.

Issue #397, opened 2016-08-17
Allow only people from my circles find me


simison:
- User joins circles
- User checks "be visible only to people from my circles" opt-in at profile settings
- People who don't belong to those circles can't find user's hosting offer via the map. See [list method](https://github.com/Trustroots/trustroots/blob/master/modules/offers/server/controllers/offers.server.controller.js#L370).
- They can still see the user via direct profile link

So from member perspective, if they're in Cyclist and in Family circles, only people who are in either one of those circles could see them and other way around.

If I want only people from Cyclists to contact me, I would stay member only of that circle.

PR #2177, opened 2021-03-16
Allow only people from my circles find me

saemy:
#### Proposed Changes

* When editing a hosting offer, users can opt-in to not appear on the map of users with which they do not have any circles in common.

#### Testing Instructions

1. Manual
  * User 1: Edit profile -> Locations -> Modify your hosting location -> check "People that are not in any of my circles should not find me."
  * User 2 (that does not have any circles in common with user1): The hosting offer of user1 no longer appears on `/search`.
2. Automatic
  * Run `npm test` to run the additional tests.

Fixes: #397

### DIFF
--- a/modules/offers/client/views/offer-host-edit.client.view.html
+++ b/modules/offers/client/views/offer-host-edit.client.view.html
@@ -124,6 +124,30 @@
                     </span>
                   </div>
                 </fieldset>
+
+                <br /><br />
+
+                <fieldset
+                  class="offer-showOnlyInMyCircles"
+                  ng-if="offerHostEdit.offer.status !== 'no'"
+                >
+                  <legend>
+                    <h4 id="showOnlyInMyCircles">
+                      Available only in my circles?
+                    </h4>
+                  </legend>
+                  <div class="checkbox">
+                    <label>
+                      <input
+                        type="checkbox"
+                        name="newsletter"
+                        ng-model="offerHostEdit.offer.showOnlyInMyCircles"
+                      />
+                      People that are not in any of my circles should not find
+                      me.
+                    </label>
+                  </div>
+                </fieldset>
               </div>
             </div>
           </div>
--- a/modules/offers/client/views/offer-host-edit.client.view.html
+++ b/modules/offers/client/views/offer-host-edit.client.view.html
@@ -127,20 +127,14 @@
 
                 <br /><br />
 
-                <fieldset
-                  class="offer-showOnlyInMyCircles"
-                  ng-if="offerHostEdit.offer.status !== 'no'"
-                >
+                <fieldset ng-if="offerHostEdit.offer.status !== 'no'">
                   <legend>
-                    <h4 id="showOnlyInMyCircles">
-                      Available only in my circles?
-                    </h4>
+                    <h4>Available only in my circles?</h4>
                   </legend>
                   <div class="checkbox">
                     <label>
                       <input
                         type="checkbox"
-                        name="newsletter"
                         ng-model="offerHostEdit.offer.showOnlyInMyCircles"
                       />
                       People that are not in any of my circles should not find
--- a/modules/offers/server/controllers/offers.server.controller.js
+++ b/modules/offers/server/controllers/offers.server.controller.js
@@ -610,11 +610,8 @@ exports.list = function (req, res) {
 
   // Filter out users that do not share any circles with the authenticated user
   // and chose to not appear in those searches.
-  const showOnlyInMyCirclesQueries = [
-    { showOnlyInMyCircles: false },
-    { showOnlyInMyCircles: { $exists: false } },
-  ];
-  req.user.member.forEach(function (membership) {
+  const showOnlyInMyCirclesQueries = [{ showOnlyInMyCircles: false }];
+  req.user.member?.forEach(function (membership) {
     // Add all the circles that the authenticated user is member of. One of them
     // must match for an offer to appear in the search result.
     showOnlyInMyCirclesQueries.push({
--- a/modules/offers/server/controllers/offers.server.controller.js
+++ b/modules/offers/server/controllers/offers.server.controller.js
@@ -36,6 +36,7 @@ const publicOfferFields = [
   'location',
   'updated',
   'validUntil',
+  'showOnlyInMyCircles',
 ];
 
 // Offer fields users can modify
@@ -46,6 +47,7 @@ const allowedOfferFields = [
   'maxGuests',
   'location',
   'validUntil',
+  'showOnlyInMyCircles',
 ];
 
 /**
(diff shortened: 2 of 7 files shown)


simison:
This is great!

I'll try to get into reviewing this over the weekend! 

This seems like a good first step towards this and none of the following ponderings area meant as blocking. Just some very very brief thoughts:

- We might wanna work on making the copy of the toggle and the header above it super-duper clear since people might not have context what it exactly means? Can we help clarity e.g. by listing their circles under the toggle? (I'll think about the copy but suggestions welcome!)
- Was pondering if this makes sense as user setting, or setting in the offer like it's here? Cons and pros with both;
   - Where do members expect to set, or change the setting?
   - Should it apply also to meet-offers, or just hosting offers?
   - Should the entire profile be invisible to other circles? (I don't think so but curious for thoughts)
- Once set, can we surface this somehow in the UI / emails /etc afterwards?

modules/offers/tests/server/offer-search.server.routes.tests.js
```diff
@@ -1147,6 +1147,135 @@ describe('Offer search tests', function () {
...
+          user3.save(done);
+        },
+
+        // Update the hosting offers.
+        function (user3, done) {
+          // Save hosting offer 1 (user2, showOnlyInMyCircles=true).
+          const o1 = new Offer(offer1);
+          o1.user = user2Id;
+          o1.location = testLocations.Europe.location;
+          o1.showOnlyInMyCircles = true;
```
simison, line 1211: An alternative way to write this would be:

```js
const o1 = new Offer({
  ...offer1,
  user: user2Id,
  location: testLocations.Europe.location,
  showOnlyInMyCircles: true,
});
```

Up to you!

modules/offers/client/views/offer-host-edit.client.view.html
```diff
@@ -124,6 +124,30 @@ <h4 id="maxGuests">How many guests can you accommodate?</h4>
...
                 </fieldset>
+
+                <br /><br />
+
+                <fieldset
+                  class="offer-showOnlyInMyCircles"
+                  ng-if="offerHostEdit.offer.status !== 'no'"
+                >
+                  <legend>
+                    <h4 id="showOnlyInMyCircles">
```
simison, line 135: No need ID here — other sections had ID because of `aria-labelledby="maxGuests"` needs.

modules/offers/client/views/offer-host-edit.client.view.html
```diff
@@ -124,6 +124,30 @@ <h4 id="maxGuests">How many guests can you accommodate?</h4>
                     </span>
                   </div>
                 </fieldset>
+
+                <br /><br />
+
+                <fieldset
+                  class="offer-showOnlyInMyCircles"
```
simison, line 131: No need `class` here since there is no CSS applying to this CSS selector either.

modules/offers/server/controllers/offers.server.controller.js
```diff
@@ -608,6 +608,25 @@ exports.list = function (req, res) {
     }
   }
 
+  // Filter out users that do not share any circles with the authenticated user
+  // and chose to not appear in those searches.
+  const showOnlyInMyCirclesQueries = [
+    { showOnlyInMyCircles: false },
+    { showOnlyInMyCircles: { $exists: false } },
```
simison, line 615: This is fine with `$exists` — that said, alternatively we could run a query to add this field as `false` to _every_ existing offer in the database, and have the query simpler.

modules/offers/server/controllers/offers.server.controller.js
```diff
@@ -608,6 +608,25 @@ exports.list = function (req, res) {
...
+  ];
+  req.user.member.forEach(function (membership) {
```
simison, line 617: We currently have about 13K members without `member` key in their profile — I can run query to add it to everyone so that this won't fail due to that inconsistency.

First it would be good to add `default: []` to here though:

https://github.com/Trustroots/trustroots/blob/5daa64f6af253ce4274bfdf5719cddf2b88cc109/modules/users/server/models/user.server.model.js#L337-L340

simison (approved):
Testing really well, neat! Code looks good too. Just left a couple data consistency questions that we might need to handle before deploying.

Feel free to hit squash+merge even without changes, I'm happy to handle those database questions before deploying.

modules/offers/server/controllers/offers.server.controller.js
saemy: I'll add the default. I don't know what is better to have, the `$exists` and the `req.user.member?.forEach(...)` or force-adding those attributes into the db? (note `req.user` is known to exist as we verify that at the very start of that function so we don't need the `?` behind `req.user`).

simison: `req.user` will always exist and thus `?` is needed only in `member?`, correct! :-)

simison: I'll add those both (`member` and `showOnlyInMyCircles`) attributes to the DB anyway, just because I want to have the data as predictable as possible. That said, it's good to be a little bit defensive in coding when it makes sense, as long as it's not to the point that it would hide genuine bugs.

Adding `$exists` in the query makes the query a tiny bit slower/more complex so I'd probably leave that out to keep things simple.

saemy:
> This seems like a good first step towards this and none of the following ponderings area meant as blocking. Just some very very brief thoughts:
> 
> * We might wanna work on making the copy of the toggle and the header above it super-duper clear since people might not have context what it exactly means? Can we help clarity e.g. by listing their circles under the toggle? (I'll think about the copy but suggestions welcome!)

Good idea. I'll try to push some change for this in the next days.

> * Was pondering if this makes sense as user setting, or setting in the offer like it's here? Cons and pros with both;
>   
>   * Where do members expect to set, or change the setting?

I was thinking about that as well. Honestly, I didn't consider meetups yet. However, I think having it on the offer allows one to have e.g. a public meetup and a circle-only hosting offer.  

>   * Should it apply also to meet-offers, or just hosting offers?

I think we could also extend the meet-offers to have this property.

>   * Should the entire profile be invisible to other circles? (I don't think so but curious for thoughts)

Off the bat I would say no, too.

> * Once set, can we surface this somehow in the UI / emails /etc afterwards?

Yes, please. :smile:

modules/offers/server/controllers/offers.server.controller.js
saemy: How do we ensure that these attributes are also added to existing databases of developers?
Is there some sort of database update scripts?

simison: Here are the database migration scripts: https://github.com/Trustroots/trustroots/tree/dd7ee07ab92f466d41c15df8f6a8852cc1377e2b/migrations

But unlikely readme says, those don't run on `npm start` automatically anymore and need to be run manually.

I wouldn't worry too much about locale databases — it's quick to wipe out the database and re-populate it with new test data.

simison:
I'll merge this already, and we can work on follow-ups in other PRs. :-) I'll make a note to add the field on all database entries before deploying this.

simison:
Query to update all offers before this patch:

```js
db.getCollection('offers').update({}, { $set: { 'showOnlyInMyCircles': false } }, false, true );
```

PR #2181, opened 2021-03-21
Copy: use localized format for date

simison:
#### Proposed Changes

* Change to localized date format so that translators don't need to change this according to their own locale manually.

From https://momentjs.com/docs/#/displaying/format/

Month name, day of month, year | LL | September 4, 1986
-- | -- | --
  | ll | Sep 4, 1986

#### Testing Instructions

* Add contact, then remove/revoke them:
    Screenshot 2021-03-21 at 16 29 54
    Screenshot 2021-03-21 at 16 30 05
* Open any profile and see in the sidebar when they joined
    
    Screenshot 2021-03-21 at 16 29 11

* See "launched" text at http://localhost:3000/media
* See "Enabling the latent trust" text at http://localhost:3000/statistics


modules/contacts/client/components/RemoveContact.js
```diff
@@ -21,23 +21,23 @@ export default function RemoveContact({
...
   const isConfirmed = contact.confirmed;
-  const created = new Date(contact.created);
+  const date = new Date(contact.created);
 
   if (isConfirmed) {
     // Remove confirmed contact
     labelTitle = t('Remove contact?');
     labelConfirm = t('Yes, remove contact');
-    labelTime = t('Connected since {{created, MMM D, YYYY}}', { created });
+    labelTime = t('Connected since {{date, LL}}', { date });
```
simison, line 30: I think this was broken already just because it was `created` and not `date` which causes the right filter to work in i18n-next.

simison:
On 2nd thought, maybe best if I don't touch those `public/locales/LANGUAGE/**` files and only modify source file strings? Would that avoid rebase issues in Weblate like we had in past, @hlibbabii do you know?

Issue #1485, opened 2020-05-23
[Overview] i18n / translations


nicksellen:
We made a lot of progress with i18n since the [original discussion](https://github.com/Trustroots/trustroots/issues/492), but still some threads to draw together to get it finally up and running:

- [x] work out how and when to extract the translations from the source (see https://github.com/Trustroots/trustroots/issues/1486)
- [x] work out how to incorporate the changes back from Weblate (see https://github.com/Trustroots/trustroots/issues/1487)
- [x] make/fix/update/tweak the language selecting flow (maybe it's fine, but lets check! it's enabled on https://dev2.trustroots.org/ (but don't expect to see all the language content)
- [ ] update guide from dev perspective (https://team.trustroots.org/i18n.html) (see https://github.com/Trustroots/community/issues/87)
- [x] create/update guide from translator perspective (is pretty much ready --> https://team.trustroots.org/Translating-Getting-Started.html - just awaiting some review, ask Martin in slack) -- still needs to be updated?
- [ ] make the rest of the site content translatable
   - [ ] Clientside translations depend on React migration. See #1334
   - [ ] Server side translations (mostly emails & error messages)
- [x] Support Right-To-Left (RTL) languages like Hebrew.

PR #2296, opened 2021-05-10
Add RTL language support

simison:
#### Proposed Changes

* Adds Right-to-Left language support, relevant for locales like Arabic, Hebrew, Pashto, Persian, Sindhi, Syriac, Dhivehi, Urdu, Yiddish, etc.

See https://www.w3.org/International/questions/qa-html-dir

#### Testing Instructions

Run `npm run build` (as building itself doesn't currently work on dev build, needs a bit more work but I think it can be separate PR)

After first run, everything should work on both `npm start` and `npm run start:prod`

* Open language switcher and switch to Hebrew
    
    image
* Page turns flipped right-to-left:
    
    image

* Notice different CSS tag getting appended at the end of `<head>`; if you switch between multiple languages, the CSS ever loads just once and isn't removed in between.
* Notice `lang` and `dir` get updated in HTML tag in DOM. `dir="ltr"` puts loaded CSS in affect. 
* Refresh the page, RTL CSS should remain. This is because language code is stored in your browser as a cookie and in user settings when logged in. On initialization, the CSS gets loaded again.
    * Ideally this would be from server-side but client-side is fine for now methinks
* There is more work to do how CSS rendering looks like but that's good for follow-up. 
  
Part of https://github.com/Trustroots/trustroots/issues/1485

### DIFF
--- a/config/shared/locales.json
+++ b/config/shared/locales.json
@@ -73,8 +73,7 @@
     "label": "עִבְרִית",
     "code": "he",
     "english": "Hebrew",
-    "production": true,
-    "rtl": true
+    "production": true
   },
   {
     "label": "Türkçe",
--- a/config/shared/locales.json
+++ b/config/shared/locales.json
@@ -73,7 +73,8 @@
     "label": "עִבְרִית",
     "code": "he",
     "english": "Hebrew",
-    "production": false
+    "production": true,
+    "rtl": true
   },
   {
     "label": "Türkçe",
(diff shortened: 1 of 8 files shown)


config/shared/locales.json
```diff
@@ -73,7 +73,8 @@
     "label": "עִבְרִית",
     "code": "he",
     "english": "Hebrew",
-    "production": false
+    "production": true,
```
simison, line 76: TODO before merging: set back to `false`

PR #2378, opened 2021-07-14
Point volunteers directly to the volunteering page

mariha:
A person who is interested in volunteering has to do many hops until they get to the information they are looking for.

https://www.trustroots.org/volunteering ->
https://team.trustroots.org/ -> 
https://team.trustroots.org/Volunteering.html

I removed the intermediate step.


simison:
Thanks!

PR #2387, opened 2021-08-01
Up to date number of members based on actual stats

mariha:
Number of members on the front page was hardcoded to 59K and got outdated as we crossed 60K.

#### Proposed Changes

Number of members on the page is now read from the stats and rounded down to the nearest 5000.

#### Testing Instructions

I tried to kind of test it locally with different hardcoded values, haven't tested api errors. No automated tests either. And this is my first ever javascript code written in IJ without JS plugin so give it some special care.

Fixes https://trustroots.slack.com/archives/C017CNA5UQY/p1627680505000500

### DIFF
--- a/modules/pages/client/components/Home.component.js
+++ b/modules/pages/client/components/Home.component.js
@@ -80,23 +80,25 @@ export function getSignupUrl(circleSlug) {
   return '/signup';
 }
 
+function rounded(number, round = 5000) {
+  return parseInt(number / round) * round;
+}
+
 export default function Home({ user, isNativeMobileApp, photoCredits }) {
   const { t } = useTranslation('pages');
   // `tribe` route supported for legacy reasons, deprecated Feb 2021
   const { circle: circleRouteParam, tribe: tribeRouteParam } = getRouteParams();
   const circleRoute = circleRouteParam || tribeRouteParam;
 
-  const [stats, setStats] = useState([]);
-  useEffect(() => {
-    async function fetchData() {
-      const stats = await statisticsAPI.get({ limit: 3 });
-      setStats(stats);
-    }
-    fetchData();
+  const [statistics, setStats] = useState(false);
+  useEffect(async () => {
+    const { data } = await statisticsAPI.get();
+    setStats(data);
   }, []);
   // @TODO change this to be based on UI language rather than browser locale
+  const lastKnownMemberCount = 60000;
   const memberCount = new Intl.NumberFormat().format(
-    !stats ? 60000 : stats?.total,
+    rounded(statistics ? statistics?.total : lastKnownMemberCount),
   );
 
   // TODO get header height instead of magic number 56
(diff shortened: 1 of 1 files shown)


nonno:
It looks a great first time, congrats 👏  (I would just round to the closest, lower 1000 instead of 5000).

mariha:
Done. 
Actually though... why not display just exact number of members? it won't get stale so no need to round it...

nonno:
> Done.
> Actually though... why not display just exact number of members? it won't get stale so no need to round it...

I would agree with you, but I also know that for nerd people things work differently :) Probably from the communication side that "Trustroots is over (big, round number) members strong and growing!" sounds easier and more friendly.

mariha:
I was around but didn't do any coding in TR earlier, so don't have merge rights so I guess you have to merge?

nonno:
> I was around but didn't do any coding in TR earlier, so don't have merge rights so I guess you have to merge?

I'm "the last wagon wheel", no rights anywhere, and it's the first time I see a NodeJS PR. I guess someone soon or later will appear, considering also that you wrote on Slack too.

modules/pages/client/components/Home.component.js
```diff
@@ -79,14 +80,26 @@ export function getSignupUrl(circleSlug) {
...
+
 export default function Home({ user, isNativeMobileApp, photoCredits }) {
   const { t } = useTranslation('pages');
   // `tribe` route supported for legacy reasons, deprecated Feb 2021
   const { circle: circleRouteParam, tribe: tribeRouteParam } = getRouteParams();
   const circleRoute = circleRouteParam || tribeRouteParam;
 
+  const [statistics, setStats] = useState(false);
+  useEffect(async () => {
+    const { data } = await statisticsAPI.get();
```
simison, line 95: This API returns a whole lot of things, and does some ~10-15 database queries to get all that data:

https://github.com/Trustroots/trustroots/blob/0ef09face509894c22bbd79f98ea9b9fbe1a5eda/modules/statistics/server/controllers/statistics.server.controller.js#L244-L247

It's driving the public stats page https://www.trustroots.org/statistics

We can probably handle that for every logged out homepage load, but on the other hand it would be small effort to add new API route returning only user count. :-) Would you be up for it?

If yes, the route config is here: https://github.com/Trustroots/trustroots/blob/master/modules/statistics/server/routes/statistics.server.routes.js

It ties to a function that returns the content in controller: https://github.com/Trustroots/trustroots/blob/master/modules/statistics/server/controllers/statistics.server.controller.js

and there's a helper to just get the user count: 
https://github.com/Trustroots/trustroots/blob/0ef09face509894c22bbd79f98ea9b9fbe1a5eda/modules/statistics/server/controllers/statistics.server.controller.js#L22

You could add the rounding directly in the API if you'd like.

mariha: hey Mikael, good feedback, thanks! Will get back to it whenever I have a moment. 

I am actually really very glad you did the review, thanks!

simison: no problem, thanks for patience! :-)

mariha: I didn't have time to look at this this week and I am travelling next 2-3 weeks, so in the meantime opened #2394. 

From Sept I hope to have a bit more time 🤞🏼 for everything other than my kiddo, as the person who takes care of her will be back and we are moving forward with our child care/forest kindergarten plans. (child care is unimaginably neglected in Poland, mostly relies on grandparents who are not an option for us :/)

PR #2408, opened 2021-10-08
Add whatsapp number as button user setting option

chagai95:
I'm stuck on how add a user state

Proposed Changes
have an extra way of contacting a user - the user can choose in the settings if they want to activate this.
Testing Instructions
check out the Account section in edit profile and if you add externalContact field to the user in the db you should see this on their profile - have not tested this yet. 
Fixes #

The hard dependency on Email notifications which is annoying at the least and even broken sometimes.

### DIFF
--- a/modules/users/client/components/AddExternalContactInfo.component.js
+++ b/modules/users/client/components/AddExternalContactInfo.component.js
@@ -1,0 +1,58 @@
+import React, { useState } from 'react';
+import '@/config/client/i18n';
+import { useTranslation } from 'react-i18next';
+import 'react-phone-number-input/style.css';
+import PhoneInput from 'react-phone-number-input';
+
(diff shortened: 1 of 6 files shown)


karlkeefer:
I would think this raises trust/safety concerns. Displaying folks phone numbers on their profile creates a big opening for spam/abuse. I could imagine an option to have further details visible _once you agree to host someone_ that would mitigate most of the trust/safety issues, but then it doesn't solve the _alternative notification method_ problem you're hoping for.

Seems like fixing email notifications or implementing push notifications (if they don't already exist?) might be better?

chagai95:
Of course it's better, this is a short term solution but never the less very important to me. And the other stuff you talked about were all mentioned before and I think they are all great and nobel things to do, would love to do them with you 😉

PR #2421, opened 2021-10-13
Use place_name for US locations in location service

### LINKED ISSUE #2350: City names in the USA are not country-unique
**Describe the bug**
Portland, Oregon is the famous one, but Portland is also one of the biggest cities in the state of Maine. There are also many smaller towns called Portland.  

Screenshot of the location-selection form

If you picked a city or town at random, there's a ~75% chance that you can find another city in the U.S. with the same name.  This makes the location selection form impossible to use for most places in the US.  

**To Reproduce**
1. Go to `/profile/edit/locations`. 
2. Enter `Denver` in one of the fields. 
3. Try to select Denver, CO on the first try.

karlkeefer:
#### Proposed Changes

* use the mapbox place_name for US auto-complete suggestions

#### Testing Instructions

* Go to edit profile locations and type in the name of non-unique US city.

Fixes #2350

### DIFF
--- a/modules/core/client/services/location.client.service.js
+++ b/modules/core/client/services/location.client.service.js
@@ -208,6 +208,13 @@ function LocationService($log, $http, SettingsFactory) {
             title += ', ' + geolocation.context[i].text;
           } else if (geolocation.context[i].id.substring(0, 8) === 'country.') {
             title += ', ' + geolocation.context[i].text;
+            if (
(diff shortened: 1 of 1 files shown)


karlkeefer:
With this in place, I noticed it returned things like Portland International Airport... maybe the `tr-location` directive needs to be more restrictive where it limits the types of locations?

Issue #2430, opened 2021-10-15
Improvements to support contact form


simison:
**Goal:** improve the clarity of the support form and speed up the support team's process.

Note that any changes that pull information about people (avatar, name) cannot be done for the "logged out" version of the form for privacy reasons.

**Code**: 
- [Support module](https://github.com/Trustroots/trustroots/tree/5711b1da0b346a90f5f691e5d9c831ff40c36e6d/modules/support)
- [API for blocking users](https://github.com/Trustroots/trustroots/blob/0d00a492636eb06e309a38c1df3afb00f95c913d/modules/users/client/api/block.api.js)

## Visuals

The form is visible at https://www.trustroots.org/support?report=dummy

### Currently

**Form**

![image](https://user-images.githubusercontent.com/87168/137460878-449bfaf9-825d-4026-88c5-b99ec3bf4f93.png)

**Confirmation**

![image](https://user-images.githubusercontent.com/87168/137463286-41e10d39-e388-4d90-accf-ab5a1e221682.png)

**Logged out view**

![image](https://user-images.githubusercontent.com/87168/137479685-e9e60605-841e-4099-9f7d-cd01c56895ee.png)

### After

![image](https://user-images.githubusercontent.com/87168/137468533-290f6f21-bd50-44e9-850b-660e06b85c1f.png)

## Priority improvements

- [ ] Add checkbox to grant access to look into user's messages between them and whom they reported, checked by default:
  > "Allow support to see messages between the reported member and me." (Feel free to adjust copy)

    ![image](https://user-images.githubusercontent.com/87168/137466415-fffcbf13-500e-4366-be98-b99498213c87.png)

- [ ] Under _"See also"_ add _"Trustroots & safety"_ link once the page is done at `/safety`

- [ ] Under _"See also"_ add _"Blocking another member"_ link to `/faq` once the question is added in the FAQ.

## "Nice to have" improvements

Something that would make the experience better but isn't as important as the above things.

- [ ] Add ~10px spacing between the "see also" list

   ![image](https://user-images.githubusercontent.com/87168/137468244-9fdb7c5e-1ec6-434a-9928-5f8ecafa27a1.png)

- [ ] At the confirmation page, add "via email" to the copy for regular messages to:

    > "I’m just a small website robot but I’ve sent your message to our support people. Expect them to get back to you via email very soon!"

[...]

PR #2431, opened 2021-10-15
Improve block & report buttons visually

simison:
#### Proposed Changes

* Add new "flag" icon (using https://fontello.com/)
* Make block & report buttons more visible and harder to miss

Not included:

- Make buttons visible on message thread when on mobile; can be a separate PR

##### Before

Screenshot 2021-10-15 at 14 05 16

Screenshot 2021-10-15 at 14 04 56

image

##### After

Screenshot 2021-10-15 at 13 58 44

Screenshot 2021-10-15 at 13 58 12

Screenshot 2021-10-15 at 13 59 44

#### Testing Instructions

* Open message thread with someone (in desktop sized screen) and try buttons on the right.
* Open someone's profile (both desktop & mobile) and try buttons

Related to improving safety flows https://github.com/Trustroots/trustroots/issues/2430


karlkeefer:
This is definitely an improvement!

The screenshots from after make me wonder if we should have a flag/report _in the yellow alert box_ as well. It's really an attention-grabbing interface element, and the sidebar buttons just don't have much weight with a big orange thing on screen.

karlkeefer:
I was just thinking in the "you have blocked this member" warning, _if_ that is showing, it could also include a report this user button. It's _so_ attention-grabbing that I think someone that blocked a user might no longer see the "report" feature!

Definitely don't think it should be present all the time! :)

simison:
Ohh now I understood you right! 😅 My bad. Yeah, that's a good idea! Wanna pull request? 😉

PR #2454, opened 2021-11-04
profile/edit/about component

karlkeefer:
#### Proposed Changes

* migrate to a react component for the profile/edit/about screen
* use a custom hook to make change handlers simple for form fields

#### Testing Instructions

* go to your profile edit screen and make various changes

#### Status
So far it pulls in the existing data, you can edit the text fields and description field, and the save button actually persist the data.

#### TODOs, some of which need input from @simison 
- port confirm-on-exit
- port form validation
- create a date widget (or plug in an existing one)
- integrate with messageCenter somehow

PR #2456, opened 2021-11-07
Eslint: fail on use of console only in prod, warn on dev

simison:
#### Proposed Changes

* Fails Eslint only in CI / production when using `console.log`/`.warn`/`.error` etc without accompanying lint-ignore comment. Discussed in https://github.com/Trustroots/trustroots/pull/2454#discussion_r743189050

#### Testing Instructions

* Add console.log somewhere and it shouldn't fail your build locally when `npm start`
* Should fail when running `npm run start:prod`
* Try commit console.log and it should fail in PR in Github

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,8 @@ const reactVersion = semver.coerce(
   require('./package.json').dependencies.react,
 ).version;
 
+const isDevelopment = process.env.NODE_ENV === 'development';
+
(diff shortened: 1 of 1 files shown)


karlkeefer:
Can't test right now, but looks correct to me 👍👍

karlkeefer (approved):
Seems low risk to merge without testing for this change.

PR #2480, opened 2021-12-05
Admin: add ability to get circle members

simison:
#### Proposed Changes

* Add an admin tool to list circle member for email sending
* 
image

#### Testing Instructions

* /admin/newsletter


simison:
Something's funky with server tests but it isn't specific to this PR, tests well locally.

PR #2483, opened 2021-12-09
updated indentation

vasantisuthar:
#### Proposed Changes

* Updated indentation in file

#### Testing Instructions

*

Fixes #
fixed indentation in file


vasantisuthar (commented):
I have updated the indentation in the file, which helps in good readability for others.

simison:
Hi! Thanks for taking a look at Trustroots!

We use [Prettier](https://prettier.io/) to autoformat the code, so any manual changes would just get overwritten by Prettier changing it back. It's not always optimal like human would do, but having it automated and consistent is quite nice. :-)

You can try this locally to see it change back:

```
npm run reformat-files modules/users/client/views/authentication/signup.client.view.html
```

Issue #2486, opened 2021-12-11
Switch to ESM from CommonJS


simison:
We're using CommonJS `const foo = require('foo')` format in backend code, while the frontend already mostly uses modern ESM `import foo from 'foo'` statements.

Let's migrate for consistency!

### Todo

- [x] Migrate away from using `path` inside require, likely needs a [custom codemod](https://www.sitepoint.com/getting-started-with-codemods/). This change can be deployed independently from the rest.

    Consider example file in `modules/something/test.js` and how it now would load config:

    ```js
    const path = require('path');
    const config = require(path.resolve('./config/config'));
    ```
    Becomes:
    ```js
    const config = require('../../config/config');
    ```
    
    So: `path` isn't used anymore, and instead of the path from the root, we use a path relative to the file's current folder.

    We need to loop all the cases with a codemod (e.g. using [jscodeshift](https://github.com/facebook/jscodeshift)) and within the codemod resolve a new true path for the file:
    1. Take the original path inside the require (e.g. `./config`)    
    2. Calculate the required file's relative path to the current folder using [`path.relative`](https://nodejs.org/docs/latest/api/path.html#pathrelativefrom-to) function.

    Here's a demo that does something like that:
    ```js
    // Assume this file exists in modules/test.js
    // Run it from root folder `node modules/test.js`

    const path = require('path');

    const original = path.resolve('./config'); // Relative from the root
    const output = path.relative(__dirname, original); // __dirname is "current file's folder"; not sure if that works within     codemod but something similar should be available https://nodejs.org/en/knowledge/getting-started/globals-in-node-js/

    console.log(output); // Should give `../config`
    ```

    You might need to tidy up codemod's output with [Prettier](https://prettier.io/). You can use the command `npm run reformat-files .` to run Prettier on all files.

- [ ] Review if any remaining [path imports](https://github.com/Trustroots/trustroots/pull/2507#issuecomment-1021143262) need cleaning out

- [ ] Migrate to ESM import/exports:
    Add `"type": "module"` to `package.json`

    Use codemod to migrate to ESM:
    Install [5-to-6 codemods](https://github.com/5to6/5to6-codemod) (there might be others too but I'm familiar with this one)
    ```
    npm install 5to6-codemod --no-save
    ```

    Run this for each folder on TR root:
    ```
    npx jscodeshift --extension=js -t node_modules/5to6-codemod/transforms/exports.js [FOLDER NAME]
[...]

PR #2492, opened 2021-12-12
No more 'path' inside require

mariha:
A step towards #2486


simison:
@mariha Yay! I see this is in progress so hope you don't mind some advice. :-)

- No need `file:///` at the beginning of paths
- You need to loop all the cases with a codemod (e.g. using [jscodeshift](https://github.com/facebook/jscodeshift)) and within the codemod resolve a new true path for the file:
   - Take the original path inside the require (e.g. `./config`)
   - Calculate the required file's relative path to the current folder using [`path.relative`](https://nodejs.org/docs/latest/api/path.html#pathrelativefrom-to) function.

Here's a demo that does something like that:
```js
// Assume this file exists in modules/test.js
// Run it from root folder `node modules/test.js`

const path = require('path');

const original = path.resolve('./config'); // Relative from the root
const output = path.relative(__dirname, original); // __dirname is "current file's folder"; not sure if that works within codemod but something similar should be available https://nodejs.org/en/knowledge/getting-started/globals-in-node-js/

console.log(output); // Should give `../config`
```

Lemme know if I can help otherwise!

mariha:
(it may not seem like 😊, but I am working on it. I'll probably just abandon this PR and start a new one once there is something to share)

Issue #98, opened 2014-12-25
Missing languages


simison:
Language list is pretty well covering all common languages, but something might still be missing:

Add:
- [ ] Dutch (Flemish)
- [x] Esperanto
- [x] Latin
- [x] Ancient Greek
- [x] Creole languages
- [x] Sign languages
- [x] Serbo-Croatian
- [ ] Na'vi
- [ ] Aluku
- [ ] Wompanoag
- ?

simison:
This one is a bit odd (for "Greek"):
>Modern Greek (1453-)

simison:
>It's missing some languages that I speak: Aluku, one bushinengee language important here in French Guiana. As I'm learning and teaching for aluku kids, I would like to show it to my possible hosts, it's something that I offer in exchange.
>Also, French Creole and English Creole. It could be any creole tongue from any of the old colonies.

simison:
>I wanted to add Esperanto to list of my languages, but did not find it in the selection.

simison:
>There are no sign languages in languages to put in :-( Please add them whenever you can :-)

simison:
@RichardLitt #290:
>I am unable to add Latin, Ancient Greek, or Na'vi to my languages. Granted, I am unlikely to speak those languages to a guest - but it's not impossible, and I would like to be able to add that I speak them. I also cannot add Wompanoag - a minority language, chosen at random and which I don't speak, with a few hundred speakers in Massachusetts.

>Adding the ability to add your own language by pressing enter would alleviate this issue.

simison:
>the language spoken in Slovenia is "Slovene", while "Slovenian" means a person living in Slovenia.

simison:
Got:
>Hi please add punjabi as one of language in language selection box. Thank you.

simison:
Missing "Serbo-Croatian":

>I've noticed Serbian, Bosnian and Croatian are separate languages even though they are actually one language, Serbo-Croatian. So I suggest that you can copy CS on that one, offer BCMS languages separate (including Montenegrin which is not listed there) and also offer Serbo-Croatian, so that people can choose whatever they feel most comfortable with. And I feel like Serbo-Croatian is the best option for me personally because then I don't have to type that I speak four different languages and appear to be some super-multilingual person.
>
>Of course, if there are some reasons why you decided to skip Montenegrin and not offer Serbo-Croatian, that's cool too, I just wanted to offer my feedback and this link:
>
[...]

PR #2494, opened 2021-12-12
Update locale handling and selection UI

simison:
#### Proposed Changes

* Pull way more languages into spoken languages list
* Swap jquery based Angular select input to React version https://react-select.com/
* Generate Array version alongside Object version as a file, so that we don't have to generate it on the fly at the server

#### Testing Instructions

* Create user, modify their profile. Try adding and removing languages, saving in between.
   * You should find much more languages than before
   * List works better on mobile than previously
   * You should be able to save / remove languages just like before
 * Add hosting offer with the user
 * Create another user
   * Go to map and apply language search filter, use same and different languages as you saved for previous profile
   * You should see the previous user with right locale filter
   * The input should work in identical way to profile editing

A good small follow-up will be to apply more TR colored [theme](https://react-select.com/styles#overriding-the-theme) to this input but for now I just wanted to move this forward.

Resolves https://github.com/Trustroots/trustroots/issues/98


modules/users/client/views/profile/profile-edit-about.client.view.html
```diff
@@ -212,19 +212,11 @@
...
-              tr-languages="profileEditAbout.user.languages"
-              ng-change="profileEdit.unsavedModifications = true"
-              aria-labelledby="label-languages"
-            ></div>
-            <p class="help-block">Add languages you speak.</p>
-          </div>
-        </div>
+        <profile-edit-languages
+          onChangeLanguages="profileEditAbout.onChangeLanguages"
+          ng-change="profileEdit.unsavedModifications = true"
```
simison, line 217: TODO: remove

modules/search/client/controllers/search.client.controller.js
```diff
@@ -133,15 +133,11 @@ function SearchController(
...
-    // `vm.filters.languages` is still out of sync at this point,
-    // but in next cycle after `$timeout` we have updated version.
-    $timeout(function () {
-      // Save new value to cache
-      FiltersService.set('languages', vm.filters.languages || []);
-
-      onFiltersUpdated();
-    });
+  function onLanguageFiltersChange(languages) {
+    console.log('onLanguageFiltersChange:', languages); //eslint-disable-line
```
simison, line 137: TODO: remove `console.log`

package.json
```diff
@@ -176,7 +176,9 @@
     "react-i18next": "11.13.0",
     "react-map-gl": "5.3.16",
     "react-medium-editor": "1.8.1",
+    "react-query": "3.18.1",
     "react-responsive": "8.2.0",
+    "react-select": "4.3.1",
```
simison, line 181: TODO: update to v5

config/languages/README.md
```diff
@@ -1,10 +1,15 @@
...
+```bash
+cd config/languages
+node generate.js
+```
 
 ### Todo
 
 - See [Missing languages](https://github.com/Trustroots/trustroots/issues/98)
-- If the search input for languages can one day handle 7000+ items, we can include all the living languages and dialects. Until that we're limiting this to ~500 or so.
+- If the search input for languages can one day handle 7000+ items performance wise, we can include all the living languages and dialects. Until that we're limiting this to ~500 or so.
```
simison, line 15: The new input in this PR should be capable enough, as long as we can async-load the list instead of including it out on every page load as we do now.

config/languages/generate.js
```diff
@@ -37,47 +34,47 @@ function fixName(name) {
...
     return true;
   }
 
   // Include all "living" languages
   // Include 19 constructed languages such as "Esperanto"
   // Ignores all other types (e.g. extinct)
-  if (['constructed', 'living'].indexOf(language.type) === -1) {
-    return false;
+  if (['constructed', 'living'].includes(type)) {
+    return true;
```
simison, line 53: Note; changed to return _every_ living language.

modules/core/server/controllers/core.server.controller.js
```diff
@@ -94,3 +94,27 @@ exports.renderServiceWorkerConfig = function (req, res) {
...
+
+  return langsArr;
+}
+
+exports.getLanguages = (req, res) => {
+  let languages = require(path.resolve('config/languages/languages.json'));
+
+  // Reformat object into array if so requested
+  if (req?.query?.format === 'array') {
+    languages = languageObjectToArray(languages);
```
simison, line 116: Todo; instead of doing this on the fly with the server, how about we do this at compile-time, store the file at the server and simply read it?

simison:
Some UI work needed to remove odd scrolling:

Screenshot 2021-12-23 at 17 20 59

simison:
We'll need to either revert or fix this since it caused a bug in messages, described in https://github.com/Trustroots/trustroots/issues/2511

Can't revert from GitHub UI due to:

> Sorry, this pull request couldn’t be reverted automatically. It may have already been reverted, or the content may have changed since it was merged.

Needs  some manual work. :-)

PR #2499, opened 2021-12-21
Update locale names

simison:
#### Proposed Changes

* Updates capitalization of some names
* Update Brazil->Brasil
* Update https://en.wikipedia.org/wiki/Czech_language

### DIFF
--- a/config/shared/locales.json
+++ b/config/shared/locales.json
@@ -1,5 +1,5 @@
 [
-  { "label": "česky", "code": "cs", "english": "Czech", "production": false },
+  { "label": "Čeština", "code": "cs", "english": "Czech", "production": false },
   { "label": "Deutsch", "code": "de", "english": "German", "production": true },
   {
     "label": "English",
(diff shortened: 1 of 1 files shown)


config/shared/locales.json
```diff
@@ -1,5 +1,5 @@
 [
-  { "label": "česky", "code": "cs", "english": "Czech", "production": false },
+  { "label": "Čeština", "code": "cs", "english": "Czech", "production": false },
```
simison, line 2: Both work but Wikipedia uses Čeština so I went with it. https://en.wikipedia.org/wiki/Czech_language

config/shared/locales.json
```diff
@@ -136,31 +136,31 @@
...
   {
-    "label": "latviešu",
+    "label": "Latviešu",
     "code": "lv",
     "english": "Latvian",
     "production": false
   },
   {
-    "label": "српски / srpski",
+    "label": "Srpski (latinica)",
```
simison, line 151: We can add српски separately if someone wants to translate it.

config/shared/locales.json
```diff
@@ -37,22 +37,22 @@
...
   {
-    "label": "português",
+    "label": "Português",
     "code": "pt",
     "english": "Portuguese",
     "production": true
   },
   {
-    "label": "português (Brazil)",
+    "label": "Português (Brasil)",
```
simison, line 48: Also Brazil → Brasil typo.

PR #2507, opened 2022-01-02
Move away from require with path.resolve().

mariha:
The jscodeshift code used for the transformation:

```js
const path = require('path');

export default function transformer(file, api) {
  const j = api.jscodeshift;

  return j(file.source)
    .find(j.CallExpression, {callee: {type: "Identifier", name: "require"}})
    .filter(callExpr => {  // has path.resolve() as an argument
      return callExpr.value.arguments[0].type === "CallExpression" // call expression (not a literal)
        && callExpr.value.arguments[0].callee.property.name === "resolve"
        && callExpr.value.arguments[0].arguments[0].type === "Literal";
    })
    .forEach(callExpr => {
      const originalPath = callExpr.value.arguments[0].arguments[0].value;
      const absolutePath = path.resolve(originalPath);
      const relativePath = path.relative(path.dirname(file.path), absolutePath);

      // files from the same directory resulted in 'Error: Cannot find module' if './' was not added
      callExpr.value.arguments[0] = j.literal(relativePath.startsWith('.') ? relativePath : "./" + relativePath);
    })
    .toSource({quote: 'single'});
}
```

A step towards #2486

### DIFF
--- a/bin/admin/trshell.js
+++ b/bin/admin/trshell.js
@@ -3,7 +3,6 @@ process.chdir(__dirname);
 process.chdir('../../');
 
 var mongoose = require('mongoose');
-var path = require('path');
 var mongooseService = require('../../config/lib/mongoose');
 
 mongooseService.connect();
--- a/bin/admin/trshell.js
+++ b/bin/admin/trshell.js
@@ -4,7 +4,7 @@ process.chdir('../../');
 
 var mongoose = require('mongoose');
 var path = require('path');
-var mongooseService = require(path.resolve('config/lib/mongoose'));
+var mongooseService = require('../../config/lib/mongoose');
 
 mongooseService.connect();
 mongooseService.loadModels();
--- a/bin/db-maintenance/archive-done-agenda-jobs.js
+++ b/bin/db-maintenance/archive-done-agenda-jobs.js
@@ -16,7 +16,6 @@
  */
 
 const { MongoClient } = require('mongodb');
-const path = require('path');
 const chalk = require('chalk');
 const async = require('async');
 const config = require('../../config/config');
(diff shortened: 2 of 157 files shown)


simison (approved):
Great job with this! I've smoke tested different things, including email notifications, and things are working smoothly.

There will likely be a couple of cases we need to sort out manually when moving to `imports` but that can be done in separate PR. (I searched for `require(path.resolve` and see 6 cases).

Linter check in this PR is failing because of a fork but ran it locally (`npm run lint`) and it passed.

mariha:
> Linter check in this PR is failing because of a fork but ran it locally (npm run lint) and it passed.
For whatever reason I can not push changes/create branch directly to TR repo.

mariha:
There are a few more `require('path')`, haven't looked into them.

```zsh
➜  trustroots grep -rn "require('path')" . 
./bin/generate-circle-images.js:15:const path = require('path');
./config/config.js:8:const path = require('path');
./config/webpack/templateloader.js:1:const { join, resolve } = require('path');
./config/webpack/webpack.config.js:17:const { join, resolve } = require('path');
./config/webpack/webpack.shims.js:20:const { join } = require('path');
./config/webpack/webpack.push-messaging-sw.config.js:1:const { join } = require('path');
./config/lib/express.js:21:const path = require('path');
./config/lib/mongoose.js:6:const path = require('path');
./scripts/exportEmailsAsEml.js:14:var path = require('path'),
./modules/core/tests/server/jobs/send-push-message.server.job.tests.js:3:const path = require('path');
./modules/core/server/services/email.server.service.js:5:const path = require('path');
./modules/users/tests/server/user-removal.server.routes.tests.js:6:const path = require('path');
./modules/users/server/config/users.config.server.js:6:const path = require('path');
./modules/users/server/controllers/users.profile.server.controller.js:5:const path = require('path');
./modules/users/server/controllers/users.avatar.server.controller.js:6:const path = require('path');
```

simison:
I sent an invite to join the Trustroots repo, you should have it in email and somewhere in github interface, maybe at https://github.com/Trustroots/

simison:
Yep, those might also be valid reasons to use `path`, so they will just need to be converted into `import`s.

PR #2508, opened 2022-01-05
Add an option to link to Couchers.org on profile

aapeliv:
#### Proposed Changes

* As discussed on Slack, adds an option to link to Couchers.org on profile

Unfortunately I tried a few different ways of running the app locally, but I failed to get it running... So this is a bit of an "armchair" contribution... I'd appreciate someone checking over it or testing it locally?


modules/users/client/components/ProfileViewBasics.js
```diff
@@ -133,6 +133,20 @@ export default function ProfileViewBasics({ profile }) {
               )
             );
           })}
+        {/* Couchers.org */}
+        {profile.extSitesCouchers && (
+          <li className="social-profile">
+            <i className="social-profile-icon icon-fw icon-lg icon-bw"></i>
```
simison, line 139: This icon will need updating:

Screenshot 2022-01-05 at 20 31 53

aapeliv: Ah good point, thanks

simison:
Overall testing well! Just icon will need updating.

image

Screenshot 2022-01-05 at 20 32 14

modules/users/client/components/ProfileViewBasics.js
aapeliv: Fixed

aapeliv:
@simison: I think I fixed the review comment. But I'm not sure how to fix the linter error in CI?

simison:
Linter fails for forks currently (https://github.com/Trustroots/trustroots/pull/2491 might fix it) but you can run it locally with:
```
npm ci
npm run lint
```

( I can also check before merging)

aapeliv:
I can't run `npm ci` locally, that fails for me.

simison (approved):
Works well now, thank you!

image

Linter is happy locally, too.

simison:
BTW side note that CI translation commit in this PR was from Github action I added a while back (https://github.com/Trustroots/trustroots/pull/2438) but looks like I need to adjust how it works with forks.

Issue #2513, opened 2022-01-12
Add filter option: can accommodate multiple guests


TMC89:
Through support:

**It would be great if in the filters exited an option such as "hosts that can accomodate more than 1 person", it would be much easier to find someone who can host more people.**

PR #2546, opened 2022-03-27
Add filter option: can accommodate multiple guests #2513 

vssdeo:
Through support:

It would be great if in the filters exited an option such as "hosts that can accommodate more than 1 person", it would be much easier to find someone who can host more people.

[bug 2513]

- on /search below "Hosts or meetups ?" have added a inc and dec buttons
- Text is "How many persons can the host accommodate?"
- based on the selected value the search is modified for maxGuests
- html filedset is taken from existing code

#### Testing Instructions

/search page from side bar select / increment / decrement the number of people which the host can accept.

### DIFF
--- a/modules/offers/server/controllers/offers.server.controller.js
+++ b/modules/offers/server/controllers/offers.server.controller.js
@@ -608,6 +608,21 @@ exports.list = function (req, res) {
     }
   }
 
+  // offer maxGuests filter
+  if (filters.hasObjectFilter('offer')) {
+    // Accept only valid values, ignore the rest
+    // @link https://lodash.com/docs/#filter
+    const maxGuestsReq = filters.offer.maxGuests;
+
+    query.push({
+      $match: {
+        maxGuests: {
+          $gte: maxGuestsReq,
+        },
+      },
+    });
+  }
+
   // Filter out users that do not share any circles with the authenticated user
   // and chose to not appear in those searches.
   const showOnlyInMyCirclesQueries = [{ showOnlyInMyCircles: false }];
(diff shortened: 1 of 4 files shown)


modules/offers/server/controllers/offers.server.controller.js
```diff
@@ -608,6 +608,21 @@ exports.list = function (req, res) {
     }
   }
 
+  // offer maxGuests filter
+  if (filters.hasObjectFilter('offer')) {
```
simison, line 612: Should we avoid running the query if value is 1? Would speed things up in default cases.

modules/offers/server/controllers/offers.server.controller.js
```diff
@@ -608,6 +608,21 @@ exports.list = function (req, res) {
...
+    // Accept only valid values, ignore the rest
+    // @link https://lodash.com/docs/#filter
```
simison, line 614: Validation was missing here but no need to use lodash, can just ensure with `parseInt(value, 10)` and maybe `Math.min(10, value)` to avoid larger than 10?

modules/search/client/views/search-sidebar-filters.client.view.html
```diff
@@ -6,6 +6,52 @@ <h4>Hosts or meetups?</h4>
   </div>
   <!-- /Offer type filter -->
 
+  <!-- Number of People offer maxGuests filter -->
+  <fieldset class="offer-maxguests">
+    <legend>
+      <h4 id="maxGuests">How many persons can the host accommodate?</h4>
```
simison, line 12: I wonder if there's shorter ways to say the same? This works fine too tho.

vssdeo: 1. How many persons can the host accommodate ?
2. Persons host can accommodate ?
3. Number of persons host can c ?
4. Max number of guests allowed ?
5. Max guests host can accommodate ?
6. Max number of guests allowed ?

PR #2558, opened 2022-04-05
2522 add more info about a member on the map

vssdeo:
#### Proposed Changes

*

#### Testing Instructions

*

Fixes #

### DIFF
--- a/modules/search/client/controllers/search.client.controller.js
+++ b/modules/search/client/controllers/search.client.controller.js
@@ -41,6 +41,11 @@ function SearchController(
   // Visibility toggle for search place input on small screens
   vm.isPlaceSearchVisible = false;
 
+  if (offer && offer.user) {
+    const momentDate = moment(offer.user.seen);
+    vm.offer.user.lastSeen = momentDate.fromNow();
+  }
+
   // Init search from the URL, `tr-location` directive attached
   // to search input will take care of the rest.
   // `Replacing underscore with space is to make search queries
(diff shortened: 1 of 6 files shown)


modules/search/client/controllers/search.client.controller.js
```diff
@@ -41,6 +41,11 @@ function SearchController(
   // Visibility toggle for search place input on small screens
   vm.isPlaceSearchVisible = false;
 
+  if (offer && offer.user) {
```
simison, line 44: You can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in case like this:

`if (offer?.user?.seen)`

modules/search/client/views/search-sidebar-results.client.view.html
```diff
@@ -97,6 +97,8 @@ <h4>
                   'btn-offer-hosting-maybe': search.offer.status === 'maybe'
                 }"
           ></span>
+          <div>Online {{search.offer.user.lastSeen}}</div>
```
simison, line 100: Could we put this in some shared place and re-use?

https://github.com/Trustroots/trustroots/blob/061847721413977c17c9623b23c68ed5351a8a0b/modules/users/client/components/ProfileViewBasics.js#L31-L36

Lemme know if I can help to show how to use React components in Angular templates.

modules/search/client/views/search-sidebar-results.client.view.html
```diff
@@ -97,6 +97,8 @@ <h4>
...
+          <div>Reply Rate: {{search.offer.user.replyRate}}</div>
```
simison, line 101: Same as above about re-use: https://github.com/Trustroots/trustroots/blob/061847721413977c17c9623b23c68ed5351a8a0b/modules/users/client/components/ProfileViewBasics.js#L41-L51

PR #2640, opened 2023-04-25
Reuse mapbox-gl Map to decrease costs

mrkvon:
#### Proposed Changes

Add `reuseMaps` parameter to react-map-gl maps. It may decrease mapbox-gl `Map` loads and [thus decrease costs](https://docs.mapbox.com/mapbox-gl-js/guides/pricing/?utm_medium=pricing).
https://visgl.github.io/react-map-gl/docs/get-started/tips-and-tricks#minimize-cost-from-frequent-re-mounting

#### Testing Instructions

* Check that the app still works and loads maps. Whether this improves the cost issue, we'll have to wait and see.

### DIFF
--- a/modules/core/client/components/Map/index.js
+++ b/modules/core/client/components/Map/index.js
@@ -32,6 +32,7 @@ export default function Map(props) {
 
   return (
     <ReactMapGL
+      reuseMaps
       dragRotate={false}
       height={320}
       mapboxApiAccessToken={MAPBOX_TOKEN}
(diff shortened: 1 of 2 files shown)


chagai95:
LGTM, nice catch!

chagai95 (approved):

mariha:
For high bills issue, when I shortly volunteered with WelcomeToMyGarden.org project, I remeber someone wrote to MapBox support with a question if WTMG could get a discount as a not-for-profit / non-profit organization, and they got 90% discount, afair. Might be worth exploring similar option for TR.

PR #2647, opened 2023-09-30
change some links to https from http

chagai95:
#### Proposed Changes

* change some links to https from http because http does not work well in embedded browsers

#### Testing Instructions

*

Fixes #

### DIFF
--- a/deploy/files/prod-public-assets/maintenance.html
+++ b/deploy/files/prod-public-assets/maintenance.html
@@ -60,7 +60,7 @@
 	<br>
 	<p>Follow us on <a href="https://twitter.com/trustroots">Twitter</a> and <a href="https://www.facebook.com/trustroots.org">Facebook</a></p>
 	<br>
-	<p>Check our <a href="http://status.trustroots.org/">status page</a></p>
+	<p>Check our <a href="https://status.trustroots.org/">status page</a></p>
     </div>
 </article>
 
--- a/modules/messages/tests/server/message.server.routes.tests.js
+++ b/modules/messages/tests/server/message.server.routes.tests.js
@@ -344,7 +344,7 @@ describe('Message CRUD tests', function () {
         // Create html in message
         const htmlMessage = message;
         htmlMessage.content =
-          '<strong>strong</strong><br><img src="http://www.trustroots.org/">' +
+          '<strong>strong</strong><br><img src="https://www.trustroots.org/">' +
           '<foo>blockquote</foo><p>' +
           '<script></script>' +
           '<a href="https://www.trustroots.org/">link</a>' +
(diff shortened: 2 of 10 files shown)


modules/pages/client/components/Foundation.component.js
```diff
@@ -124,7 +124,7 @@ export default function Foundation({ user }) {
                   <Trans t={t} ns="pages">
                     Working on big freegan/travel projects such as{' '}
                     <a href="https://hitchwiki.org/">Hitchwiki</a>,{' '}
-                    <a href="http://hitchgathering.org/">
+                    <a href="https://hitchgathering.org/">
```
mrkvon, line 127: The https version hangs for me. Only the http version seems to redirect to a meaningful page

deploy/files/prod-public-assets/maintenance.html
```diff
@@ -60,7 +60,7 @@ <h1>We&rsquo;ll be back soon!</h1>
 	<br>
 	<p>Follow us on <a href="https://twitter.com/trustroots">Twitter</a> and <a href="https://www.facebook.com/trustroots.org">Facebook</a></p>
 	<br>
-	<p>Check our <a href="http://status.trustroots.org/">status page</a></p>
+	<p>Check our <a href="https://status.trustroots.org/">status page</a></p>
```
mrkvon, line 63: This one doesn't seem to function properly. In any case https is better than http. Review and fixing of links can be done in separate PR.

modules/pages/client/components/FaqTechnology.component.js
```diff
@@ -42,7 +42,7 @@ export default function FaqTechnology() {
           limited software you can imagine that still works for hospitality. We
           simply haven&apos;t had time to build most of the features yet. Please
           be patient and be mindful about{' '}
-          <a href="http://www.productstrategymeanssayingno.com/">
+          <a href="https://www.productstrategymeanssayingno.com/">
```
mrkvon, line 45: The https version hangs for me at the moment. Only the http link redirects to a meaningful website atm.

mrkvon (changes requested):
Great step improving users' privacy!

Linter failed because of prettier. You can fix that with

```bash
npm run lint -- --fix
```
Some links only show meaningful content when using http. So it would be good either to keep the http, or (in case of redirect) to use the redirected link. I added notes when i noticed this.

Some links are dead or dysfunctional. I added notes where i noticed it, but it's out of scope of this PR to fix those. Feel free to ignore those comments, they're just for future reference or something... 😉 

I'll be happy to approve when linter passes and when the links which only work with http get reverted or otherwise fixed... 👍🏾 

_edit: i also added suggestion how to fix the failing test_

modules/messages/tests/server/message.server.routes.tests.js
```diff
@@ -380,7 +380,7 @@ describe('Message CRUD tests', function () {
                     '<br />blockquote' +
                     '<p>' +
                     '<a href="https://www.trustroots.org/">link</a>' +
-                    '<a href="http://www.trustroots.org">www.trustroots.org</a>' +
+                    '<a href="https://www.trustroots.org">www.trustroots.org</a>' +
```
mrkvon, line 383: This test fails: the original link `www.trustroots.org` actually formats to `<a href="http://www.trustroots.org">www.trustroots.org</a>`

(to http)

So this change should be reverted, in order for tests to pass again.

PR #2649, opened 2023-10-16
Increase member count to 100k

mrkvon:
#### Proposed Changes

* Update the member count on main page from 80k to 100k

#### Testing Instructions

* It's just a change of 2 characters. If you really want to test, run the app, and scroll a bit down on the main page, and see that number says 100k. This change shouldn't affect translations, since the number is provided dynamically.


chagai95 (approved):
LGTM🎉

PR #2651, opened 2023-10-23
Upgrade Node version in extract-i18n github workflow

mrkvon:
#### Proposed Changes

* Upgrade Node to v16 in extract-i18n github workflow

One of the dependencies, listr2, [expects Node v16](https://github.com/listr2/listr2/blob/a55543d886afc88573bf67a37ae8a91d4b0deafa/package.json#L15), otherwise fails

#### Testing Instructions

* we'll see whether the extraction of translation strings passes after merge (i suspect that there will be permissions error, but that's another issue)

Fixes https://github.com/Trustroots/trustroots/actions/runs/6618933992/job/17978426609#step:8:329

### DIFF
--- a/.github/workflows/extract-i18n.yml
+++ b/.github/workflows/extract-i18n.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 14
+          node-version: 16
       - name: Cache Node.js modules
         uses: actions/cache@v2
         with:


mrkvon:
Tests are failing due to some unrelated issue with graphicsmagick installation, maybe will fix later.

mrkvon:
Would anybody consider approving this? 🙂

chagai95 (approved):

chagai95:
Sorry I didn't do this earlier, thx 😊

mrkvon:
Thank you @chagai95! 😊

The update of translations still fails because of [lack of permissions to push to protected branch](https://github.com/Trustroots/trustroots/actions/runs/7147095013/job/19466006999#step:8:331), but this PR fixed what it was supposed to fix 🎉 

Maybe we could tweak the workflow script to open a PR instead of direct push... then merge manually.

PR #2658, opened 2024-01-17
Extract translation strings

mrkvon:
The translation string extraction has been failing because of branch protection rules. So we have to do it manually until the issue is fixed.

Just run the following:

```bash
npm run i18n:extract
npm run i18n:fix-weblate
```

#### Proposed Changes

This just updates the translation strings

#### Testing

Nothing to test, really. :)


chagai95 (approved):
Not sure I understood everything but it looks like just adding strings and other minor changes, thx!

mrkvon:
> Not sure I understood everything but it looks like just adding strings and other minor changes, thx!

Indeed! Just running those two scripts, really. Thank you for the fast merge! ❤️ 🚀

PR #2678, opened 2024-02-20
rm .github/workflows/*

### LINKED ISSUE #2677: clean up .github/workflows
https://github.com/Trustroots/trustroots/tree/master/.github/workflows

Both lint.yml and test.yml don't seem to work as intended. Guess we can just remove them?

guaka:
#### Proposed Changes

*

#### Testing Instructions

*

Fixes #2677


chmac:
As per #2677 I don't think this is the right approach. We need to fix the dependency issue.

PR #2676, opened 2024-02-20
Update Feedback.js

guaka:
"Leave your public feedback here. Remember to be respectful of others and restrain from any kind of abusive behaviour."

#2539

#### Proposed Changes

*

#### Testing Instructions

*

Fixes #

PR #2681, opened 2024-02-20
Force git to use https in workflows

### LINKED ISSUE #2679: GitHub workflows are broken
See #2676 for context.

Some references:

- https://github.com/actions/setup-node/issues/214
- https://github.com/npm/cli/issues/2610

Essentially, the `git+https://github.com` URL in `package.json` is getting converted by npm of its own accord into a `git+ssh` url, which in turn fails. There are a few potential solutions:

- Add a username and password to the `git+https://` url
  - Apparently this stops it getting converted to an ssh url
  - Unclear if there is any username / password which will work
- Add a step in the workflow to reconfigure git to always use https

chmac:
fixes #2679

Again, not sure if this will work.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,6 +28,10 @@ jobs:
           restore-keys: |
             ${{ runner.OS }}-node-
             ${{ runner.OS }}-
+      - name: Reconfigure git to use HTTP authentication
+        run: >
+          git config --global url."https://github.com/".insteadOf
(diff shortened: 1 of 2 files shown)


chmac:
Worked. Should finish test run shortly, then I will merge and update #2676 and then merge that also.

chmac:
Tests failed, but I think they were always failing, they seem to be broken. :-(

PR #2725, opened 2026-04-20
implement shadowban ghost-send and suppress message-related delivery

guaka:
Implement consistent shadowban behavior for messaging: shadowbanned users can continue to send and read their own messages, while delivery remains hidden from recipients. Also suppress message-related email notifications for shadowbanned/suspended participants (unread reminders and contact-confirm emails), and update tests to cover the new behavior.

Made-with: Cursor

#### Proposed Changes

- Add ghost-send behavior for restricted senders in messaging:
  - save messages as sender-visible but recipient-hidden (`shadowHidden`)
  - allow sender thread/sync access while filtering hidden incoming messages from recipient thread/sync
  - avoid recipient thread delivery updates for hidden messages
- Suppress message-related emails when either participant is `shadowban`/`suspended`:
  - unread reminder emails
  - contact-confirm emails
- Expand server tests for:
  - unread reminder suppression
  - contact-confirm suppression
  - shadowban messaging visibility/delivery behavior

#### Testing Instructions

- Run:
  - `NODE_ENV=test ./node_modules/.bin/mocha "modules/core/tests/server/services/email.server.service.tests.js" --exit`
  - `NODE_ENV=test node -r should -e "<bootstrap>"` for `modules/messages/tests/server/jobs/message-unread.server.job.tests.js`
- Verify shadowban behavior manually:
  - shadowbanned user can send and read own thread/sync
  - recipient does not see those messages in thread/sync
  - no unread/contact email is generated for shadowbanned/suspended participants

### DIFF
--- a/modules/contacts/server/controllers/contacts.server.controller.js
+++ b/modules/contacts/server/controllers/contacts.server.controller.js
@@ -90,7 +90,7 @@ exports.add = function (req, res) {
 
       // Find friend
       function (messageHTML, messagePlain, done) {
-        User.findById(req.body.friendUserId, 'email displayName').exec(
+        User.findById(req.body.friendUserId, 'email displayName roles').exec(
           function (err, friend) {
             if (!friend)
               return done(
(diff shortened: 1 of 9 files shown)


chmac (commented):
I think the helper should be in a single file and reused, it doesn't make sense to duplicate the same code everywhere, but that's not a reason to block deployment. I haven't checked it in detail, but reading it on my phone it looks okay.

Issue #2723, opened 2026-03-21
better error message for signups with username with _


guaka:
just found this out: when user tries to sign up with e.g. `username_number` they get "Snap! Something went wrong. If this keeps happening, please contact us."

this likely relates to many other support messages related to failed signups

(but maintenance mode)

PR #2726, opened 2026-04-28
Broader account validation policy and follow-ups

guaka:
Deferred follow-up to #2820, which extracts the signup error messages, client validation feedback and submission guards without changing account policy. This draft is based on `codex/signup-validation-messages` so its diff shows the remaining work. Merge #2820 first; this draft must remain unmerged pending a separate policy review.

Remaining changes:
- Restrict new and changed usernames to 3–34 lowercase letters and digits with at least one letter, and expand reserved names.
- Move username policy enforcement to routes while retaining legacy username lookups, NIP-05 compatibility and profile help text.
- Add name-format restrictions and stronger temporary-email validation.
- Adjust disabled welcome-email jobs and tests, plus documentation formatting.
- Document the proposed policy in OpenSpec and test its legacy-identity behaviour.

Outstanding review items:
- Reject non-string username updates before truthiness checks: numeric `0` and boolean `false` can bypass validation and the username-change restriction with the relaxed model.
- Preserve unchanged existing names when saving unrelated account fields; the name restrictions currently reject some previously accepted names.
- Restore the role-injection regression using the valid privileged `admin` role.
- Decide which policy changes are wanted, and separate unrelated email/job maintenance and formatting.

The narrower fix for #2723 is now in #2820. This draft does not block that fix.


guaka:
relates to #2723

guaka:
let's try this simpler way first: https://github.com/Trustroots/trustroots/commit/d8fc934ba31e33c1ca51d455f17e7e2d6d490f12

chmac:
Definitely seems like an easier option!

guaka:
## Review follow-ups implemented

Addressed the items from the critical review. **PR description updated** to match actual scope.

### Code changes (commit `75d2b07ae` on branch — needs push from local)

**1. Route-only policy documented (`user.server.model.js`)**
Mongoose `validateUsername` comment now states explicitly that format/reserved-name policy is enforced only at signup, signup-validation, and profile username-change routes — not at the schema layer.

**2. Legacy NIP-05 lookup documented (`authentication.server.service.js`)**
`isLegacyUsernameLookupValid` documents why it uses a separate regex (pre-policy charset, no letter requirement) and warns not to merge with `isUsernameFormatValid`.

**3. Async validator shows API messages (`tr-validate-username`)**
Server `message` from `/api/auth/signup/validate` is stored on `ngModel.$usernameValidationMessage` and surfaced by `getUsernameValidationError` in signup and profile-edit-account. Reserved names like `nostr` now show “Username is not available.” instead of the generic “already in use” copy.

**4. Test `illegalStrings` documented (`config/env/test.js`)**
Comment clarifies this is a subset of `default.js` for route tests; extend when adding reserved names that need explicit coverage.

### Already in earlier commits
- Centralized server/client constants + `getUsernameRejectionMessage`
- NIP-05 legacy username lookup fix
- Legacy profile UX (`hasLegacyUsername` help text)

### Tests
Client tests updated and passing for directive message storage + controller error mapping.

---

**Note:** `git push` failed here (SSH agent permission). Please push `75d2b07ae` when convenient so CI re-runs on the follow-up commit.

guaka (commented):
**Review: approve with follow-ups applied**

The architecture (strict policy at routes, tolerant model for legacy data) is sound and well-tested. The main gap was documentation/UX polish rather than correctness.

Follow-up commit addresses:
- explicit “policy at routes, not Mongoose” documentation
- legacy vs new-regex separation for NIP-05
- signup-validation API messages in client tooltips
- test config `illegalStrings` subset documented
- PR description rewritten to reflect full scope

PR #2727, opened 2026-04-28
Fix profile reply stats display

guaka:
## Summary
- fix the profile sidebar reply stats instead of removing reply rate
- render reply time as the formatted duration returned by the API, for example `3 hours`
- keep displaying reply rate, including `0%`, and add focused component coverage

## Validation
- `npm run test:client -- modules/users/tests/client/components/ProfileViewBasics.tests.js`
- `NODE_OPTIONS="--require /private/tmp/jest.hide-canvas.js" ./node_modules/.bin/jest modules/users/tests/client/components/ProfileViewBasics.tests.js --no-watchman --runInBand`
- `./node_modules/.bin/eslint modules/users/client/components/ProfileViewBasics.js modules/users/tests/client/components/ProfileViewBasics.tests.js`

## Notes
- Existing unrelated commits/files remain in this PR from before this fix, including CI workflow updates, local Docker config, and signup test updates.
- `cursor/fix-profile-reply-stats` points at the same new commit as the PR head.


guaka:
Superseded by #2742 so the PR uses the correctly named branch `cursor/fix-profile-reply-stats`.

PR #2732, opened 2026-06-05
devcontainer settings

awongh:


guaka:
I found one integration issue in the devcontainer merge: the shared deploy/docker .env sets COMPOSE_PROJECT_NAME=trustroots, which Docker Compose treats as a CLI-level project-name override. Because the devcontainer consumes deploy/docker/docker-compose.yml first, that env value wins over the devcontainer overlay's name and can make the devcontainer share the normal local stack's project/network/volumes. I'll fix this by making the shared stack use an interpolated Compose name with the same default, while leaving COMPOSE_PROJECT_NAME free for callers/tools to override when they really mean to.

PR #2742, opened 2026-06-12
Fix profile reply stats display

guaka:
## Summary
- fix the profile sidebar reply stats instead of removing reply rate
- render reply time as the formatted duration returned by the API, for example `3 hours`
- keep displaying reply rate, including `0%`, and add focused component coverage
- merge current `main` so the branch has no conflicts

## Validation
- `npm run test:client -- modules/users/tests/client/components/ProfileViewBasics.tests.js`
- `./node_modules/.bin/eslint modules/users/client/components/ProfileViewBasics.js modules/users/tests/client/components/ProfileViewBasics.tests.js`
- `git diff --check`

## Notes
- Replaces #2727 so the PR uses the correctly named branch: `cursor/fix-profile-reply-stats`.
- Existing unrelated commits/files remain from the original branch history, including CI workflow updates, local Docker config, and signup test updates.


guaka:
does it fix this? "I wanted to reply on the same day, but it seems he deleted his profile so I couldn't. But this reduced my reply rate which is not fair. I really want to reply to everybody, as I thing everybody deserves at least a reply. Could you please restore my reply rate back to 100%?"

Issue #2375, opened 2021-07-09
Message textarea: Cursor does not go to new line when a new line is created


sitarane:
Checkout screen recording.

When writing a message to another user, creating a new line (shift enter) will create the new line below the one I'm on, but the cursor stays on the one I'm on. I need to move the cursor down myself after.

Granted, you're supposed to just hit enter and create a new paragraph. But "shift-enter" is often used by reflex because "enter" sometimes means "send the message" (in other apps).

Firefox on Linux.

https://user-images.githubusercontent.com/10289010/125042481-2b3dfc00-e09a-11eb-8b65-3d9655630a60.mp4

krumelmonster:
On firefox linux, I sometimes get this behaviour too and sometimes even if I don't press shift.

On Fennec F-Droid (using androids onscreen keyboard) I get such unintended behaviour more often than not without doing anything unusual. Also, pressing enter sometimes inserts a newline before inserting a space. Double line breaks get trimmed into a single line break. To work around these issues I've tried drafting on mobile in another app and copying over but even in this case, linebreaks get trimmed and other weird things happen.

My "workflow" for using trustroots in my mobile browser has become to type the message in a note-taking application, then copy it over into the message box and eventually fix some formatting issues that appeared when copying the message over. This is of course quite frustrating.

There are event listeners on the message-reply-content div to sentry. When I remove those, the behaviour is as expected (on desktop).

PR #2744, opened 2026-06-13
[codex] Fix various compose area issues

### LINKED ISSUE #2722: text drops of bottom of screen
In Firefox browser when writing in the text messages box, the box expands down but the page doesn't scroll. The result is the message and Send buttons disappear below the screen and cannot be accessed. See screen shot.

Image

guaka:
## What changed

- Bounds the message thread flex layout so Firefox can shrink the message list instead of pushing the reply composer below the viewport.
- Keeps the reply editor internally scrollable after a capped height while leaving the send controls visible.
- Fixes caret jumps and IME input in `TrEditor` by keeping `MediumEditor` out of React's re-render path while typing (also fixes `createRef` → `useRef` and stabilises the ctrl-enter subscription).
- Adds a targeted Firefox Playwright project with regression tests for long-reply overflow and multiline caret / composition input.
- Keeps the caret/composition test on Chromium via the `message-actions` project as well.
- Updates the e2e browser setup so Firefox is available in CI alongside Chromium.
- Limits Playwright JS coverage collection to Chromium (Firefox lacks the CDP coverage API).

Fixes #2722

## Validation

- `npm run test:client -- TrEditor`
- `npm run test:client -- ThreadReply`
- `bash -n scripts/e2e/test-e2e.sh`
- `npx eslint modules/core/client/components/TrEditor.js modules/messages/client/components/Thread.component.js modules/messages/client/components/ThreadMessages.js modules/messages/client/components/ThreadReply.js playwright.config.js tests/e2e/test.js tests/e2e/features/messages/messages-layout.spec.js tests/e2e/features/messages/message-actions.spec.js tests/e2e/support/message-reply-editor.js`
- `TRUSTROOTS_E2E_SKIP_COVERAGE_REPORT=true npm run test:e2e -- --project=messages --project=messages-firefox-layout --grep "reply composer"`

## Notes

- The homepage member-count change was moved directly to main and is no longer part of this PR.
- The full e2e suite was also attempted. The new `messages-firefox-layout` test and the existing `messages` project passed, but the full run ended with unrelated public/auth/member failures and later `ENOSPC: no space left on device` errors after generating large Playwright artifacts locally.


guaka:
we might as well want to fix other stuff related to the compose area

- [ ] when going back with the cursor when there are multiple lines, weird stuff happens
- [ ] this came in through support: "In the edit box for reply to messages (differently from the one where i write now) it is not possible to use dead keys, and compose key, that lead to be impossible to write certain words in certain languages."

guaka:
this may fix https://github.com/Trustroots/trustroots/issues/2375  (to review)

PR #2754, opened 2026-07-02
create react shell alongside angular *not to be merged*

awongh:
## Summary

Not for merging. 

Outline of some architectural changes as a mid-point to replace angular code in trustroots with react.

And as part of a larger effort to upgrade all dependencies in the app.

Enabled by e2e tests that cover most of the codebase.

Just for documentation / discussion purposes- don't merge this.

## Technical Description

- Added an incremental React-owned app root for selected public/support/statistics routes while the existing AngularJS app continues to own the rest of the site.
- Introduced shared route ownership metadata so server rendering and the React route shell agree on which paths are React-owned.
- Added separate React entry/template/assets (`react-main`) alongside the existing Angular `main` bundle to avoid dual bootstrapping on the same page.
- Wrapped server-provided globals in a React bootstrap/provider layer for user, settings, env, native-app state, and React Query.
- Updated Angular compatibility behavior so existing React components can run safely without Angular bootstrapped, with full-page handoff from Angular-owned pages to React-owned routes.

## Testing

- Added unit and React Testing Library coverage for bootstrap data, route ownership, React shell behavior, and Angular compatibility fallback.
- Added server coverage for React vs Angular root selection and direct-load behavior.
- Verified targeted client/server tests and Webpack build.


guaka:
Superseded by #2769, which builds on this foundational React shell and completes the Angular-to-React migration. Thank you @awongh for the architecture and implementation work that made the follow-up possible.

PR #2758, opened 2026-07-03
Make Pages artifact names unique per run attempt

guaka:
## Summary
- Give the GitHub Pages upload artifact a run-attempt-specific name.
- Configure deploy-pages to deploy that same artifact name so reruns do not see duplicate github-pages artifacts.

## Verification
- Parsed .github/workflows/test.yml with js-yaml.
- Ran git diff --check for the workflow file.


guaka:
Closing this PR because Pages deploy is skipped for pull_request runs; applying the same fix directly to main instead.

PR #2772, opened 2026-07-15
Add historical spam account cleanup

guaka:
## Summary

Adds a deliberately narrow, one-off maintenance command for the two identified historical automated-signup campaigns (4–6 July 2021 and 27–28 February 2023).

- Runs as a dry-run by default; deletion requires `--delete`.
- Deletes only non-public `user` + `suspended` accounts with no community activity, uploaded avatar, or moderation record.
- Preserves manually suspended accounts: any AdminNote excludes the account.
- Adds the OpenSpec proposal, design, specification, and rollout procedure.

## Why

The campaigns account for roughly 417k suspended records, obscuring the administration tool. Their BSON payload is about 401 MB and the targeted documents compress to about 85.6 MB in a Mongo dump (around 26% of the current dump).

## Validation

- Prettier check
- ESLint
- Node syntax checks
- Mongo-backed server test for dry-run, deletion, campaign bounds, and protected activity types
- `openspec validate purge-historical-spam-accounts --strict`

No browser end-to-end test: this is a manually invoked database-maintenance command with no user-facing route or UI.


guaka (commented):
## Review summary

**Verdict: Approve with minor nits.** This is a well-scoped, safety-conscious one-off maintenance tool. The selection criteria are narrow, dry-run is the default, deletion is explicit, and the pre-delete revalidation is a good touch.

---

## What works well

**Safety model is layered and appropriate**

1. Hard-coded campaign windows only (4–6 Jul 2021, 27–28 Feb 2023).
2. `public: false` and exactly `['user', 'suspended']` roles — excludes admins, moderators, shadowbanned accounts, etc.
3. User-document guards: empty `member` / `blocked` / `pushRegistration`, no uploaded avatar.
4. Cross-collection activity checks before counting/deleting.
5. Second activity check immediately before `deleteMany`, plus re-application of `candidateQuery()` in the delete filter.

That matches the intent: remove bot-wave records, keep anything with real activity or manual moderation.

**Implementation quality**

- Cursor + 1000-ID batches is the right shape for ~400k records.
- Service is testable (`candidateQuery`, `CAMPAIGN_WINDOWS`, `batchSize` exported/parameterised).
- CLI rejects unknown args and uses `exitCode = 1` on failure.
- OpenSpec proposal, design, archive, and living spec are in place.
- Skipping e2e is reasonable — no UI/route; server test against Mongo is the right layer.

**Test coverage (core path)**

The single integration-style test exercises dry-run, deletion, batching (`batchSize: 2`), every protected activity type, AdminNote, and document-level exclusions. The count math (13 candidates / 1 eligible / 12 protected) is correct.

---

## Issues / suggestions

### 1. Living spec still has a TBD purpose (minor, pre-merge)

`openspec/specs/historical-spam-cleanup/spec.md` still has:

> TBD - created by archiving change purge-historical-spam-accounts. Update Purpose after archive.

Worth replacing with a one-line purpose before merge so the living spec is complete.

### 2. Test gaps vs. tasks.md claims (minor)

`tasks.md` mentions error coverage; the test file has no failure-path test (e.g. simulated DB error). Not blocking, but either add a small unit test or soften the task wording.

Also worth adding:

- One account in the **Feb 2023** window (only July 2021 is used today).
- A **non-eligible role** case, e.g. `['user', 'suspended', 'admin']` or `['user', 'shadowban']`, to lock in the `$size: 2` constraint.

### 3. Operator safety UX (optional)

`archive-done-agenda-jobs.js` warns loudly on destructive reverse. Consider a similar `--delete` banner ("this permanently removes User documents; backup required") so a mistyped npm script is harder to run by accident.

### 4. Orphaned related data (acceptable, document)

Deletion is `User.deleteMany` only — no cascade like `removeProfile`. That is fine given the criteria (no messages, offers, contacts, etc.), but you may still leave orphans in collections this tool does not check, e.g. `SupportRequest`. Low risk for these bots; worth a line in the design/rollout doc.

### 5. Production load expectations (operational)

Dry-run and delete each do **12 queries per 1000 candidates** (24 when deleting, due to the re-check). For ~400k records that is thousands of round-trips. Acceptable for a one-off, but the rollout doc should say to run off-peak and expect measurable duration.

### 6. `npm` script placement (nit)

`cleanup:historical-spam` sits among `build:*` scripts. Grouping it near `agenda-maintenance` would make maintenance scripts easier to find.

---

## Security / data-loss assessment

| Risk | Mitigation |
|------|------------|
| Wrong accounts deleted | Campaign windows + role/public filters + activity checks |
| Manual suspension lost | AdminNote check |
| Concurrent restore during run | Pre-delete revalidation + `candidateQuery()` in delete |
| Accidental production delete | Dry-run default; explicit `--delete` |

I did not find a realistic path for a legitimate member account in these windows to be selected, assuming campaign identification in the proposal is accurate.

One subtle edge case: if an account is restored (`public: true` or roles changed) between the activity check and `deleteMany`, it may still count as "eligible" in stats but will not be deleted because `candidateQuery()` is re-applied. Harmless, but operators should know totals are "eligible at scan time", not "actually deleted".

PR #2774, opened 2026-07-15
Fix mobile map fallback

guaka:
## What changed

- Render the non-WebGL Leaflet fallback with Mapbox Streets raster tiles when a public Mapbox token is available; keep OpenStreetMap as the tokenless fallback.
- Recalculate the Leaflet search map after responsive mobile layout settles, preventing blank tile areas on iPhone.
- Remove the duplicate mobile Notes toggle; Community Notes remains in Filters.
- Clarify that the real-life-connection total is a lower bound based on submitted experiences since 2016.

## Why

The WebGL fallback always selected raw OpenStreetMap tiles and could initialise before the mobile map pane reached its final width. This made the fallback visually inconsistent and left unfilled tile space on iPhone.

## Validation

- Map, search-map, filter, and statistics tests passed.
- `npm run test:client` was also run. Its unrelated Angular directive suites fail on `origin/main` because Bootstrap template requests (tooltip, typeahead, datepicker) are not registered; the modified map and statistics suites passed.

### DIFF
--- a/modules/core/client/components/Map/LeafletMap.js
+++ b/modules/core/client/components/Map/LeafletMap.js
@@ -3,9 +3,7 @@ import L from 'leaflet';
 import PropTypes from 'prop-types';
 import React, { useEffect, useRef } from 'react';
 
-const OSM_TILE_URL = 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png';
-const OSM_ATTRIBUTION =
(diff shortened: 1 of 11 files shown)


guaka:
Added a concise history note to the real-life-connections statistic: it is a lower bound because members do not always submit an experience, and the underlying feedback flow dates from 2016. The mobile Notes shortcut introduced in #2767 was also removed because Community Notes already belongs in Filters.

PR #2769, opened 2026-07-15
Upgrade to Node.js 24 and finish Angular to React transition

guaka:
## Summary

- standardise development, CI, devcontainers and production on Node.js 24 with npm 11
- upgrade Webpack 4 to Webpack 5 and regenerate the dependency graph as a strict npm 11 lockfile
- complete the Angular-to-React transition and remove the Angular runtime, compatibility layer, routes, controllers, directives and templates
- move authentication, profiles, contacts, circles, search, offers, messages and supporting pages onto the React application shell
- preserve existing URLs through React route matching, redirects and navigation helpers
- document the supported runtime and React client architecture in OpenSpec

## Motivation

The previous application depended on an obsolete Node/npm combination, packages without Node.js 24 support, Webpack's legacy OpenSSL provider and a hybrid Angular/React client architecture.

This change establishes a current runtime and a single React client application, removing the remaining Angular bootstrap and compatibility code while retaining the existing user-facing routes and behaviour.

## Review guide

The commits are split by concern:

1. **Upgrade runtime and tooling to Node 24**
   - runtime declarations, container images, CI, dependencies, server rendering and runtime specifications

2. **Migrate the client from Angular to React**
   - React application shell, route ownership, migrated pages and removal of Angular code

3. **Restore migration behaviour and full test coverage**
   - React component, API, route, server and edge-case tests

4. **Stabilise React end-to-end coverage**
   - deterministic Playwright setup, selectors and migrated user-flow coverage

The dependency lockfile and removal of legacy Angular files account for most of the diff size.

## Acknowledgements

The React application shell and admin migration build on the foundational work in #2754 and #2755 by @awongh. That architecture and implementation helped make this complete migration possible.

## Compatibility and rollout

- Node.js 24 and npm 11 become the required development and deployment platform.
- The package lock moves to lockfile version 3.
- The browser application is React-only; Angular is no longer loaded at runtime.
- Existing public and authenticated URLs remain supported by the React router.
- Agenda remains on 1.0.3 and the direct MongoDB driver remains on 3.6.11.
- MongoDB server remains on 4.4; no database or data migration is required.
- Agenda and MongoDB driver upgrades are separated into draft follow-up #2800.
- The existing end-to-end coverage baseline is retained.

## Verification

- strict clean installation under Node.js 24 and npm 11
- full lint and formatting checks
- client: 216 suites and 1,266 tests passing
- client coverage: 100% statements, branches, functions and lines
- server coverage: 100% statements, branches, functions and lines
[...]

PR #2800, opened 2026-08-04
Upgrade Agenda and MongoDB drivers

guaka:
## Summary

- upgrade the application from Agenda 1.0.3 to 4.4.0
- upgrade the direct MongoDB driver from 3.6.11 to 4.17.2
- adapt worker scheduling, shutdown, maintenance scripts, and queued email/push calls to the newer APIs
- add Agenda integration coverage

## Relationship to #2769

This is intentionally stacked on #2769 so the Node 24 and React transition can keep the current production Agenda and MongoDB driver versions. Merge #2769 first, then retarget this PR to main.

## Before this is ready

This PR is deliberately a draft. Agenda 4 was previously introduced in #2415 and reverted in #2420 after jobs appeared with dates in the past. Before merging, it needs:

- an explicit regression test for existing/repeating jobs and nextRunAt values
- verification against production-shaped agendaJobs data
- a worker rollout and rollback plan
- the full unit, end-to-end, and production-image checks

The MongoDB server version remains 4.4; this PR changes the Node.js driver only.


guaka (commented):
## Conflict resolution + rebase notes

This PR was stacked on an August snapshot of #2769, which then evolved 82 more commits and was squash-merged, leaving GitHub seeing the whole React transition duplicated (DIRTY). The branch's unique work was a single commit, now rebased onto current `main` as one clean commit. How the three conflicts were resolved:

- **`config/lib/worker.js`** — kept this PR's Agenda 4 structure (defines in `start()`, `Promise.all` of `every()` → `start()`), dropped the `send push message` job definition since main retired push notifications (#2831), and kept main's "future push" comment in its place.
- **Push service + tests** — deleted on main by #2831; accepted the deletion rather than resurrecting the files.
- **`unlockAgendaJobs` bug (fixed during resolution)** — the commit left this on the MongoDB driver 3 callback API (`MongoClient.connect(uri, cb)`, `updateMany(…, cb)`, `client.close(cb)`), which driver 4.17.2 removed. The worker calls this at startup, so it would have crashed on deploy. The unit tests didn't catch it because they stubbed `MongoClient` in the old callback shape. Rewritten with promise chains (same error semantics: connect/update errors reach the callback, client always closed), and the three test stubs updated to the promise API.
- **`package-lock.json`** — the auto-merge had silently dropped TypeScript 7.0.2's platform-specific optional packages from main; regenerated with npm on Node 24.21.0.

**Verification:** `npm install` clean on Node 24.21.0 (also confirms dropping `legacy-peer-deps` is fine), 82 server tests pass — all 18 worker tests, all 3 agenda integration tests, email service, contacts — against a local mongo:4.4 container. Prettier + eslint clean.

## Review

The API adaptations are correct and consistent: promise-based `agenda.now()` in the email service with the matching `catchJobs` testutil change, `agenda.close()` in the gulpfile with close-error fallback, and the maintenance script correctly moves to `countDocuments()` (cursor `.count()` is gone in driver 4) with `find()` returning a cursor synchronously. The new integration tests are exactly the regression coverage this PR wants given the #2415/#2420 revert history: repeated `every()` persists a single job with `nextRunAt` set, and immediate jobs actually execute.

Non-blocking observations:

1. **`agenda._ready` in the integration test is a private API** (`modules/core/tests/server/agenda.integration.tests.js:23`). The public `ready` event would be less fragile across Agenda upgrades.
2. **Two MongoDB drivers now coexist**: mongoose 5.13 bundles driver 3.x internally while Agenda and the maintenance scripts use driver 4.17.2. Functional (and correctly keeps the server on 4.4), but it means two connection/pool implementations per process until a future mongoose 6+ upgrade unifies them — worth a line in the rollout plan.
3. **PR body is stale** — it still describes the stacking on #2769 and the pre-rebase state; worth updating before un-drafting. The "before this is ready" items (production-shaped `agendaJobs` verification, worker rollout/rollback plan) remain open.
4. **No e2e test added** — appropriate here: background-job infrastructure is covered by the new integration tests, and the user-visible flows that depend on emails already have e2e coverage.

guaka:
## Remaining open items

Code and description updates from the latest review pass are on the branch. These items are still open before merge:

### Must do before merge

1. **Approving review** — merge is still blocked on review approval.
2. **CI green** — wait for lint / client / server / e2e / production-image on the latest commit.
3. **Staging or sanitised production `agendaJobs` check** — run the verification steps in the PR body against real-shaped recurring jobs. The new integration tests cover the `#2420` pattern in isolation; they are not a substitute for a once-over on production-like documents.

### Follow-ups (non-blocking for this PR)

4. **Dual MongoDB drivers** — mongoose 5 still bundles driver 3.x while Agenda/direct usage is on 4.17.2. Expect mongoose deprecation warnings until a future mongoose 6+ upgrade unifies drivers.
5. **Agenda 4.4 still connects with the callback-style `MongoClient.connect(url, options, cb)` path** — driver 4 still accepts this via `maybeCallback`, but it is a fragile upstream pattern. No action required unless we see connect flakes.
6. **`unlockAgendaJobs` deliberately sets `nextRunAt: new Date()`** for unfinished locked jobs — those timestamps will look “just in the past” seconds later. That is intentional restart behaviour, not the `#2420` bug; operators should not treat those rows as a regression signal.
7. **No new e2e** — background-job infrastructure is covered by the integration tests; user-visible email flows already have e2e coverage.

guaka:
Review notes. The #2420 regression tests and the rollout/rollback notes are really useful. I pulled Agenda 4.4.0 down and ran it against a local `mongo:4.4` to check the worker path end to end. One blocker, then a few smaller things.

### 🔴 Blocking: the worker never starts processing jobs

`config/lib/worker.js:11` → `:72-83` now calls `agenda.every(...)` straight after `require('./agenda')`, without waiting for the connection. In Agenda 4, `every()` → `job.save()` → `saveJob()` reaches `this._collection.findOneAndUpdate(...)` synchronously, and `_collection` is only set once `MongoClient.connect` has finished (`db-init.js`). In the worker process nothing loads `config/lib/agenda` before `worker.start()`, so the connection is never ready at that point.

Reproduced with agenda@4.4.0 against mongo:4.4, using the same sequence as the PR (`define` → `Promise.all([every(...)])` → `start()`):

    start chain rejected: Cannot read properties of undefined (reading 'findOneAndUpdate')

So `agenda.start()` is never reached. `worker.js:43` logs "Error while initializing…", and the process then stays up doing nothing: no emails sent, no recurring jobs run. CI can't see this because `worker.tests.js` stubs `every`/`start`, `agenda.integration.tests.js:79` waits for `ready` before calling `every()`, and e2e doesn't start the worker.

Suggested fix: follow Agenda 4's documented order and start first. `start()` already waits for the private `_ready` promise:

    const startPromise = agenda
      .start()
      .then(function () {
        return Promise.all([
          agenda.every('5 minutes', 'check unread messages'),
          // …
        ]);
      })
      .then(function () { /* log */ });

I checked this order against the same database and it works. It could also be worth a regression test for the order itself, e.g. asserting `agenda.start.calledBefore(agenda.every)` in `worker.tests.js`, or an integration test that calls `every()` on a freshly constructed Agenda.

### 🟡 Smaller things

1. **Unhandled rejection in the retry path** — `config/lib/worker.js:146`: `job.save()` now returns a promise that nothing handles. On Node 24 an unhandled rejection kills the process, so a failed save during a retry would take the worker down. A `.catch()` that logs would be enough.
2. **Start-up failures leave a zombie worker** — `worker.js:43-49` only logs errors. Now that `worker.start()` reports failures through `done`, a `process.exit(1)` there would let the process supervisor restart the worker instead of leaving it idle. (The same already applied to unlock errors.)
3. **`.npmrc` / Dockerfiles** — dropping `legacy-peer-deps` (plus the matching `COPY` changes in `production.Dockerfile` and `deploy/docker/dev.Dockerfile`) is fine given `npm ci` passes. But it changes how everyone's local `npm install` resolves peer dependencies, so it's worth a line in the description, or its own PR.

### ✅ Checked and fine

- **Existing `agendaJobs` data on deploy:** I created a recurring job with Agenda 1.0.3, then re-registered it with Agenda 4.4.0 (with the fix above). The index name and keys (`findAndLockNextJobIndex`) are identical, so there's no index conflict. It stays a single document, and a future `nextRunAt` is kept, because `saveJob` puts a past-or-now `nextRunAt` under `$setOnInsert`.
- **Job handlers:** they all use the `(job, done)` signature, which Agenda 4 still supports (`definition.fn.length === 2`). `define` options (`priority`, `concurrency`, `lockLifetime`) and `job.attrs` are unchanged.
- **Retries:** Agenda 4 saves the job before emitting `fail`, so the retry `nextRunAt` set in the handler is still saved by its own `job.save()`.
- **Driver 4 changes:** `updateMany`/`modifiedCount`, `$unset: ''`, `countDocuments()`, `find()` returning a cursor synchronously, `agenda.close()` and promise-based `stop()` all look right. `connect-mongo` 4.6.0 now resolves to driver 4 through the existing override, which matches what it expects.

### Sequencing with #2864

#2864 is stacked on this branch, so this PR has to land first. If this one is squash-merged, #2864 will need rebasing onto `main` so it doesn't show duplicated history (the same thing that happened with #2769). The two PRs don't touch the same files beyond `package.json`/`package-lock.json`.

**Verdict:** request changes — just the worker start-up order. Once that's fixed, and ideally covered by a test, I'm happy for this to go in ahead of #2864, followed by the staging `agendaJobs` check from the description.

guaka:
Merged current main after #2863 landed, preserving the Gulp removal and regenerating the lockfile for the Agenda/MongoDB dependency changes. The PR is mergeable again; local lint passes and fresh CI is running.

guaka:
Review nit addressed in the PR description: the removal of legacy-peer-deps from .npmrc and both Dockerfiles is now documented, along with the clean-install implication. The latest completed CI run passed lint, builds, client, server, e2e and coverage; staging and rollout checks remain explicitly open.

guaka:
Merged current main (including #2862) into this branch in 7c4945d4c. Local lint passes; fresh CI is running.

guaka:
Merged the now-landed TypeScript PR from main in 98f142177. The Agenda and TypeScript manifest combined cleanly; the lockfile was regenerated and a clean npm ci plus local lint passed. Fresh CI is running.

guaka:
Merged the now-landed #2866 ESLint update from main in eb3f64eb9. Clean npm ci and local lint pass; fresh CI is running.

guaka:
Added regression coverage for worker restart recovery: an unfinished locked job is rescheduled, while a completed locked job and an already scheduled job retain their state. Also added a test confirming Agenda enqueue failures reach the email caller. The enqueue test and ESLint pass locally; the MongoDB integration case will run in CI because no local MongoDB service is available here.

guaka:
Added a further worker startup test: if recurring scheduling rejects, the startup callback receives that same failure. This covers the Agenda 4 Promise-based startup path. ESLint passes locally; CI is running the full suite.

guaka:
Added an integration regression that registers a recurring job with one Agenda instance, sets a future run time, then registers the same job through a fresh Agenda instance. It asserts the restart keeps one job and preserves that exact schedule. ESLint passes locally; CI is running the database case.

Issue #1309, opened 2020-03-17
Add keyword alarm for signups


rumwerfer:
Safety team regularly looks for new profiles with e.g. a certain name or certain expressions in their profile. There should be an automated e-mail (or message to Slack channel) when there is a new sign-up with names / expressions from a list, the list possibly editable from admin tools.

guaka:
Related work in #2801 has merged into `main`: administrator reports and acquisition-story rows now surface related-account leads based on restricted accounts, including normalised name/email-local-part signals and acquisition-story matches/similarities.

This helps manual moderation, but it does not implement a configurable keyword watchlist or automatic signup email/Slack alarms. The request here remains open.

**Deployment status (12 September 2026):** these changes are not on production yet. Deployment of `main` is expected soon; production verification is still pending.

Issue #1184, opened 2020-01-19
Shadowban


simison:
This is a master issue to track various work going to the "shadow ban" feature.

Goal:
- Add a new "shadowban" role which is like suspended, but works slightly differently.
- Add extra protection by filtering out suspended users in various features without them noticing.

PRs:

- [x] https://github.com/Trustroots/trustroots/pull/1180 don't let suspended users remove themself
- [x] https://github.com/Trustroots/trustroots/pull/1179 hide profiles with 'shadowban' role
- [x] https://github.com/Trustroots/trustroots/pull/1183 hide the messages themselves
- [x] https://github.com/Trustroots/trustroots/pull/1185 Add shadow-ban notice to rules
- [x] https://github.com/Trustroots/trustroots/pull/1186 Add moderator role
- [x] https://github.com/Trustroots/trustroots/pull/1187 Improve admin tools

Up next:
- [ ] Hide their messages in other peoples' inboxes (we're getting a lot of support messages regarding this - very annoying for the affected members)
- [ ] https://github.com/Trustroots/trustroots/pull/1181 Remove contact info in profile and offer descriptions for shadowbanned users
- [ ] https://github.com/Trustroots/trustroots/pull/1182 Do not send notifications to/from shadowbanned users
- [ ] Hide FB/social network links in other people's profiles
- [ ] #1189 Hide them from search
- [ ] Hide their contact requests
- [ ] [Don't send welcome emails to them](https://github.com/Trustroots/trustroots/blob/56adf8076a77ed1d91a88d3406196300bc99a48a/modules/users/server/jobs/user-welcome-sequence-first.server.job.js#L45-L51)

Other feature ideas:
- [ ] Don't consider their messages for reply rate calculation
- [ ] Hide their meet & hosting offers from others https://github.com/Trustroots/trustroots/pull/819
- [ ] Hide them in member search results
- [ ] Hide them as a contact in other people's profiles
- [ ] Don't send them any other automated emails
- [ ] Allow admins remove shadowban/suspended/moderator roles

TMC89:
I would like to add that a member through support wanted to remove a shadow banned member, but couldn't.

guaka:
most of it done, but good to double check

guaka:
[...]

PR #2801, opened 2026-08-04
Fix admin IP tracking and surface restricted-account matches

guaka:
## Summary

- restore the visitor address from trusted private proxy hops before Passenger records `lastIpAddress`, fixing production reports that always showed the Docker gateway `172.19.0.1`
- show acquisition stories in restricted-member reports
- surface bounded, informational related-account leads for suspended and shadowbanned members using normalised username/email-local-part signals and exact normalised acquisition stories
- show restricted-account leads on acquisition-story rows, including conservative trigram similarity for lightly edited stories
- omit suspended and shadowbanned members from existing contact lists
- keep shadow-hidden messages out of reply-rate and reply-time statistics
- exclude shadowbanned members from finish-signup reminders
- add a read-only current-role inventory to each administrator member report as a foundation for later role management
- remove the unreachable acquisition-story fallback that caused server branch coverage to fall below 100%

The IP problem was caused by the outer proxy forwarding through the Docker network while the inner Nginx/Passenger layer still treated the bridge address as the client. The configuration now trusts only private proxy sources and restores the forwarded visitor address.

## Issue #1184 audit (current `main`)

The checklist is stale in both directions.

Verified working:

- restricted members cannot remove their own account
- restricted profiles/avatars and shadow-hidden messages are hidden from other members
- member search and hosting/meet offer search exclude restricted accounts
- profile text and dedicated social/provider identifiers are hidden from shadowbanned viewers; the existing focused profile-sanitisation test still passes
- contact requests from/to restricted accounts are suppressed
- the three welcome-sequence emails and unread-message reminders are suppressed for restricted accounts

Addressed in this PR:

- existing confirmed-contact aggregation now omits restricted members
- shadow-hidden messages no longer reach reply-stat accounting
- finish-signup reminders now exclude both suspended and shadowbanned members
- administrator reports now show a read-only role inventory with descriptions; role editing and removal remain future work

Intentionally out of scope:

- no persistent rules notice is added
- no new role-mutation or remove-role action is added

The checked “moderator role” item is no longer current: the legacy role is deliberately rejected and full administrators own this access. Hiding hosting/meet offers remains unchecked in #1184 but is implemented and tested. The checked notification item should not be read as evidence for every automated email path.

Relates to #1184. I found no open PR duplicating the IP or account-matching work. Nearby work includes #1309 and #2792.

## Verification

- focused client component suite passed (21/21), with 100% focused coverage for the changed component
- focused server regression assertions passed for contacts, messages, finish-signup reminders, acquisition stories, and profile sanitisation
- admin inspection Playwright scenario passed (7/7 including setup)
- focused ESLint passed
- production Nginx syntax check passed
- all 22 current OpenSpec items validated strictly and the completed change was archived

The legacy local Gulp/Mocha server runner throws `mochaRunner.unloadFiles is not a function` after successful assertions. In the full local run, database clean-up hooks also timed out and caused unrelated follow-on failures; GitHub Actions is the authoritative full-suite check. The filtered Playwright wrapper reports incomplete repository-wide feature coverage after its seven selected tests pass because a single-file invocation cannot exercise the entire manifest.

### DIFF
--- a/deploy/docker/webapp.conf
+++ b/deploy/docker/webapp.conf
@@ -15,6 +15,19 @@ server {
     passenger_startup_file server.js;
     passenger_app_env production;
 
+    # Production terminates TLS in a host-level reverse proxy before requests
+    # cross Docker's bridge into this container. Restore that proxy's forwarded
+    # visitor address so Passenger's secure client-address header does not
+    # contain the Docker gateway (for example 172.19.0.1).
+    #
+    # Only private-network peers may supply this header. Public clients are not
+    # trusted as real-IP sources even if the container port is exposed.
+    set_real_ip_from 10.0.0.0/8;
+    set_real_ip_from 172.16.0.0/12;
+    set_real_ip_from 192.168.0.0/16;
+    real_ip_header X-Forwarded-For;
+    real_ip_recursive on;
+
     ## Sticky sessions are required for long polling support (socket.io)
     passenger_sticky_sessions on;
     passenger_min_instances 1;
--- a/modules/admin/client/components/AdminAcquisitionStories.component.js
+++ b/modules/admin/client/components/AdminAcquisitionStories.component.js
@@ -154,6 +154,7 @@ export default function AdminAcquisitionStories() {
                   onSort={sortBy}
                   sort={sort}
                 />
+                <th>Restricted matches</th>
               </tr>
             </thead>
             <tbody>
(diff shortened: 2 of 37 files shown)


guaka:
TODO: admin user search: if there's an exact match on a user it should show the exact match first. For example when searching for "alice" or "bob" there are 1000s of results. If there is a user with username alice or bob these should show at the top.

Issue #2669, opened 2024-01-26
make trustroots forkable


guaka:
Make the trustroots network forkable. If someone decides to run a fork, it should be easy for users to
- [ ] export data easily, there's https://www.trustroots.org/profile/edit/account but it's 3 different files, should be just one file
- [ ] import data easily
- [ ] claim their trustroots.org id, e.g. thru https://github.com/Trustroots/nostroots/issues/4

guaka:
We chose to go with nostr ([nostroots](https://github.com/Trustroots/nostroots)) but support for ActivityPub in trustroots seems nice, supposedly instantly connectable to sleepybike.  Possibly through https://gitlab.com/soapbox-pub/mostr

guaka:
Nostr and ActivityPub handle user identities differently, particularly in terms of portability and control. In Nostr, users maintain a consistent identity across different applications through a public/private key pair. This system allows for greater portability, as the same identity and social graph can be used in any app implementing the Nostr protocol, with the private key being central to access and identity control.

ActivityPub, used by platforms like Mastodon, links user identities to domain names and is controlled by server operators. This structure results in identities being more server-dependent. While it offers a degree of decentralization, server administrators have control over aspects like bans and server-to-server interactions. User migration between servers in the ActivityPub network can be challenging and requires cooperation between the servers.

irelativism:
I was about to suggest that bridge but seems is already referenced, I find this move very pleasing and happy to support it with whatever is necessary and to my ability feel free to reach me at irelativism@riseup.net. Cheers

P.S. https://mostr.pub/

guaka:
https://notes.trustroots.org/ is becoming a little more real, enough potential for mostr/bridge

guaka:
<img width="571" height="196" alt="Image" src="https://github.com/user-attachments/assets/435dfbc4-4d47-45b0-bc8b-47903b1d5a36" />

There's now a "Download all data" link that combines data into `trustroots-data.json`.

Simple next step would be to sign the json with a nostr key. A key used just for this purpose. If the key gets compromised the signed json created before whatever dates we think the compromise happened would still be valid. Maybe it's even possible to structure the data in a nostr way.

PR #2809, opened 2026-09-05
feat: sign the member data export with a Nostr key

leon-wbr:
Towards #2669 — the `nice to have: sign the file with a trustroots.org GPG or nostr key` checkbox, narrowed to Nostr by @guaka's July comment. Draft, because two of the decisions in here are maintainers' calls rather than mine.

## Determinism first, because signing depends on it

`GET /api/users/export` is byte-different on every download, and a signature over bytes that change on every download attests nothing useful — nobody could tell a changed digest from an unchanged one. Three separate things caused that, and only the first is obvious:

- **`exportedAt` is stamped at request time.** It stays in the file, because it is genuinely useful to the member, but it sits *outside* the signed payload. The signature's `signed` array names exactly what is covered: `format`, `version`, `profile`, `contacts`, `hostingOffers`.
- **Object key order is decided by Mongo, Mongoose, and Express**, none of which promise anything. The signed bytes are now [RFC 8785 JCS](https://www.rfc-editor.org/rfc/rfc8785) canonical JSON. A useful side effect: a file that has been through a text editor still verifies, because the verifier re-canonicalises rather than comparing raw bytes.
- **The contacts and hosting-offer aggregations return documents in no guaranteed order.** JCS does not reorder arrays, so canonicalisation alone would have missed this one. Both list sections are sorted by `_id` (canonical form as tiebreak) before they are digested *and* before they are sent, so the delivered file canonicalises to exactly what was signed.

The result is that the *attested content* is a pure function of the member's data: two downloads of unchanged data yield the same section digests and the same root digest. The attestation itself is not byte-identical between downloads — `created_at` moves, Schnorr uses fresh randomness — and it does not need to be. The root digest is the stable identifier of a member's exported state. There is a test for exactly this, and an e2e assertion that two downloads agree on everything but the timestamp and the signature.

## What the signature looks like

Detached and sectioned — a digest per section, a root binding those together with `format` and `version`, and a NIP-01 event attesting the root:

```json
"signature": {
  "alg": "nostr-schnorr-secp256k1/jcs-sha256",
  "signed": ["format", "version", "profile", "contacts", "hostingOffers"],
  "sections": { "profile": "<sha256>", "contacts": "<sha256>", "hostingOffers": "<sha256>" },
  "root": "<sha256>",
  "event": { "kind": 30410, "pubkey": "…", "created_at": 1788609600, "tags": [["x", "<root>"], ["m", "application/json"], ["d", "<username>"], ["alt", "…"]], "content": "", "id": "…", "sig": "…" }
}
```

Sectioning costs nothing and buys three things: a member can hand a fork the profile alone and have it still verify; adding a section later does not invalidate existing verifiers; and it is the shape a transparency log would later hang from.

On @guaka's *"maybe it's even possible to structure the data in a nostr way"* — my read is **envelope yes, payload no**. The envelope is a plain NIP-01 event, so any Nostr library verifies it for free, and the tag vocabulary is borrowed from NIP-94 (`x`, `m`, `alt`) though not its kind 1063, since 1063 means "a file is downloadable at this URL" and this file is generated on demand behind auth and never lives at an address. Restructuring hosting offers and contact edges into Nostr events would mean designing kinds and deciding whose key signs a contact edge — a protocol design project that would sink a "simple next step", and nostroots is the place where Trustroots data becomes Nostr-native anyway. **Kind 30410 is a placeholder** pending an assignment in `nr-common/constants.ts`.

## The property signing cannot give you

@guaka asked for: *"If the key gets compromised the signed json created before whatever dates we think the compromise happened would still be valid."*

**Signing alone cannot deliver that, and I would rather say so here than have it discovered later.** NIP-01 `created_at` is a field the signer chooses, so a thief holding the key can mint a signature dated last year. A signature proves authorship, not time.

What this PR does deliver is the weaker, honest version: once a key is declared compromised, a verifier rejects signatures whose *claimed* date falls after the compromise, and reports that an earlier claimed date is **unproven** — in those words, not as a bare "valid". If the intended reading was only "we can revoke forward without invalidating history", this delivers it.

Making the strong reading true needs an independent time anchor. The intended follow-up is OpenTimestamps: accumulate the day's root digests, publish the Merkle root, stamp it, and let Bitcoin attest that the root existed no later than a given block. **Deliberately not in this PR** — it is a daily job, an inclusion-proof endpoint, and roughly an order of magnitude more work than this. The sectioned root digest is shaped so that it is an addition, not a format break. Worth deciding which reading you meant before anyone builds it.

Related: **the key is hot by construction.** The export is generated per request, so the signing key has to be reachable from the web app's request path; compromise of the app server is compromise of the key. The alternative is a separate signing service with a narrow interface, which is more infrastructure than "a simple next step" implies. Naming it as the fork in the road it is.

## Privacy — a decision I did not make for you

This PR **does not widen what the export contains**. The field set is untouched.

But signing changes what the file *is*, and that is worth stating explicitly. The `contacts` section already carries **other members'** usernames, display names, home and current locations, Facebook IDs, and `emailHash` — an MD5 over a lowercased email address, reversible in practice for anything on a wordlist. Unsigned, that file is deniable: a third party who obtains it cannot tell it from a fabrication, so it is worth little to a data broker or a harasser. Signed, it becomes an artifact Trustroots vouches for, about people who never consented to it, transferable to anyone. That harm does not exist today; attestation creates it.

Sectioning is the first mitigation — the `contacts` attestation can be presented separately or withheld, and the verifier reports a withheld section rather than treating it as an empty one. Three further options are open, and they are a values call about the community's data rather than an engineering detail, so they are **yours to make**:

1. reduce the signed contact records to edges and identifiers, dropping the other party's location;
2. replace `emailHash` with something not reversible, or drop it, since a fork can obtain a Gravatar hash itself once the member is present;
3. accept the disclosure as it stands.

I would pick (1) and (2) together, but I did not want to make that call inside a PR about signing.

## Verifying

```
node bin/verify-data-export.js trustroots-data.json
```

Exit 0/1, no server, no database, no network. It reports rather than merely passing or failing, because a fork operator needs to know *which* section failed and what the signing key's status is:

```
Section profile: verified
Section contacts: mismatch
Root digest: mismatch
Attestation: verified
Signed by:   edcdaef9…
Claimed at:  2026-09-05T16:44:14.000Z (signer-chosen, unproven)
Key status:  active
Problem:     Section "contacts" does not match the signed digest; it has been altered.
Result:      FAILED
```

It also distinguishes an absent section from an empty one, so a partial file cannot be mistaken for a member with no contacts.

## Key custody and distribution

No key is generated in this PR, and **no nsec appears anywhere in it**. `config/data-export-signing-keys.json` ships with an empty key list; the pinned pubkey lands when a key is actually provisioned. That is the one remaining task before this can be merged as anything but a format change.

The key is a dedicated keypair used for nothing else — not the nostroots server key, not any member key, not a key that also posts — generated offline and injected at runtime as `TRUSTROOTS_EXPORT_SIGNING_NSEC`. With no key configured, the export is served **unsigned with a logged warning, never a 500**.

On distribution: a key published on trustroots.org is exactly as trustworthy as trustroots.org, which is the thing a fork exists to stop depending on. No single channel fixes that. `docs/data-export-signing.md` lists five — pinned in this repo (strongest, the `debian-archive-keyring` model), a signed git tag, `.well-known/nostr.json?name=_` (weakest, and labelled as such), a Nostr event to public relays, and a public announcement. Together they make a *silent* substitution require compromising several independent parties and a *loud* one visible. That is the achievable property; claiming more would be false. The rotation and revocation procedure is in the same doc, including why each history entry should be signed by the *next* key: otherwise whoever steals the signing key can also rewrite its own obituary.

## Testing

TDD throughout — canonicalisation and signing tests were written before the implementation, and one of them (a hostile file with a hole in the event's tag array) found a real bug: `verifyEvent` throws rather than returning false on a malformed event, and a verifier that crashes on a bad file is a verifier nobody runs.

- 63 new/updated server tests across canonicalisation, signing, verification, and the controller.
- **100% statement, branch, function, and line coverage** on all four new/changed server files, per `.nycrc-server`.
- `eslint .` clean across the repo.
- The e2e export test asserts version 2 and the determinism property; the scenario is registered in `feature-coverage.js`.
- No fixture contains real member data, and no test contains real key material — keys are generated per run.

**What I could not run here:** the full `gulp test:server` suite and the Playwright e2e suite, because this environment has no MongoDB and no Docker daemon. CI covers both. `openspec validate` likewise, as the CLI is not installed. The change proposal is at `openspec/changes/sign-member-data-export/`, unarchived, since the living spec currently *requires* the file to be unsigned and amending it belongs with the archive step once this is accepted.

## Out of scope, on purpose

The import half of #2669 (a fork's importer is the fork's code), a whole-network dump, encrypting the export, private messages, and OpenTimestamps. Worth saying plainly that signed exports whose signatures nobody checks are theatre — the value probably lands when there is an importer that checks them, which is the other unchecked box on the issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016vQPoS7hc1fyMeDmp78p4g


modules/users/server/services/data-export-verification.server.service.js
```diff
@@ -0,0 +1,276 @@
...
+    report.problems.push('The signature carries no Nostr event.');
+    return;
+  }
+
+  report.event = {
+    status: 'verified',
+    pubkey: event.pubkey,
+    kind: event.kind,
+    createdAt: event.created_at,
+    claimedAt: new Date(event.created_at * 1000).toISOString(),
```
guaka, line 96: [P2] Validate hostile event metadata before formatting or searching it

Starting from a valid exported JSON file, deleting `signature.event.created_at` (or setting it to `1e20`) throws `RangeError: Invalid time value` here. Setting `signature.event.tags = {}` instead throws from `findTag`. Both operations happen before the `verifyEvent` try/catch, so the standalone CLI terminates with a stack trace rather than its promised verification report. Validate the timestamp and tag container, or include these operations in the guarded validation path, and add regression cases for these JSON-representable malformed events.

modules/users/server/services/data-export-signing.server.service.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  try {
+    const decoded = nip19.decode(nsec);
+
+    if (decoded.type !== 'nsec') {
+      throw new Error(`Expected an nsec, got a ${decoded.type}`);
+    }
+
+    return decoded.data;
```
guaka, line 41: [P2] Validate the decoded secret scalar before accepting the key

A correctly encoded nsec is not necessarily a usable secp256k1 secret. For example, `nip19.nsecEncode(new Uint8Array(32))` decodes successfully here, but `signPayload` then throws `Expected valid bigint: 0 < bigint < curve.n` in `finalizeEvent`, outside this catch. The download controller forwards that error instead of serving the documented unsigned fallback, so a misconfigured key breaks exports for every member. Validate the decoded scalar inside the guarded key-loading path and cover this case in signing/controller tests.

guaka (changes requested):
Updated the branch with current main (7c997fb68), bringing in the existing Bookworm/node-gyp CI fixes. All applicable CI checks now pass, including both image builds, lint, server, client, end-to-end and the coverage gate: https://github.com/Trustroots/trustroots/actions/runs/34713566943

CI passed 1,699 server tests, 1,497 client tests across 262 suites, and 195 end-to-end tests. Server coverage remains 100% across all four metrics and the combined coverage gate passes. Locally, all 63 focused export tests pass under Node 16 with 100% statement, branch, function and line coverage across the four changed server files. OpenSpec strict validation of sign-member-data-export also passes.

Two reproducible error-handling gaps remain in the inline findings. Both reproductions use fictional member data and generated/test-only key material. The draft also still has the documented key provisioning and staging rotation tasks outstanding.

Issue #2829, opened 2026-09-12
Decide whether to restore or retire browser push notifications


guaka:
Do members still use browser push notifications, and should we restore or retire this feature?

Current main contains browser registration UI, Firebase client code and a service worker, but `modules/core/server/jobs/send-push-message.server.job.js` unconditionally completes every job without sending. Its comment records that notifications were disabled on 6 April 2026 because they did not work. The unreachable sender still calls the retired Firebase `sendToDevice()` API. Registration counts therefore cannot establish successful delivery.

The old mobile application is retired; its push integration is being removed separately in #2831. This issue concerns browser push only.

Before deciding:

- Check whether production has browser FCM configuration enabled; report presence only, not credentials.
- Count recent browser registrations and permission opt-ins using aggregate data, and establish whether users still see notification prompts/settings.
- Check delivery history and support reports; distinguish stored subscriptions from working delivery.
- Decide whether to restore browser push with a supported transport or retire it.

If restoring: update sender/client compatibility, verify real foreground/background delivery and click destinations, and add delivery regression coverage before enabling jobs.

If retiring: remove registration prompts/settings, Firebase client/server packages and service-worker build, handle existing subscriptions and queued jobs deliberately, and preserve email notifications.

This decision also affects the independent Node 24 upgrade (#2830): the old Firebase server dependency chain currently causes worker tests to be excluded on newer runtimes.

guaka:
#2831 now retires browser FCM as well as Expo/mobile push (rebased onto current main). Former delivery call sites are marked with `Future push:` comments if we rebuild later.

Closing this once that PR merges.

PR #2830, opened 2026-09-12
Upgrade to Node.js 24 independently of the React migration

guaka:
Extract the Node.js 24 runtime migration from #2769 while preserving the existing Angular/React frontend, Agenda 1.0.3, MongoDB driver 3.6.11 and MongoDB server 4.4.

This is an initial draft, not ready to merge. It aligns runtime declarations, CI, development containers and the proposed Passenger image, and records a validated OpenSpec proposal and compatibility investigation.

Verified: official Node 24.21.0 distribution checksum and runtime; bundled npm 11.19.0; proposal validation; formatting and diff checks.

Outstanding:
- Clean npm 11 dependency resolution currently fails: connect-mongo 4.6.0 expects MongoDB ^4.1.0; eslint-webpack-plugin 3.1.1 expects Webpack 5.
- Trial Webpack 4.47.0 resolution exposed those conflicts. The bump is deliberately not committed until a compatible graph is verified; the lockfile retains its existing dependency tree and only updates root engine metadata.
- Verify mmmagic, canvas and sharp; retain upload behaviour when fixing native compatibility.
- Resolve worker/Firebase compatibility in light of browser-push decision #2829. Push delivery is currently disabled; this PR must not silently enable it.
- Run installation, both frontend builds, full coverage, browser tests and actual Passenger application/worker startup. Local Docker was unavailable; no passing image build is claimed.
- Complete and archive OpenSpec before marking ready.

The separate mobile-push cleanup #2831 targets main. #2800 can be restacked on this runtime branch once it is viable; no existing PR branches have been rewritten or retargeted.

CI reproduced the dependency-resolution blocker in both development and production image builds: https://github.com/Trustroots/trustroots/actions/runs/34725333299 . The runtime assertions reached npm installation; application and worker startup remain unverified.


guaka:
## Review: Node.js 24 runtime upgrade

All CI checks are green and I independently verified the two riskiest claims. The technical core is solid — what stands between this and undrafting is mostly documentation hygiene.

### Independent verification

- **Strict npm 11 resolution now succeeds.** The body's headline blocker ("clean npm 11 dependency resolution currently fails") is no longer accurate. `npm install --dry-run` (npm 11, strict peers) against the committed `package.json` + lockfile is clean, and a full from-scratch resolution with the lockfile deleted also completes without `ERESOLVE`. The connect-mongo 4.4.1 downgrade + `overrides` entry + react-medium-editor removal + eslint-webpack-plugin 2.7.0 downgrade resolved the graph.
- **connect-mongo 4.6.0 → 4.4.1 is safe, and arguably a correctness fix.** 4.6.0 declares a `mongodb ^4.1.0` peer while the app runs driver 3.6.11; 4.4.1 officially supports driver 3. The override dedupes correctly (no nested `connect-mongo/node_modules/mongodb` in the lockfile). Both versions use kruptein 3.0.3, and the app only passes `{ client, collection }` (no crypto options), so existing sessions are unaffected.
- **The `TrEditor` rewrite is sound.** Confirmed in medium-editor 5.23.3's source that the placeholder extension reads the element's `data-placeholder` attribute when `options.placeholder.text` is unset, so the React-rendered attribute works — and unlike the old memoised wrapper, placeholder changes now propagate after mount (a small improvement). The IME/caret protection (never echo editor input back into the DOM) is preserved, `setContent` is guarded by `isApplyingExternalText`, and `destroy()` runs on unmount. Both call sites (`ThreadReply`, `AdminNotes`) only use supported props; `ThreadReply`'s key-based remount still works. The new tests cover callback freshness, placeholder updates, external resets and unmount cleanup — good.
- **Native modules:** CI rebuilds `sharp` and `mmmagic` from source and the server suite passes at 100% on Node 24; `canvas` 3.2.0 has Node 24 prebuilds; sharp 0.32.6's N-API (v7) prebuilds are runtime-independent, so the production stage's plain `npm ci` is fine.
- The build-time Node-24 assertion in both stages of `production.Dockerfile` is a good guard against base-image drift.

### Issues

**1. PR body and OpenSpec docs are stale (main blocker before undrafting).**
- Body: "the [webpack] bump is deliberately not committed", "lockfile … only updates root engine metadata", "dependency-resolution blocker" — all outdated; webpack 4.47.0, connect-mongo 4.4.1, the TrEditor rewrite and a regenerated lockfile are committed, and CI is green.
- `design.md` still says "That session-store downgrade was reverted" (it's in the diff) and "Strict npm 11 resolution still fails on connect-mongo 4.6.0" (no longer in the tree).
- `tasks.md` leaves completed items unchecked (server coverage, browser suite, docs alignment — all evidenced by CI).

**2. README ships PR-status language.** "This runtime migration is still a draft; clean installation and native-module compatibility remain under verification" shouldn't land on main — remove or rephrase before merge.

**3. Lockfile generated with `--legacy-peer-deps --ignore-scripts`, kept at `lockfileVersion: 2`.** `npm ci` works (CI proves it), but the first local `npm install` with npm 11 will rewrite it to lockfileVersion 3 and add peer-auto-installed packages — a large surprise diff for the next contributor. Since strict resolution demonstrably succeeds, consider regenerating strictly in this PR, or explicitly deferring with a tracked task. `--ignore-scripts` also means native builds weren't exercised during generation; Linux is covered by CI's `npm rebuild … --build-from-source`, but macOS arm64 mmmagic verification is still worth doing (already listed in design.md).

**4. Unexplained `TRUSTROOTS_BUILD_*` env vars in the e2e job.** The dev image installs `git` and the footer test passed on main via the git fallback, so it's unclear what failure this fixes on the new image. The adjacent `mkdir -p` lines have explanatory comments; these deserve one too (or a note in the PR).

**5. Minor / pre-existing (no action needed here):**
- `--openssl-legacy-provider` is still force-injected by `scripts/e2e/build-web.sh` and `test-e2e.sh`; with webpack 4.47.0's bundled md4-wasm it's likely removable — design.md already tracks this.
- `.nvmrc`/Docker pin an exact patch (24.21.0) while `engines` allows any 24.x — fine, but patch bumps now touch several files.
- `TrEditor`'s `onCtrlEnter` is `isRequired` yet has a `defaultProps` entry; `AdminNotes` passes a `className` that `TrEditor` ignores. Both pre-existing.
- Dependabot will keep proposing connect-mongo ≥ 4.6; a pin-reason comment would help.

**6. Correctly still outstanding:** actual Passenger app/worker startup in the production image (the build-time assertion doesn't prove runtime startup), and the Firebase/push question pending #2829 — good that this PR does not re-enable push delivery.

### Verdict

The hard problems (strict resolution, session store, editor wrapper, native modules) are solved and verified. Before undrafting: refresh the PR body, `design.md` and `tasks.md`, and drop the draft language from the README.

guaka (commented):
## Review

The runtime upgrade is in good shape. CI on `2a1721b1b` is green, including the new Passenger and worker startup check (web PID 82 and worker PID 1 both reported `v24.18.0`). I would update the OpenSpec record before merge; I would not block on the code.

### What holds up

The strict install path is coherent. Webpack stays on 4.47.0, `eslint-webpack-plugin` is back on the Webpack 4 line (2.7.0), and `.npmrc` pins `lockfile-version=2`, which npm 11 honours. `connect-mongo` stays at 4.6.0 and the override only forces its MongoDB peer onto the app’s 3.6.11 driver. With `{ client, collection }` it uses the Mongoose client object rather than constructing its own, and `/api/languages` going through session middleware in the startup check is a real smoke test of that combination.

`TrEditor` correctly drops `react-medium-editor` and keeps MediumEditor out of React’s update path. Placeholder text is the live `data-placeholder` attribute (MediumEditor’s CSS uses `attr(data-placeholder)`), callbacks stay fresh via refs, and external `setContent` is guarded so it is not reported as input. The existing integration test still drives the real editor, and the admin-notes end-to-end test types into `.tr-editor` and saves.

The production check is aimed at the right things: an Express route rather than a static nginx response, the Agenda ready line, and `readlink` of the process executable instead of `node` on `PATH`.

### Fix before merge

`openspec/changes/upgrade-node-runtime/design.md` and `tasks.md` still describe an earlier revision.

- The design says startup is unverified as of `2cbb5a9e7`, and that client and browser checks were still running at `f8a252a4c`. Head is `2a1721b1b`, and both the startup check and full client, server, and end-to-end runs passed.
- `tasks.md` still has “Verify production Passenger application and worker startup” and “Confirm full coverage and browser checks on the final revision” unchecked.
- The design says this change standardises on Node 24.21.0. Development and CI images do. The Passenger image is 24.18.0, which the startup log confirmed. The major-version assert in `production.Dockerfile` will not catch that gap.

The pull request body already matches the final run. The OpenSpec files should too, or the next archive step will record the wrong verification state.

### Do we need `sharp` and `mmmagic`?

Yes. They do different jobs, and only one of them is a runtime dependency.

- **`mmmagic` (production dependency)** is the libmagic binding used by `modules/core/server/services/file-upload.service.js`. Every upload calls `detectMimeType()`, which `require`s `mmmagic` unless `TRUSTROOTS_FILE_MAGIC_FALLBACK=true`. Avatar upload goes through that path. The header-byte fallback exists, so the package is not the only possible detector, but it is the detector production uses today.
- **`sharp` (dev dependency)** is required by `bin/generate-circle-images.js`. `npm run build` runs that script, and the production image builder runs `npm run build`, so `sharp` has to install and load during the image build. It is not used to serve requests. Avatar resizing uses GraphicsMagick (`gm` in `users.avatar.server.controller.js`); `.unsharp()` there is GraphicsMagick’s sharpening method, not the `sharp` package. The production image already installs the `graphicsmagick` package for that.

CI coverage of the two native modules is uneven:

- `sharp` is loaded during the production image build, because `generate-circle-images.js` requires it at the top of the file. A green `build-production-image` job means that binding loaded on the Passenger/Ubuntu base.
- `mmmagic` is not loaded by the startup check (`/api/languages` does not upload a file) and the server unit tests mock it. `npm ci` compiling it, and the dev image rebuilding it from source, do not prove `require('mmmagic')` succeeds in the running Passenger image. That is the remaining native-module gap for uploads.

### Residual risk, already mostly named

- macOS arm64 native install is still open.
- Worker and Firebase tests stay excluded, and push delivery stays off, pending #2829.
- `express.qs` is forced to the direct `qs` 6.16.0, which is outside Express 4.22.2’s `~6.15.1` range. The server suite is the evidence that query parsing still behaves. Keep that override on the removal list in the design, which it already is.
- The worker check starts `node worker.js` as PID 1, same as Compose, and Agenda did reach ready. It does not exercise a process supervisor.

